PR #20, opened 2021-06-21
Add provider check page

jamesrose:
screencapture-localhost-3000-registration-provider-check-2021-06-21-16_42_51

screencapture-localhost-3000-registration-about-npq-2021-06-21-16_43_01

### DIFF
--- a/app/views/registration_wizard/about_npq.html.erb
+++ b/app/views/registration_wizard/about_npq.html.erb
@@ -1,0 +1,55 @@
+<% content_for :before_content do %>
+  <%= render GovukComponent::BackLink.new(
+    text: "Back",
+    href: registration_wizard_show_path(@wizard.previous_step_path)
+  ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl" >Choosing an NPQ and Provider</h1>
+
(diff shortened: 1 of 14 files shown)


app/views/registration_wizard/about_npq.html.erb
```diff
@@ -0,0 +1,55 @@
...
+    <h1 class="govuk-heading-xl" >Choosing an NPQ and Provider</h1>
+
+    <h2 class="govuk-heading-m">Available courses</h2>
+
+    <p class="govuk-body">From September 2021, NPQ courses previously will be changed.</p>
+    <p class="govuk-body">These changes include:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>replacing the existing NPQs in senior leadership, headship and executive leadership</li>
+      <li>replacing the current NPQ in middle leadership with 3 new NPQs for teachers and leaders who want to develop their expertise in specialist areas of teaching practice:</li>
+      <ul class="govuk-list govuk-list--bullet">
```
asmega, line 19: i don't think this is valid markup. `ul < ul`, i would of thought it must be `ul < li`

asmega (approved):

app/views/registration_wizard/about_npq.html.erb
jamesrose: I think this is actually valid.

PR #45, opened 2021-06-25
Funding page

asmega:


app/lib/services/funding_eligibility.rb
```diff
@@ -0,0 +1,35 @@
...
+    attr_reader :course, :school
+
+    def initialize(course:, school:)
+      @course = course
+      @school = school
+    end
+
+    def call
+      case course.name
+      when "NPQ Leading Teaching (NPQLT)",
```
jamesrose, line 12: Is it worth extracting this groupings?

asmega: i think i'd rather have this at the moment. grouping is easy, ungrouping is harder. i know this is a grouping but not sure how pervasive it will be to deem it to be a grouping

jamesrose: Ok

jamesrose (approved):

PR #67, opened 2021-06-30
Set x-xss-protection using secure_headers

ethanmills:
Changes in application config are overridden by secure_headers


jamesrose (approved):

jamesrose:
Cheers @ethanmills

PR #92, opened 2021-07-19
Automate reporting

asmega:
### Context

- Generate CSV dashboard report
- So later it can be exported to BigQuery


mattonfoot (approved):
LGTM

PR #101, opened 2021-07-26
use FakeParticipantValidator in sandbox

asmega:
### Context

- Sandbox users want to be able to add TRN verified users to their test datasets
- We do not want to depend on test data that is outside of our control to be able to control this

- any user called John or Jane will be TRN verified in sandbox

### Guidance to review

- none


asmega:
I think we'll do this server side rather than client side of the api

PR #131, opened 2021-08-27
Support for local authorities

asmega:
### Context

- NPQs will be officially supporting Further Education (FE) bodies soon
- As part of that work we need to be able to support Local Authorities (LAs)
- LAs have an associated 8 digit `ukprn` and do not have a `urn`

- Add a table to persist LAs
- This is kept separate from schools incase there is different behaviour later down the line and seems like a logical to model these entities slightly differently
- On sending applications to ECF send over the `ukprn` too
- Introduce the concept of `Institution` for now simple returns a mixed collection of `School` and `LocalAuthority`. We can wrap these at a later stage if needed but is not needed at this moment in time

### Guidance to review

- This is a breaking change for users that are part way through an application
- We should switch over to server side session first. This will give us the ability to invalidate sessions with ease. Post deployment we can then invalidate all sessions
- depends on https://github.com/DFE-Digital/early-careers-framework/pull/916 which must be deployed first


app/views/registration_wizard/choose_school.html.erb
```diff
@@ -17,37 +17,37 @@
...
-          <p class="govuk-body">Please choose from schools located in <%= @form.wizard.store["school_location"] %></p>
+        <% if @form.possible_institutions.blank? %>
+          <p class="govuk-body">Please choose from schools located in <%= @form.wizard.store["institution_location"] %></p>
 
           <p class="govuk-body">No schools were found with that name, please try again</p>
 
-          <%= f.govuk_text_field :school_name,
+          <%= f.govuk_text_field :institution_name,
                                  width: "full",
                                  label: { text: "Enter your school name" } %>
```
jamesrose, line 27: These aren't just schools anymore. Do we plan on changing the copy on this page?

app/controllers/institutions_controller.rb
```diff
@@ -0,0 +1,16 @@
+class InstitutionsController < ApplicationController
+  def index
+    schools = School
```
jamesrose, line 3: I can see what you're doing here! Did you consider having two endpoints, one for schools and Local Authorities, and doing the munging in JavaScript? You'd be able to set a `type` on the form and then either do STI, or have distinct `school_id` and `local_authority_id` columns on an application.

app/views/registration_wizard/choose_school.html.erb
asmega: Yeah, these will come in once all the backend changes are in

app/controllers/institutions_controller.rb
asmega: I don't recall but if I did I probably threw it the window very quickly. accessible autocomplete out the box takes one api endpoint. so you'd have to write some custom javascript to do this munging prior to it hitting the autocomplete code. I don't see a benefit to having two endpoints then combining them when we can just do the same thing server side.

jamesrose (approved):

PR #143, opened 2021-09-03
permit seven digit urns for sandbox usage

asmega:
### Context

- For sandbox we have test schools with 7 digit urns

- Permits schools with 7 digit urns

### Guidance to review

- none


mattonfoot (approved):
LGTM

PR #146, opened 2021-09-06
schedule job to sweep stale sessions

asmega:
### Context

- Sessions are now server side and therefore need pruning once stale

- Prune any stale sessions last updated over 15 days ago

### Guidance to review

- none

### DIFF
--- a/.github/workflows/schedule_sweep_stale_sessions.yml
+++ b/.github/workflows/schedule_sweep_stale_sessions.yml
@@ -1,0 +1,81 @@
+name: Schedule sweep stale sessions
+
+on:
+  schedule:
+    - cron:  '30 3 * * *'
+
+jobs:
+  deploy:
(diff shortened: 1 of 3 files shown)


kicferk1 (approved):
If you end up with more jobs, perhaps it would be worth looking at the way R&P handles cron jobs with `delayed_job`. It's probably easier to track when it's handled inside application, rather than on another platform (github actions in this case).

asmega:
> If you end up with more jobs, perhaps it would be worth looking at the way R&P handles cron jobs with `delayed_job`. It's probably easier to track when it's handled inside application, rather than on another platform (github actions in this case).

I took a look at cron-esque things for delayed job but thought they were all mediocre at best. They all rely on enqueuing the next job after a job has completed which seems brittle to me. Which it why I opted for the workflow as I see that as far less likely to fail.

PR #157, opened 2021-09-13
funding radio labels no longer bold

asmega:
### Context

- Funding page radio labels were bold which does not match govuk component

- Unbold to match govuk component

Before
Screenshot 2021-09-13 at 10 26 12

After
Screenshot 2021-09-13 at 10 26 02

### Guidance to review

- none

### DIFF
--- a/app/views/registration_wizard/funding_your_npq.html.erb
+++ b/app/views/registration_wizard/funding_your_npq.html.erb
@@ -25,7 +25,9 @@
         :value,
         :text,
         :description,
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):
Are there any other radio buttons with bold labels that we don't want to be bold?

asmega:
> Are there any other radio buttons with bold labels that we don't want to be bold?

Just these ones i think. It's because of https://github.com/DFE-Digital/govuk-formbuilder/issues/315 and I don't think we have hint texts on any other radios

PR #158, opened 2021-09-13
Govuk components 2 1 0

asmega:


kicferk1 (commented):
How confident are you in your test coverage?

Things that I don't remember breaking were `button: true` links. Other than that things tended to raise exceptions, so as long as you have a test rendering a page you should be fine.

asmega:
> How confident are you in your test coverage?
> 
> Things that I don't remember breaking were `button: true` links. Other than that things tended to raise exceptions, so as long as you have a test rendering a page you should be fine.

Test suite picked up on all the breaking changes. The journey is fairly small so I walked through by hand afterwards. Only fixes I recall were the button links and the fieldset legend (although that was an earlier regression I only caught now)

kicferk1 (approved):

PR #162, opened 2021-09-15
certain urns are eligible for funding

asmega:
### Context

- Certain URNs which are FE related are eligible for funding regardless of course taken

- short circuit special urns to be eligible for funding

### Guidance to review

- none

### DIFF
--- a/app/lib/services/funding_eligibility.rb
+++ b/app/lib/services/funding_eligibility.rb
@@ -9,6 +9,8 @@ module Services
     end
 
     def call
+      return true if eligible_urns.include?(institution.urn)
+
       case institution.class.name
       when "School"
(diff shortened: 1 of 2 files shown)


asmega:
> Does R&P repo duplicate any logic around npq funding eligibility, or is this repo the source of truth for that?

Funding eligibility logic is only found here. There is no duplication elsewhere. 

> Also, are these arbitrary? Or is there some place outside of us that maintains that list?

We just get handed the CSVs, but there is reasoning behind them that's not visible here. I foresee this change being temporary and this class will change again potentially quick drastically in the near future so i'm not worried about it

kicferk1 (approved):

PR #167, opened 2021-09-20
remove set-value js dependency

asmega:
### Context

- remove a js dependency that does not appear to be needed or used

- remove js dependency

### Guidance to review

- none


kicferk1 (approved):
Hell yeah.

PR #174, opened 2021-09-22
add convenience method on Application

asmega:
### Context

- Changes to policy may affecting funding eligibility
- Support queries to create new applications on users behalf, this allows for a quick mechanism to workout the funding eligibility for an application

- Add convenience method that wraps a service to use on the command line as quick way to check funding eligibility for an application

### Guidance to review

- none

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -3,4 +3,8 @@ class Application < ApplicationRecord
   belongs_to :course
   belongs_to :lead_provider
   belongs_to :school, foreign_key: "school_urn", primary_key: "urn", optional: true
+
+  def calculate_funding_eligbility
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):
so this is so that we can do
```
application.calculate_funding_eligbility
```
instead of
```
Services::FundingEligibility.new(course: application.course, institution: application.school, headteacher_status: application.headteacher_status).call
```
right?

asmega:
> so this is so that we can do
> 
> ```
> application.calculate_funding_eligbility
> ```
> 
> instead of
> 
> ```
> Services::FundingEligibility.new(course: application.course, institution: application.school, headteacher_status: application.headteacher_status).call
> ```
> 
> right?

Yes, thats what I'm doing at the moment I still have a whole load of improvements I want to add to ease with support queries

PR #178, opened 2021-09-23
add brakeman

asmega:
### Context

- add brakeman to CI suite

- added brakeman to CI suite

### Guidance to review

- test were run with an error which causes build to fail
- adding ignore file caused build to pass


kicferk1 (approved):
Nice

PR #179, opened 2021-09-23
add possible funding page

asmega:
### Context

- If we think user is eligible for funding we now show possible funding screen before the user checks their answers
- If they are not eligible for funding they would get the "Choose your funding" page instead

- Add extra page to journey
- Fixed back button on check answers page, now goes to where it should

#### Screenshots

Screenshot 2021-09-23 at 16 45 32

### Guidance to review

- none

### DIFF
--- a/app/lib/forms/possible_funding.rb
+++ b/app/lib/forms/possible_funding.rb
@@ -1,0 +1,15 @@
+module Forms
+  class PossibleFunding < Base
+    def next_step
+      :check_answers
+    end
+
+    def previous_step
+      :choose_school
+    end
+
+    def course
+      @course ||= Course.find(wizard.store["course_id"])
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/lib/forms/possible_funding.rb
```diff
@@ -0,0 +1,15 @@
+module Forms
+  class PossibleFunding < Base
```
kicferk1, line 2: Do we want a potentially very basic test for this?

asmega: It's mostly covered by the capybara test. I can add a unit test though

kicferk1: I only mentioned it because I think other pages have unit tests, and we probably would want it for code consistency. But it's not like the page is very complex...

asmega: I've just added some, should cover everything.

kicferk1 (approved):

PR #188, opened 2021-10-04
Tweak start page

fofr:
Paired with James Liddell

- Remove the single bullet point list
- Moved who should use the service into the top section, it’s not something you do before you start, it’s an eligibility thing
- Rearranged the points in the "You must" list to follow the order that someone would go through

| Before | After |
|--|--|
|Screenshot 2021-10-01 at 16 10 11 | Screenshot 2021-10-01 at 16 09 31|


asmega:
Sorry, this'll need rebasing due to the ASO changes

PR #189, opened 2021-10-05
Add Additional Support Offer Journey

asmega:
### Context

- Adds `Additional Support Offer` aka `ASO` to the service
- This is treated as a course
- New user journey to be able to register for this course

- Adds a new course - This will be added to the database at deployment time by hand
- This course is a lookup inside ECF and the change has already been deployed
- Adds new journey for ASO
- Ordering of pages have been tinkered with to slot in the new journey
- Reused the same columns for `headteacher_status` and the `course` so no change needed inside ECF

### Guidance to review

- We don't have review apps
- Since this is the next thing to release, we'll merge, do manual integration testing, apply any patches then release

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,14 +1,6 @@
 class Course < ApplicationRecord
-  def studying_for_headship?
-    name == "NPQ for Headship (NPQH)"
-  end
-
   def npqh?
     name == "NPQ for Headship (NPQH)"
-  end
-
-  def studying_for_aso?
-    name == "Additional Support Offer for new headteachers"
   end
 
   def aso?
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -14,4 +14,8 @@ class Course < ApplicationRecord
   def aso?
     name == "Additional Support Offer for new headteachers"
   end
+
+  def npqsl?
+    name == "NPQ for Senior Leadership (NPQSL)"
+  end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -3,7 +3,15 @@ class Course < ApplicationRecord
     name == "NPQ for Headship (NPQH)"
   end
 
+  def npqh?
+    name == "NPQ for Headship (NPQH)"
+  end
+
   def studying_for_aso?
+    name == "Additional Support Offer for new headteachers"
+  end
+
+  def aso?
     name == "Additional Support Offer for new headteachers"
   end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -2,4 +2,8 @@ class Course < ApplicationRecord
   def studying_for_headship?
     name == "NPQ for Headship (NPQH)"
   end
+
+  def studying_for_aso?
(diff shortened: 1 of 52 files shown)


app/models/course.rb
```diff
@@ -2,4 +2,20 @@ class Course < ApplicationRecord
   def studying_for_headship?
     name == "NPQ for Headship (NPQH)"
   end
+
+  def npqh?
+    name == "NPQ for Headship (NPQH)"
+  end
```
kicferk1, line 8: Could this just be `alias_method : npqh?, : studying_for_headship?`?

app/models/course.rb
```diff
@@ -2,4 +2,20 @@ class Course < ApplicationRecord
...
+
+  def studying_for_aso?
+    name == "Additional Support Offer for new headteachers"
+  end
+
+  def aso?
+    name == "Additional Support Offer for new headteachers"
+  end
```
kicferk1, line 16: Ditto here with alias

app/models/registration_wizard.rb
```diff
@@ -114,6 +124,18 @@ def form_for_step(step)
 
 private
 
+  def aso_needs_funding?
+    store["aso_funding"] == "yes"
+  end
+
+  def studying_aso?
+    course.name == "Additional Support Offer for new headteachers"
+  end
```
kicferk1, line 133: `course. aso?`?

db/seeds.rb
```diff
@@ -6,8 +6,9 @@ def seed_courses!
     { name: "NPQ for Senior Leadership (NPQSL)", ecf_id: "a42736ad-3d0b-401d-aebe-354ef4c193ec" },
     { name: "NPQ for Headship (NPQH)", ecf_id: "0f7d6578-a12c-4498-92a0-2ee0f18e0768" },
     { name: "NPQ for Executive Leadership (NPQEL)", ecf_id: "aef853f2-9b48-4b6a-9d2a-91b295f5ca9a" },
+    { name: "Additional Support Offer for new headteachers", ecf_id: "7fbefdd4-dd2d-4a4f-8995-d59e525124b7", description: "The Additional Support Offer is a targeted support package for new headteachers." },
```
kicferk1, line 9: No descriptions for other courses?

app/models/course.rb
asmega: I've removed all the duplicated methods and consolidated. so everything now goes through `course.npqh?` or `course.aso?`

app/models/registration_wizard.rb
asmega: this is gone now in favour of `#aso?`

db/seeds.rb
asmega: No, only for this one. alternatively i can not have the column and use i18n. When the list out the courses this is the only one with a hint text

kicferk1: Alright, I think it's fine, just wanted to make sure.

kicferk1 (approved):

PR #207, opened 2021-10-20
POST participant validate endpoint

asmega:
### Context

- This is the opposite half of https://github.com/DFE-Digital/early-careers-framework/pull/1315
- There will be subsequent change to remove the old endpoint which will be unused once this PR is deployed

- Switchover to use POST endpoint for TRN validation

### Guidance to review

- API endpoint has been deployed

### DIFF
--- a/app/lib/services/participant_validator.rb
+++ b/app/lib/services/participant_validator.rb
@@ -10,8 +10,9 @@ module Services
     end
 
     def call
-      request = Net::HTTP::Get.new(uri)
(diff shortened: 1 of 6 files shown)


app/lib/services/dqt_client.rb
```diff
@@ -1,39 +0,0 @@
-require "net/http"
-
-module Services
-  class DqtClient
```
kicferk1, line 4: Why are we deleting this?

PR #273, opened 2022-01-04
add exporter for user with a TRN

asmega:
### Context

- Internal users have requested data for users with trns for their `ecf_id`, `id` and `trn` 

- Add exporter to output required data to stdout

### Guidance to review

- none


leoapost (approved):

asmega:
may not be needed anymore

PR #285, opened 2022-01-13
Add rake task for syncing applications to ecf

FriedSock:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-852

Corresponding PR: https://github.com/DFE-Digital/early-careers-framework/pull/1622

Adds a rake task to sync all the `eligible_for_funding` statuses with ecf


asmega (approved):
Code looks fine. My only concern would be large volume of calls when the rake task is ran which may cause issues elsewhere.

PR #350, opened 2022-04-21
[CST-660] Implement ChildcareProviders model and importer

Lockyy:
### Context

Within the new flow we need the ability to store ChildcareProviders separately to Schools so they can be looked up. Despite this, they are essentially the same concept that fill the same niche. 

To allow the two to serve the same purpose but differ in subtle ways I've moved the `schools` table to `educational_institutions` and added single table inheritance to add a `School` and `ChildcareProvider` model as children of `EducationalInstitution`.


app/lib/services/childcare_providers/importer.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      class ChildcareProviderCSVRow
+        attr_reader :csv_row
+
+        def initialize(csv_row:)
+          @csv_row = csv_row
+        end
+
+        def urn
+          csv_row["Provider URN"] # IS THIS RIGHT? Could it be 'Registered person URN' instead?
```
jamesrose, line 83: Check with Lauren or Umer

PR #353, opened 2022-04-28
[CN-109] - Add "Choose your Nursery" pages

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CN-109

Users who tell us that they are working for or as a private childcare provider should be directed down a path where they can tell us whether they have an Ofsted Unique Reference Number (Ofsted URN). To support this, we'll need to import the Ofsted childcare provider data into our database for lookup purposes.

- Rename `schools` table to `educational_institutions` and move `School` to be a subclass of the new parent entity `EducationalInstitution`.
- Adds `PrivateChildcareProvider` as second subclass of `EducationalInstitution`.
- Add importer and rake task for importing Ofsted childcare providers as `PrivateChildcareProviders`.
- Adds `/registration/have-ofsted-urn` page gathering information on whether a user who has reached the page has an Ofsted URN or not. 
- Adds `/registration/choose-private-healthcare-provider` page allowing user to search `PrivateChildcareProviders` by URN.

Uploading changes_small.mov…


app/lib/forms/have_ofsted_urn.rb
```diff
@@ -0,0 +1,37 @@
+module Forms
+  class HaveOfstedUrn < Base
+    attr_accessor :has_ofsted_urn
+
+    validates :has_ofsted_urn, presence: true, inclusion: { in: %w[yes help no] }
```
asmega, line 5: should `help` be here? or is the option missing in `next_step`?

app/views/registration_wizard/have_ofsted_urn.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      <%= f.govuk_radio_buttons_fieldset(:has_ofsted_urn,
+            legend: {
+              text: "Do you have an Ofsted unique reference number (URN)?",
+              size: "xl",
+              tag: "h1"
+            }
+          ) do %>
+
+        <%= render GovukComponent::DetailsComponent.new(summary_text: "I need help with this") do %>
+          <p>You can <%= govuk_link_to "search the public Ofsted register (opens in a new tab)", 'https://reports.ofsted.gov.uk/childcare' %> to find a URN.</p>
```
asmega, line 26: it says `opens in a new tab` but pretty sure it won't open in a new tab?

app/webpacker/packs/institution-picker.js
```diff
@@ -33,14 +33,14 @@ institutionPicker.enhanceSelectElement = (configurationOptions) => {
       if (object === undefined) {
         return ""
       } else {
-        return object.name + " - " + object.address
+        return [object.urn, object.name, object.address].filter(n => n).join(' - ')
```
asmega, line 36: can `filter(n => n)` be removed? does it actually have any affect?

app/lib/forms/have_ofsted_urn.rb
Lockyy: No it shouldn't, this was an old option in the designs that I removed everywhere except here.

asmega (approved):

PR #355, opened 2022-04-28
[CN-66] Targeted support funding

asmega:
### Context

- https://dfedigital.atlassian.net/browse/CN-66
- There is another eligibility field that needs to be stored
- This won't actually be used till much later so it just for storage purposes at the moment

- Added the ability to refresh all schools from GIAS
- This is so we can populated the `number_of_pupils` field
- Added class so we can determine if a school is eligible for `targeted_support_funding_eligibility`
- Store the determination of this outcome onto the application
- There is also an explicit list with FE bodies that are eligible

### Guidance to review

- Calling `Services::ImportGiasSchools.new(refresh_all: true).call` should now refresh all data from GIAS
- Complete a journey that meets the criteria for `targeted_support_funding_eligibility`
- The outcome of this should now be saved against the application

### DIFF
--- a/app/lib/services/eligibility/targeted_support_funding.rb
+++ b/app/lib/services/eligibility/targeted_support_funding.rb
@@ -9,6 +9,7 @@ module Services
 
       def call
         return false if institution.nil?
+        return true  if eligible_fe_ukprns.include?(institution.ukprn)
         return false if institution.is_a?(LocalAuthority)
         return false if institution.number_of_pupils.nil?
         return false if institution.number_of_pupils.zero?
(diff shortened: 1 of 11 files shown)


app/lib/services/eligibility/targeted_support_funding.rb
```diff
@@ -0,0 +1,187 @@
...
+          41, # Studio schools
+          42, # Academy alternative provision converter
+          43, # Academy alternative provision sponsor led
+          44, # Academy special converter
+          45, # Academy 16-19 converter
+          46, # Academy 16 to 19 sponsor led
+        ].map(&:to_s)
+      end
+
+      def eligible_fe_ukprns
```
Lockyy, line 62: How often is this list expected to change? If it's infrequent leaving it hardcoded works, however if it's going to be happening more than a few time a year I feel like storing this in the DB against the eligible institution may be better so that adjustments don't require deployment.

A consideration against that though would be that the institutions are split amongst two separate tables, schools and local_authorities. And that may potentially become three if private_childcare_providers are split out of schools in the other PR.

The above (splitting URNs across multiple tabbles) is making me reconsider back in favour of storing PrivateChildcareProvider and School models within the same single table inheritance typed table.

app/lib/services/handle_submission_for_store.rb
```diff
@@ -33,6 +33,7 @@ def call
           works_in_school: store["works_in_school"] == "yes",
           employer_name: store["employer_name"].presence,
           employment_role: store["employment_role"].presence,
+          targeted_support_funding_eligibility: targeted_support_funding_eligibility,
```
Lockyy, line 36: We're adding it to the DB here, should we also be adding it to the body sent over in Services::NpqProfileCreator in prep for ECF receiving this data? Or should we wait for the ECF side of things to define the API?

app/lib/services/eligibility/targeted_support_funding.rb
asmega: Info from ticket suggests once or twice a year. I'm not too fussed about the re-deploying as it's not particularly painful. But if it does become more frequent we can swap this to a DB based solution as you suggested.

`PrivateChildcareProvider` will have to duck type with `ukprn` otherwise we can add en extra guard in when using this class

Lockyy (approved):
Looks good 👍

PR #361, opened 2022-05-03
Add new courses

asmega:
### Context

- Added new NPQ courses plus other minor tweaks so unblock development of courses

- add new NPQ courses
- this unblocks other journeys for development
- tests are pended and will need updating to reflect new requirements
- added ability to hide courses
- added ability to render courses in desired order

### Guidance to review

- Assuming everything has been seeded
- Go through the journey
- When choosing NPQ
- Should see new courses
- Should not see ASO course
- Courses should be ordered correctly
- Selecting a new course will likely not take you to correct journey as it is yet to be implemented

### DIFF
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -60,9 +60,9 @@ module Forms
 
     def courses
       if wizard.query_store.inside_catchment? && wizard.query_store.works_in_school?
-        Course.all
+        Course.where(display: true).order(:position)
       else
-        Course.all - Course.where(name: "Additional Support Offer for new headteachers")
+        Course.where(display: true).order(:position) - Course.ehco
       end
     end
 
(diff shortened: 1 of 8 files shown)


Lockyy (approved):
This looks good to me, though the order doesn't seem to match the flowchart Dominique provided.

| Flowchart | Seeded Result |
|-|-|
| courses | app |

asmega:
> This looks good to me, though the order doesn't seem to match the flowchart Dominique provided.
> 
> Flowchart	Seeded Result
> courses	app

It does match the prototype though... We can tweak the ordering later, the mechanism is now there at least so should be trivial to move around

Lockyy:
Yeah sure, happy to fix later. To clarify though, Dominique and I discussed what the source of truth should be when there are differences between the two and decided that the flowchart has the final say. I'll attach the latest PDF export from Lucid to the epic ticket so we can work off the same document.

PR #367, opened 2022-05-04
add Targeted support funding eligibility to admin

asmega:
### Context

- Admins are not able to see `Targeted support funding eligibility` flag
- Adding this will enable this and also help with PO review to ensure everything is working as expected

- In admin area surface `Targeted support funding eligibility` when viewing an application

### Guidance to review

- Login as an admin
- View applications
- Should be a field labelled `Targeted support funding eligibility` which correctly mirrors that of the viewed application


Lockyy (approved):
lgtm

PR #380, opened 2022-05-09
Send funding flag

asmega:
### Context

- https://dfedigital.atlassian.net/browse/CN-133
- Send the new funding flag to ECF on submission
- The other half of this PR can be found at https://github.com/DFE-Digital/early-careers-framework/pull/2052
- May want to review commit by commit as there is refactor commit that just namespaces a load of classes

- On submission we now send over the new funding flag to ECF

### Guidance to review

- Fire up both NPQ and ECF apps locally
- Configure them to point at each with needed tokens
- Submit an application that hits criteria for new funding flag
- ECF should now have the application with funding flag set to true


Lockyy (approved):
Looks good to me, I took a look at this commit by commit to make sure the namespacing and functional changes made sense separately. :)

PR #395, opened 2022-05-17
[CN-99] Add funding ineligibility flow

Lockyy:
### Context

We want to inform people that they are ineligible for funding.

This adds in a pathway through the app that informs the user that they are not eligible for funding and what conditions need to be met to achieve funding. It doesn’t specify exactly which condition they validated, but it should help.

Along with that, the “where do you work” page has been moved forwards to occur at the same point in the flow as the ineligible for funding pages as it largely fulfils the same role.

For now all applicants with a PrivateChildcareProvider institution will return funded? as true. This is a temporary measure, for now they will all be approved. This will be adjusted so they reflect the actual rules in subsequent PRs.

### Guidance to review

The ineligible for funding pages within the flowchart cannot all be reached yet as two aspects of the funding are not implemented, cohort 2 and an EY applicant not doing an EY NPQ. Other than this though, the two other NPQ ineligibility pages should be reachable. Along with that the tell us about where you work page should be reached as described in the Lucid flowchart attached to the epic.

To review copy for this page you can visit these paths to review each ineligibility message directly:
- `/registration/ineligible-for-funding?registration_wizard[version]=school/has_already_been_funded`
- `/registration/ineligible-for-funding?registration_wizard[version]=school/outside_catchment_or_ineligible_establishment`
- `/registration/ineligible-for-funding?registration_wizard[version]=early_years/outside_catchment_or_not_on_early_years_register`
- `/registration/ineligible-for-funding?registration_wizard[version]=early_years/not_applying_for_NPQEY`

### DIFF
--- a/app/controllers/registration_wizard_controller.rb
+++ b/app/controllers/registration_wizard_controller.rb
@@ -2,7 +2,8 @@ class RegistrationWizardController < ApplicationController
   def show
     @wizard = RegistrationWizard.new(current_step: params[:step].underscore,
                                      store: store,
-                                     request: request)
+                                     request: request,
+                                     params: wizard_params)
 
     @form = @wizard.form
     @form.flag_as_changing_answer if params[:changing_answer] == "1"
--- a/app/lib/forms/aso_new_headteacher.rb
+++ b/app/lib/forms/aso_new_headteacher.rb
@@ -44,6 +44,7 @@ module Forms
       Services::FundingEligibility.new(
         course: course,
         institution: institution,
+        inside_catchment: wizard.query_store.inside_catchment?,
         new_headteacher: new_headteacher?,
       ).funded?
     end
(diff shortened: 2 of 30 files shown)


app/controllers/registration_wizard_controller.rb
```diff
@@ -2,7 +2,8 @@ class RegistrationWizardController < ApplicationController
   def show
     @wizard = RegistrationWizard.new(current_step: params[:step].underscore,
                                      store: store,
-                                     request: request)
+                                     request: request,
+                                     params: wizard_params)
```
Lockyy, line 6: This is to temporarily allow the versions param to be passed through into the forms via a GET param. That way we can override the genuine page that would otherwise be rendered. This will be removed before launch, it is a temporary measure for copy testing.

leoapost (commented):
LGTM. Do we need to wait for the pending copy?

leoapost (approved):

PR #399, opened 2022-05-18
[CN-97] Update the copy in the funding your npq page

leoapost:
### Context

We want to update the copy in the funding your npq page

Updated the copy based on the latest lucid flows.

I could see the same copy in the lucid flows for the `funding your aso` page too, so both pages were updated to keep the answers consistent.

### DIFF
--- a/app/lib/forms/funding_your_npq.rb
+++ b/app/lib/forms/funding_your_npq.rb
@@ -26,11 +26,11 @@ module Forms
 
     def options
       [
-        (option("school", "My school or college is covering the cost", link_errors: true) if works_in_school?),
+        (option("school", "My workplace is covering the cost", link_errors: true) if works_in_school?),
         (option("trust", "My trust is paying") if works_in_school? && inside_catchment?),
         (option("employer", "My employer is paying") if !inside_catchment? || !works_in_school?),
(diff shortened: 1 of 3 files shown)


app/lib/forms/funding_your_npq.rb
```diff
@@ -26,11 +26,11 @@ def course
 
     def options
       [
-        (option("school", "My school or college is covering the cost", link_errors: true) if works_in_school?),
+        (option("school", "My workplace is covering the cost", link_errors: true) if works_in_school?),
         (option("trust", "My trust is paying") if works_in_school? && inside_catchment?),
         (option("employer", "My employer is paying") if !inside_catchment? || !works_in_school?),
```
leoapost, line 31: I couldn't see this option in the latest lucid flow, but it looked to me like we should keep it.

Lockyy: That's a good point, would you be able to run that past James and see what he thinks?

leoapost: James mentioned we should check with Dom, but we both feel it's better to leave it for now and remove it later if we need to.

Lockyy (approved):

PR #415, opened 2022-05-30
Various bug fixes

Lockyy:
### Context

Fixes some bugs that're coming down from reviewers.

- [CN-195] Skip EY flow if applicant says they work in a school, regardless of their inside catchment status.
- [CN-72] Remove ASO related redirection in choose_your_npq, now checks for EHCO and redirects to EHCO flows.
- Remove /registration/aso-funding-contact page, nowhere links to it anymore.
- Update choose_your_provider to redirect back based on funding eligibility rather than specific question answers.
- Remove several references to ASO in copy and locales.
- Update seeds to hide ASO.


app/lib/forms/choose_your_npq.rb
```diff
@@ -17,8 +17,8 @@ def next_step
       if changing_answer?
         if no_answers_will_change?
           :check_answers
-        elsif course.aso?
-          :about_aso
```
Lockyy, line 21: This page doesn't exist and nobody will be able to register for EHCOs, this should've been swappe over already.

app/lib/forms/choose_your_provider.rb
```diff
@@ -22,10 +22,12 @@ def previous_step
...
-      elsif course.aso? && wizard.store["aso_funding"] == "yes"
-        :funding_your_aso
-      elsif wizard.store["aso_new_headteacher"] == "yes"
-        :aso_possible_funding
+      elsif course.ehco?
+        if eligible_for_funding?
+          :aso_possible_funding
+        else
+          :funding_your_aso
+        end
```
Lockyy, line 30: This section was duplicating eligibility funding logic to determine whether the user came from the eligible or ineligible path, if we leverage eligible_for_funding we can remoev this duplication and improve accuracy of this calculation.

db/seeds.rb
```diff
@@ -1,41 +1,26 @@
...
+    { position: 9, name: "Early Headship Coaching Offer", ecf_id: "0222d1a8-a8e1-42e3-a040-2c585f6c194a", description: "The Early Headship Coaching Offer is a package of structured face-to-face support for new headteachers.", display: true },
   ].each do |hash|
-    Course.find_or_create_by!(name: hash[:name], ecf_id: hash[:ecf_id], description: hash[:description])
+    course = Course.find_or_initialize_by(ecf_id: hash[:ecf_id])
+    course.update!(
+      name: hash[:name],
+      description: hash[:description],
+      position: hash[:position],
+      display: hash[:display],
+    )
```
Lockyy, line 20: Collapsing this seed file into one section with a find_or_initialize_by and update! speeds this up ever so slightly but more importantly mean that we won't create duplicate records when data changes.

Normally seeds of course aren't run all the time but they're actually run before specs in the npq_application so keeping them slim and updating existing records rather than duplicating is important.

StupidCodeFactory, line 20: suggestion:
```ruby
course.update!(hash.slice(:name, :description, :position, :display))
```

StupidCodeFactory (approved):

db/seeds.rb
Lockyy: That would be an improvement. Next time I'm in the seeds file I'll make that change.

PR #416, opened 2022-05-30
[CN-92] Update LeadProvider lists

Lockyy:
### Context

Lead Provider lists need updating to include new LeadProvider and correctly list providers only for the NPQs they provide.

The lists as they appear in this PR are still pending final confirmation. We're going to go ahead with this info and then will update as we get better data.

- Alphabetise LeadProvider lists on choose_your_provider page
- Add specific lists for further NPQs.
- Add National Institute of Teaching LeadProvider. See https://github.com/DFE-Digital/early-careers-framework/pull/2176.

### Guidance to review

Run through flow choosing differing NPQs and see that provider lists match lists in https://dfedigital.atlassian.net/browse/CN-92.


app/models/lead_provider.rb
```diff
@@ -1,24 +1,79 @@
...
-    else
-      all
-    end
-  end
+    course_specific_list = COURSE_TO_PROVIDER_MAPPING[course.name]
+
+    return all if course_specific_list.blank?
+
+    ecf_ids = ALL_PROVIDERS.slice(*course_specific_list).values.compact_blank
+    raise "Missing provider ECF_ID for available providers list" if ecf_ids.count != course_specific_list.count
```
Lockyy, line 75: Guards against providers in the COURSE_TO_PROVIDER_MAPPING that aren't in the ALL_PROVIDERS hash.

StupidCodeFactory (approved):

PR #423, opened 2022-06-01
Reject funded NPQs

leoapost:
### Context

[Ticket CN-22](https://dfedigital.atlassian.net/browse/CN-22)

We want users to not be eligible for scholarship funding for the same NPQ, if they have already received one in the past.

- Added the NpqFunding class to handle the api call
- Added a custom parser for the NpqFunding to parse the response from the ECF API
- Update the ineligibility copy
- Add an ineligibility page for the EHCO NPQ (a.k.a ASO)
- Update the funding eligibility service to require a TRN
- Added course mappings for the ECF course identifiers

### Guidance to review
The previously funded eligibility check is institution agnostic - will run for Schools, Early Years etc.

Also, the npq funding endpoint in the ECF API requires a TRN and a course identifier, but the NPQ service only knows the `ecf_id`. It make sense to update the ECF API at some point to require the `ecf_id` instead, but for now I have added a mapping in the `Course` model to resolve the `identifier` based on the `ecf_id`.

#### How to test
Run NPQ and ECF services on your local and go through an eligible journey. Choose users and NPQs that were funded in the past and then check the copy in the ineligibility page is correct.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -12,6 +12,19 @@ class Course < ApplicationRecord
     EHCO: "Early Headship Coaching Offer",
   }.with_indifferent_access.freeze
 
+  COURSE_ECF_ID_TO_IDENTIFIER_MAPPING = {
+    "15c52ed8-06b5-426e-81a2-c2664978a0dc" => "npq-leading-teaching",
+    "7d47a0a6-fa74-4587-92cc-cd1e4548a2e5" => "npq-leading-behaviour-culture",
+    "29fee78b-30ce-4b93-ba21-80be2fde286f" => "npq-leading-teaching-development",
+    "a42736ad-3d0b-401d-aebe-354ef4c193ec" => "npq-senior-leadership",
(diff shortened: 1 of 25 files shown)


app/models/course.rb
```diff
@@ -12,6 +12,19 @@ class Course < ApplicationRecord
     EHCO: "Early Headship Coaching Offer",
   }.with_indifferent_access.freeze
 
+  COURSE_ECF_ID_TO_IDENTIFIER_MAPPING = {
```
asmega, line 15: you should already have these mappings in the database on `Course` so you're duplicating here. you'd only want to want to write this if you wanted to to save yourself the DB query

asmega: The alternative would be shift everything in the database to in memory. i recommend one strategy and not have both in place

leoapost: I couldn't find the identifier in the NPQ app and wanted to avoid any db columns as I was planning to see if we can replace the `identifier` in the ECF API with the `ecf_id`.

PR #424, opened 2022-06-01
[CN-200] Invalid lead provider clearing

Lockyy:
### Context

There was a workaround within the system that allowed you to apply for an NPQ with a LeadProvider that didn't supply that NPQ. If you selected an NPQ they did provide, select the provider and then go back and change your NPQ, then you could submit with an invalid lead provider.

- Check LeadProvider on the choose_your_npq form object after save and clear the lead provider if it is no longer valid.
- Check for if LeadProvider is invalid during choose_your_npq next step to ensure user is asked to continue the flow from there and select a new provider.

### Guidance to review

1. Select NPQLBC
2. Progress to Choose Your Provider page
3. Select "Best Practice Network (home of Outstanding Leaders Partnership)" provider
4. Progress to Check answers page
5. Choose to change your NPQ
6. Select NPQLTD
7. See that you return to check answers screen because LeadProvider remains valid  
8. Choose to change your NPQ
9. Select NPQEYL
10. See that you have not returned to check answers screen and are back in the flow and need to reselect your provider

### DIFF
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -13,8 +13,19 @@ module Forms
       ]
     end
 
+    def after_save
+      return if lead_provider_valid?
+
+      wizard.store["lead_provider_id"] = nil
+    end
+
     def next_step
-      if changing_answer?
+      # If your lead provider remains valid we can progress down the changing answer path
+      # as it is fine for us to end up going back to the check_answers page.
+      # If it is no longer valid due to the NPQ changing though we will need to be
+      # reinserted back into the flow so that later on the user can be asked to
+      # choose a new provider.
+      if changing_answer? && lead_provider_valid?
         if no_answers_will_change?
           :check_answers
         elsif course.ehco?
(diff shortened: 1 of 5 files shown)


app/lib/forms/choose_your_npq.rb
```diff
@@ -74,6 +85,14 @@ def course
 
   private
 
+    def lead_provider_valid?
```
asmega, line 88: could this not have been added as an active model validation on the form?

asmega (approved):

app/lib/forms/choose_your_npq.rb
Lockyy: It's not really a validation, it's more checking a status to direct behaviour of the form.

I suppose we could throw an error message but then we're telling the user to go change their provider and come back to change their NPQ which is clunky and in certain circumstances may not make sense or be easily manageable. Going down this path of having it just clear out the data and allowing the user to progress throws less roadblocks in front of progression.

PR #428, opened 2022-06-06
[CN-47] Add rake task for sending registration opening notifications

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CN-47

Adds rake task to send notification emails

### Guidance to review

1. Create RegistrationInterest locally
2. Run rake task with count of at least 1
3. Check that RegistrationInterest has been locally marked notified
4. Check Notify to see that email has been sent


lib/tasks/registration_interest.rake
```diff
@@ -0,0 +1,10 @@
+namespace :registration_interest do
+  desc "Send notifications to users indicating NPQ applications are open"
+  task :send_notifications, %i[count] => :environment do |_t, args|
+    count = args["count"].to_i
+
+    RegistrationInterest.order("RANDOM()").first(count).each do |registration_interest|
```
leoapost, line 6: How about we only pull records that have not been notified already?

Lockyy: Good spot

leoapost (approved):

PR #447, opened 2022-06-17
[CN-275] Fix how works_in_private_childcare_provider? is determined and improve data integrity protections

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CN-275

There is a bug where if you say you work in childcare but not in a nursery you will be treated as if you work in a public nursery instead of a private nursery despite having been directed to select a private childcare provider.

This is occurring because the code that determines whether you are working for a private childcare provider specifically checks if you work in a nursery and your “kind of nursery” is set to “private nursery”. This is incorrect as there are alternative routes that allow you to reach the private childcare provider selector.

There are several implications of this bug:
- The possibly funded page will render incorrectly as it believes you do not work in a private childcare provider 
  app/lib/forms/possible_funding.rb:L16
- Private Childcare Provider information will not be displayed on check answers page
  app/lib/services/handle_submission_for_store.rb:L91
- Your private childcare providers URN will not be stored in the database when you submit your application and so will be lost and not sent to ECF.
  app/lib/services/handle_submission_for_store.rb:L66

The last of these is the biggest issue. This means that for anyone who has gone down the following path will have had lot information and we won’t have stored their private childcare provider data:

1. Say you work in England 
2. Say you do not work in a school
3. Say you work in childcare
4. Say you do not work in a nursery
5. Say you have an Ofsted URN
6. Enter your Ofsted URN
7. Finish your application

- Improve all end to end tests, adding explicit checks for *all* stored data, on the created Application record and on the User record. This is an improvement over the previous system that only checked some of the data that was deemed relevant whilst writing the spec.
- Add spec for missing non-nursery private childcare routes. This, on top of the previous change, demonstrates the CN-275 bug, failing without the fix.
- Resolve other issues highlighted by the stricted data checking.
  - Do not store Employer and Work details unless the conditions during application match those that lead to Employer data being presented during the Check Answers step.
  - Clear user entered country during the Teacher Catchment step if the user returns to this step and selects a radio option.
- Store raw application data to ensure applicant data cannot be lost after they click submit.

#### Before

1. Say you work in England 
2. Say you do not work in a school
3. Say you work in childcare
4. Say you do not work in a nursery
5. Say you have an Ofsted URN
6. Enter your Ofsted URN
7. Reach Check Answers screen.
8. You will not see your selected private childcare provider
9. Submit application
10. Check database and see that private childcare provider is not stored  

#### After

1. Say you work in England 
2. Say you do not work in a school
3. Say you work in childcare
4. Say you do not work in a nursery
5. Say you have an Ofsted URN
6. Enter your Ofsted URN
7. Reach Check Answers screen.
8. You should see your selected private childcare provider
9. Submit application
10. Check database and see that private childcare provider is stored


app/lib/services/query_store.rb
```diff
@@ -39,8 +39,7 @@ def works_in_public_childcare_provider?
   end
 
   def works_in_private_childcare_provider?
-    works_in_nursery? &&
-      Forms::KindOfNursery::KIND_OF_NURSERY_PRIVATE_OPTIONS.include?(store["kind_of_nursery"])
+    works_in_childcare? && !works_in_public_childcare_provider?
```
Lockyy, line 42: This is the core fix to CN-275. It makes it so that if you say you work in childcare and go down any path other than saying you work in a public nursery, you'll be treated like your work in a private childcare provider, which is how the questionnaire flow works.

spec/features/happy_journeys_spec.rb
```diff
@@ -1256,24 +1771,326 @@
...
+        "trn_verified" => true,
+        "verified_trn" => "1234567",
+        "works_in_childcare" => "yes",
+        "works_in_nursery" => "yes",
+        "works_in_school" => "no",
+      },
+    )
+  end
+
+  scenario "registration journey while working at private childcare provider but not a nursery" do
```
Lockyy, line 1866: This is the spec that I added to red/green test the CN-275 bug.

PR #454, opened 2022-06-28
new ehco applications applied 2021 cohort

asmega:
### Context

- https://dfedigital.atlassian.net/browse/CPDLP-1472
- The business and providers and EHCO declarations to be paid for ASAP
- Therefore we will apply EHCO to the 2021 cohort so they can be paid out before the 2022 cohort starts

- All new EHCO applications will be applied the 2021 cohort
- Everything else remains the same with 2022 cohort

### Guidance to review

- none

### DIFF
--- a/app/lib/services/handle_submission_for_store.rb
+++ b/app/lib/services/handle_submission_for_store.rb
@@ -39,7 +39,7 @@ module Services
           works_in_nursery: store["works_in_nursery"] == "yes",
           works_in_childcare: store["works_in_childcare"] == "yes",
           kind_of_nursery: store["kind_of_nursery"],
-          cohort: Application::TARGET_COHORT,
+          cohort: course.default_cohort,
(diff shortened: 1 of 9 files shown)


app/lib/services/handle_submission_for_store.rb
```diff
@@ -198,5 +198,13 @@ def school
     def user
       @user ||= User.find_by(email: store["confirmed_email"])
     end
+
+    def target_cohort
+      if course.ehco?
+        2021
+      else
+        Application::DEFAULT_COHORT
+      end
```
Lockyy, line 207: This feels like it should be something being returned by the course, rather than being calculated inside service.

asmega: it's now a field/column on the course

Lockyy (approved):

PR #464, opened 2022-07-14
Refactor the journey feature specs

peteryates:
### Context

~~Note this follows on from #463~~

Related to https://dfedigital.atlassian.net/browse/CN-312

####  Split happy_journeys_spec up into multiple files

The common behaviour at the top of `happy_journeys_spec.rb` has been shifted to the `JourneyHelper` module and a shared context that overrides the retrieval of the latest application data.

#### Begin to remove repeated code from journey specs

The specs are repetitive in some areas and not always specifying the current page makes them a difficult to follow.

This change adds the `JourneyAssertionHelper` that provides us with methods

* `expect_page_to_have`
* `navigate_to_page`
* `expect_check_answers_page_to_have_answers`

This allows us to remove some repeated lines (axe checking) and gives the test some hierarchy (via blocks) making it easier to see which assertions happen on which page.

The actual test logic hasn't changed.

### Guidance to review

This is going to be quite tricky to review because there's a lot of it and it's very repetitive 😅 

The final assertions in each scenario haven't changed, just the structure of the steps above. For the most part this should prove the structural changes haven't changed what's actually happening.

The main that've changed are that:
* (nearly) every assertion or action on a page happens inside a block that tells us which page it's happening on
* the often repeated things like `click_button('Continue')` and `be_axe_valid`) are handled by the wrapping method

image

I don't think this is perfect, there's way more we can do, but a step in the right direction.


spec/support/helpers/journey_assertion_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def given_i_am_on_page(path, submit_form: true, submit_button_text: "Continue", axe_check: true, &block)
+      expect(page.current_path).to eql(path)
+
+      page_checks(axe_check:, &block)
+
+      page.click_button(submit_button_text) if submit_form
+    end
+
+    alias_method :now_i_should_be_on_page, :given_i_am_on_page
```
peteryates, line 21: Perhaps `now_i_should_be_on_page` could be better named - it doesn't really explain that we also will be ensuring the page is Axe valid and that we'll be submitting it. Couldn't come up with anything though.

Lockyy: The three of these don't fully encapsulate what is going on IMO, it's unclear that navigate_to_page will include expectations or execute a block. 

Given that navigate_to_page is mostly duplicating logic from given_i_am_on_page, I'd pitch removing the page_checks and submission parts of that method and instead expect the tester to call given_i_am_on_page.

I'd also propose updating the names of the latter two methods to use language that is more idiomatic of rspec+capybara, for example:
- given_i_am_on_page -> expect_page_to_have (and turn path into a keyword parameter)
- and_the_check_your_answers_page_should_contain -> expect_check_answers_page_to_have_answers

Example calls:
```
expect_page_to_have(path: "/registration/choose-school", submit_form: true) do
  expect(page).to have_text("Choose your workplace")

  within ".npq-js-reveal" do
    page.fill_in "Enter the name of your workplace", with: "open"
  end

  expect(page).to have_content("open welsh school")

  page.find("#school-picker__option--0").click
end
```
Note that I've made submit_form false by default here so that you have to be explicit about taking an action, rather than have to suppress the behaviour.

```
expect_check_answers_page_to_have_answers(
  "cohort" => 2022,
  "course_id" => Course.find_by_code(code: :NPQSL).id,
  ...
)
```

Lockyy (approved):

PR #475, opened 2022-07-25
CN-312 change from 'do you work in a school?' to 'what setting do you work in?'

peteryates:
### Context

Follows on from #464

[CN-312](https://dfedigital.atlassian.net/browse/CN-312)

The spec in the JIRA expects:

> Rather than having a yes/no question, to avoid confusion, we’ve translated the H1 into radios.
>
> - If a user clicks on `a school`, `a academy trust`, `a 16-19 organisation` they will:
>   1. be shown the screen where it’s assumed they’ll have a TRN 
>   2. go down the GIAS route
> 
> - If a user clicks on `early years` they will:
>   1. shown the screen where it’s assumed they don’t have a TRN 
>   2. will go down the Ofsted route
> 
> - If a user clicks on `other` they will:
>   1. shown the screen where it’s assumed they don’t have a TRN 
>   2. will go down the ‘Non-school user route’. (Tweaks to the other route will be another card)
>   3. once they have chosen their NPQ, they will see the “Tell us about where you work” and “How are you paying page” - These will eventually be removed as part of updates to the ‘Non-school user route’.

In addition to the requirements I renamed the page from `work-in-school` to `work-setting`, to keep the filenames and contents aligned and easy to follow.

The 'do you work in early years or childcare' screen has also been dropped as it's covered by the 'work setting' question.

The change was a bit more involved than expected because a more fundamental question is asked early on and had several repercussions through the rest of the journey. This is the reason this change was held off until #464 was ready.

### Guidance for reviewing

I'd recommend going through the commits one by one. Most of the changes are updating other feature specs so they're aware of the page rename and early years question removal.


app/lib/forms/work_setting.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def after_save
+      # we are inferring `works_in_school` and `works_in_childcare` to maintain
+      # consistency with older records
+
+      case work_setting
+      when "a_school", "an_academy_trust", "a_16_to_19_educational_setting"
+        wizard.store["works_in_school"] = "yes"
+        wizard.store["works_in_childcare"] = "no"
+        wizard.store["works_in_nursery"] = nil
```
Lockyy, line 21: This is such a nice side-effect of these changes. That having these two major branches determined in one place means we can reliably clean up data related to related answers.

There's also kind_of_nursery and has_ofsted_urn that we could clean up in situations where the user says they work in a school. And off of that we could possibly clear institution_name and institution_identifier if we check that the work_setting has changed from one to the other.

app/lib/forms/work_setting.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    validates :work_setting, presence: true, inclusion: { in: VALID_WORK_SETTING_OPTIONS }
+
+    def self.permitted_params
+      %i[work_setting]
+    end
+
+    def after_save
+      # we are inferring `works_in_school` and `works_in_childcare` to maintain
+      # consistency with older records
```
Lockyy, line 15: Should this after_save have a guard clause that exits if work_setting hasn't changed? Is it possible to detect that?

spec/lib/forms/work_setting_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        "works_in_school" => "no",
+        "works_in_childcare" => "yes",
+        "works_in_nursery" => nil,
+      },
+      "other" => {
+        "works_in_school" => "no",
+        "works_in_childcare" => "no",
+        "works_in_nursery" => nil,
+      },
+    }.each do |option, expectations|
```
Lockyy, line 42: This is great 😁

app/lib/forms/dqt_mismatch.rb
```diff
@@ -13,8 +13,10 @@ def next_step
         else
           :choose_your_npq
         end
+      elsif wizard.query_store.inside_catchment? && wizard.query_store.works_in_nursery?
```
Lockyy, line 16: Should this be works_in_childcare?

app/models/registration_wizard.rb
```diff
@@ -95,31 +95,29 @@ def answers
...
-                     OpenStruct.new(key: "Ofsted registration details",
-                                    value: institution_from_store.registration_details,
-                                    change_step: :have_ofsted_urn)
-                   else
-                     OpenStruct.new(key: "Do you have a URN?",
-                                    value: store["has_ofsted_urn"].capitalize,
-                                    change_step: :have_ofsted_urn)
-                   end
-        end
+    if inside_catchment? && query_store.works_in_childcare? && !query_store.works_in_school?
```
Lockyy, line 98: Since works_in_school and works_in_childcare are exclusive now, can !query_store.works_in_school? be removed?

Lockyy (commented):
This is looking really good, I've just left a few comments to discuss before approval.

Overall this is such a good change for the codebase and the flow in terms of being able to reliably know which path a user is on. I knew when this was being planned that this would make managing the branches easier but I am still excited for this to be in place.

Lockyy:
One other thing, should we follow this PR up with pushing the new work_setting data through to the CPD database?

app/models/registration_wizard.rb
peteryates: Yep, good call.

app/lib/forms/work_setting.rb
peteryates: I did have a look but couldn't see a tidy way of doing it, other than recording a list of modified fields inside `#save!`

app/lib/forms/work_setting.rb
peteryates: Ah yeah - I wondered about that but it looked like them still being set didn't have any side effects, even if they're unused.

peteryates: I've pushed I've pushed [1b36056](https://github.com/DFE-Digital/npq-registration/pull/475/commits/1b3605618ced556b3a01180956c57bdea1a9c317) that also removes `kind_of_nursery` and `has_ofsted_urn` - but it removes the keys rather than sets them as `nil`. This should make testing a bit more straightforward. that also removes `kind_of_nursery` and `has_ofsted_urn` - but it removes the keys rather than sets them as `nil`. This should make testing a bit more straightforward.

app/lib/forms/dqt_mismatch.rb
```diff
@@ -5,19 +5,14 @@ def previous_step
...
-      else
-        :choose_your_npq
+      return :check_answers if changing_answer?
+
+      if wizard.query_store.inside_catchment?
+        return :find_school if wizard.query_store.works_in_school?
+        return :work_in_nursery if wizard.query_store.works_in_childcare?
       end
+
+      :choose_your_npq
```
Lockyy, line 15: This is a big improvement to readability.

app/lib/forms/work_setting.rb
Lockyy: I think we'd have to add in [ActiveModel::Dirty](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html) to get that functionality brought across from ActiveRecord. 

That'd be a lot of work though and not worth the time implementing.

Lockyy (approved):

PR #476, opened 2022-07-26
Add FAQs

peteryates:
### Ticket

https://dfedigital.atlassian.net/browse/CN-424

### Context and changes

There are now three pages of FAQs:

* early years
* schools
* other

Some of the questions are shared between sections so they're all stored in the same directory and referenced by partial. The approach was taken (rather than just hardcoding) to ensure contents list links always match the heading ids.

The CSS for the contents list [was lifted from GOV.UK Publishing Components](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss).

### DIFF
--- a/app/views/pages/faqs/_contents-list.html.erb
+++ b/app/views/pages/faqs/_contents-list.html.erb
@@ -1,0 +1,11 @@
+<nav class="contents-list" aria-label="Frequently asked questions">
+  <h2 class="contents-list__title">Contents</h2>
+
+  <ul class="contents-list__list">
+    <% questions_and_answers.each_key do |question| %>
+      <li class="contents-list__list-item">
+        <%= govuk_link_to(question, "#" + question.parameterize, class: "contents-list__link", no_visited_state: true) %>
+      </li>
+    <% end %>
+  </ul>
+</nav>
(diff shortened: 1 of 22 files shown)


Lockyy (changes requested):
The partials for this are using hyphens instead of underscores in their names as separators. We should swap this over to make sure things are consistent across the app and match with Rails' conventions.

On a related note, @asmega do you have any info on why we went down the hyphen path for URLs @asmega? Is it for SEO? (It was added [here](https://github.com/DFE-Digital/npq-registration/commit/28e77c9336e5c800f734c0d7a94bf2285257b6ee)).

peteryates:
I think dashes are easier to read than underscores in URLs, I'm in favour of using them on public-facing stuff. GOV.UK definitely favour them too ([example](https://www.gov.uk/browse/abroad/travel-abroad)).

Agree it makes sense in the partial filenames to standardise on underscores though.

Lockyy (approved):
I definitely agree on using them for URLs, I'm just curious as to what prompted the switch. 🙂

asmega:
my understanding is in general hyphens in url paths are preferable hence the change

Lockyy:
@peteryates @DominiqueSherwood1 Are we happy to move ahead on merging and releasing this?

peteryates:
Once Dom gives the goahead I'll get it merged.

PR #481, opened 2022-07-28
Add admin page for uploading and downloading manual validation files

peteryates:
### Context

This is a proof of concept that will allow admins to download CSV files containing details of people who need to be manually validated and then upload the outcomes.

No tests etc as of yet, will follow if this is deemed useful.

| Before | After |
| ------ | ------- |
| Screenshot from 2022-07-28 16-53-53 | Screenshot from 2022-07-28 16-54-11 |

### Final steps

If we go ahead we'll need

* [x] some kind of indicator as to which records were skipped in the flash message
* [x] tests
* [x] maybe some guidance on the page


peteryates:
Now with some guidance on the upload format and information about the number of updated records.

| Upload guidance | Upload summary |
| ------------- | ------------------|
| Screenshot from 2022-08-01 18-35-35 | Screenshot from 2022-08-01 18-35-46 |

peteryates:
I'm closing this as I don't think it's the right approach. The NPQ record should be left in tact as a record of what was entered during registration. Any subsequent changes to the applicant's details should be made in ECF.

PR #499, opened 2022-08-17
Solve missing interpolation argument bug when validating institution identifiers

peteryates:
### Context

[This bug](https://sentry.io/organizations/dfe-teacher-services/issues/3294833521/?project=5817541) has popped up a few times in Sentry over the last few months. The error message is

```
I18n::MissingInterpolationArgument: missing interpolation argument :urn in "No private
childcare providers with the URN %{urn} were found, please try again" ({:model=>
"Choose private childcare provider", :attribute=>"Institution identifier", :value=>
"School-123456"} given)
```

It happens because the `:invalid` error message in the locales file expects a `urn:` argument, which it received from the custom error message but not from the validates format macro.

- Add choose private childcare provider step specs
- Move institution identifier validation to method

### DIFF
--- a/app/lib/forms/choose_private_childcare_provider.rb
+++ b/app/lib/forms/choose_private_childcare_provider.rb
@@ -4,10 +4,9 @@ module Forms
 
     attr_accessor :institution_name, :institution_identifier
 
-    validates :institution_identifier, format: { with: /\APrivateChildcareProvider-.+\z/, unless: -> { institution_identifier.blank? || institution_identifier == "other" } }
     validates :institution_name, length: { maximum: 64 }
 
-    validate :validate_institution_exists
+    validate :validate_institution_identifier
     validate :validate_private_childcare_provider_name_returns_results
 
     def self.permitted_params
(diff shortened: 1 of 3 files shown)


app/lib/forms/choose_private_childcare_provider.rb
```diff
@@ -4,10 +4,9 @@ class ChoosePrivateChildcareProvider < Base
 
     attr_accessor :institution_name, :institution_identifier
 
-    validates :institution_identifier, format: { with: /\APrivateChildcareProvider-.+\z/, unless: -> { institution_identifier.blank? || institution_identifier == "other" } }
```
Lockyy, line 7: Do we need to do this in Forms::ChooseSchool and Forms::ChooseChildcareProvider?

peteryates: I don't believe so, in the locales it's that the generic `invalid` is paired up with `%{urn}` that causes the issue.

```yml
invalid: No private childcare providers with the URN %{urn} were found, please try again
```

There probably is scope to make sure the generic error messages are nicely worded but I think that should be a separate piece of work.

Lockyy (approved):

Lockyy:
@peteryates Can this be merged after rebasing?

peteryates:
@Lockyy ah yeah - sorry this dropped off my radar. Will do it now!

PR #503, opened 2022-08-19
change straight quotes to curly quotes

EmmaFrith:
This quote should be curly:

Screenshot 2022-08-19 at 15 37 46

Like this:

Screenshot 2022-08-19 at 15 39 02


peteryates (approved):
Looks good to me.

PR #504, opened 2022-08-19
this p should not be capitalised

EmmaFrith:
Make the 'P' in 'provider' lower case:

Screenshot 2022-08-19 at 16 11 35


Lockyy (approved):
Nice catch

PR #505, opened 2022-08-22
[CPDNPQ-362] Split up registration/your-role page

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CPDNPQ-362

### Changes proposed in this pull request

- Split registration/your-work into three pages:
  - registration/york-employment
  - registration/york-role
  - registration/york-employer
- Move "Your Work" flow before registration/choose-your-npq
- Direct non-school users to registration/choose-your-provider after selecting NPQ.

I've tried something new in this PR with how the forms are implemented. I've used only shared partials for the forms themselves to try and move us away from repeatedly reimplementing the same page format. Since ideally every page will only have one question and follow the same formatting I hope that this structure will work for new pages going forward and for most existing pages as well.

### Guidance to review

Running through the non-school journey should see the three new your work pages moved ahead of the choose your npq flow and skip right to choose your provider page selecting your NPQ.

See ticket on Jira for expected flow.

PR #512, opened 2022-08-25
Refactor portion of radio type questions to use govuk-form-builder localisation

Lockyy:
### Context

https://github.com/DFE-Digital/npq-registration/pull/505 began the process of moving towards leveraging the built in govuk-form-builder localisation, this PR is a follow up to continue that work.

It also set up a generic _question_page template that could be used to fully abstract how a page is rendered away from the form using it. I've used that in a few of the updated questions here as well.

Further PRs will be required to get this all fully swapped over.

- Upgrade the following pages to use govuk-form-builder's built in localisation
  - /registration/aso-headteacher
  - /registration/work-in-nursery
  - /registration/aso-new-headteacher
  - /registration/aso-headteacher
  - /registration/have-ofsted-urn
  - /registration/funding-your-npq
  - /registration/work-setting
  - /registration/kind-of-nursery
  - /registration/npqh-status
  - /registration/funding-your-aso
  - /registration/teacher-catchment
  - /registration/work-in-nursery
- Upgrade the following to use the shared question_page partial
  - /registration/funding-your-npq
  - /registration/teacher-catchment

### Guidance to review

Go through the above paths and make sure that they are identical to their appearance before the change.

### DIFF
--- a/app/lib/forms/base.rb
+++ b/app/lib/forms/base.rb
@@ -90,7 +90,7 @@ module Forms
         value:,
         link_errors:,
         divider:,
-        revealed_question:
+        revealed_question:,
       )
     end
   end
--- a/app/lib/forms/base.rb
+++ b/app/lib/forms/base.rb
@@ -83,17 +83,15 @@ module Forms
       wizard.query_store
     end
 
-    def option_locale_scope
-      raise NotImplementedError
-    end
+    def build_option_struct(value:, link_errors: false, divider: false, include_if: true)
+      return unless include_if
 
-    def build_option_struct(value:, link_errors: false, divider: false)
       options = {
         value:,
         link_errors:,
       }
 
-      options[:divider] = divider if divide
+      options[:divider] = divider if divider
 
       OpenStruct.new(options)
     end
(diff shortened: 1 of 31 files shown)


app/lib/forms/base.rb
```diff
@@ -82,5 +82,16 @@ def reset_store!
     def query_store
       wizard.query_store
     end
+
+    def build_option_struct(value:, link_errors: false, divider: false, include_if: true, revealed_question: nil)
+      return unless include_if
```
peteryates, line 87: The param name `include_if` threw me a bit here. Until I looked at the code I wasn't sure what was being included.

Does it need to be passed in? Would conditionally adding it to the array work instead?

```
build_option_struct(value: "school", link_errors: true),
build_option_struct(value: "trust") if works_in_school? && inside_catchment?,
build_option_struct(value: "self"),
```

Or perhaps it could benefit from a name change (`render`?)

peteryates (approved):
Looks great. Left one minor comment about the param `include_if` but I don't feel strongly about it.

PR #515, opened 2022-08-30
Add deploy tagging to dev auto-deploy

Lockyy:
Testing out auto-tagging during action deployments


peteryates (approved):
Looks good. Probably worth trimming the trailing lines but happy for that to wait if you're planning on adding any more to this flow.

PR #529, opened 2022-09-27
Add tracking pixels to the app

peteryates:
### Context

DfE needs to know how successful NPQ advertising is and how many people are being driven to the app from promotional material.

Pixels have been added for LinkedIn, Twitter and Meta (Facebook) to record this. The related CSP settings have been updated accordingly, we will need to test these in production once deployed.


app/views/layouts/application.html.erb
```diff
@@ -24,6 +24,10 @@
 
     <%= render partial: "shared/analytics/google" %>
 
+    <% if show_tracking_pixels? && !Rails.env.test? %>
```
Lockyy, line 27: Do we want this active in develop? `show_tracking_pixels? && Rails.env.production?` would prevent us from polluting tracking data with local development.

Lockyy (changes requested):

Lockyy (approved):

PR #563, opened 2022-11-10
[CPDNPQ-604] Integrate Get an Identity service into questionnaire flow

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CPDNPQ-604

- Adds Devise for user authentication
- Adds TRA omniauth integration
- Adds feature flag for swapping from regular user creation flow (start -> initial questions -> have trn? -> email -> user details) with TRA user creation flow (start -> have trn? -> GAI service -> back to questions)


app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,34 @@
+class FeatureFlagsController < ApplicationController
```
Lockyy, line 1: This was used to allow people to turn the GAI feature flag on/off during testing. Making a note to remove this before we merge in.

app/controllers/omniauth_controller.rb
```diff
@@ -0,0 +1,65 @@
+class OmniauthController < Devise::OmniauthCallbacksController
+  skip_before_action :verify_authenticity_token, only: [:tra_openid_connect]
+  before_action :remove_from_get_an_identity_pilot, only: :tra_openid_connect, if: :tra_response_excludes_trn?
```
Lockyy, line 3: Without a TRN the user can't finish their application, to resolve this we redirect them out of the pilot.

app/components/status_box.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      <% if url %>
+        <a class="govuk-link govuk-link--no-visited-state big-number-link" href="<%= url %>">
+      <% end %>
+        <%= number %> <%= label %>
+        <% if percentage %>
+          - <%= percentage %>%
+        <% end %>
+      <% if url %>
+        </a>
+      <% end %>
```
peteryates, line 13: Not one for now but I think this could be made a little easier to follow using a method or two on the component class:

Something along the lines of:

```ruby
if url
  govuk_link_to(box_content, url, no_visited_state: true, class: "big-number-link")
else
  box_content
end
```

app/lib/forms/teacher_reference_number.rb
```diff
@@ -12,17 +12,41 @@ def self.permitted_params
...
+    end
+
     def next_step
       case trn_knowledge
       when "yes"
-        :contact_details
+        # This should never happen, as the button that would lead here should be replaced
+        # in this scenario with a link to GAI. However, just in case we have this fallback to take
+        # you to an interstitial page that can only go to GAI.
+        # This can happen if JS is disabled.
```
peteryates, line 33: Should this open with "This should only happen if JS is disabled, ..."?

peteryates (approved):
Looks great. Left a few really minor suggestions here and there but nothing that needs immediate attention. Dashboard looks great 👌🏽

app/lib/forms/teacher_reference_number.rb
Lockyy: I’ll improve this comment. There are multiple ways to reach this state, they’re just obscure. I’ve catalogued them since writing this.

app/components/status_box.html.erb
Lockyy: Very good point, I'll follow up with a PR to move to that setup.

app/components/subnav_component/nav_item_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<li class="app-subnav__list-item<%= current_page?(path, check_parameters: true) ? " app-subnav__list-item--selected" : "" %>">
+  <%= govuk_link_to path, class: "app-subnav__link govuk-link--no-visited-state", "aria-current" => (current_page?(path) ? "page" : nil) do %>
```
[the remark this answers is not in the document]
Lockyy: Will move that over. This is as it is as it came directly from the ECF app.

PR #567, opened 2022-11-15
Use beginning of day for dashboard calculations

Lockyy:
### Context

Noticed that figures oscillated through the day due to timing of when previous applications were made.

Fix the "last 7 days" cut off to midnight.


peteryates (approved):
Makes sense

Lockyy:
Thanks @tvararu @peteryates

PR #568, opened 2022-11-15
Enhance error logging in omniauth controller

Lockyy:
### Context

To track failures we need proper error logging

Capture various errors in OmniauthController and send them to Sentry

### DIFF
--- a/app/controllers/omniauth_controller.rb
+++ b/app/controllers/omniauth_controller.rb
@@ -13,17 +13,43 @@ class OmniauthController < Devise::OmniauthCallbacksController
 
       sign_in_and_redirect @user
     else
+      send_error_to_sentry(
+        "Could not persist user after omniauth callback",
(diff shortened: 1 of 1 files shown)


peteryates (approved):

app/controllers/omniauth_controller.rb
```diff
@@ -54,6 +80,15 @@ def tra_response_excludes_trn?
   end
 
   def remove_from_get_an_identity_pilot
+    send_error_to_sentry(
+      "User removed from Pilot",
+      contexts: {
+        status: {
+          trn_blank: tra_response_excludes_trn?,
+        },
```
Lockyy, line 88: I've purposefully not included any user information here, just that it's happened and that it was because they were missing a TRN. This'll provide us a count of how many people this is impacting relative to the amount of people who make it through the pilot to complete an application.

PR #574, opened 2022-11-16
Prompt TRA to clear login when application created

Lockyy:
### Context

For testing, and shared computers, we want to make sure that if you complete an application people can't re-use your TRA login. 

Mark session as needing to tell TRA to use a fresh session when NPQ applications completed.


peteryates (approved):
Looks good

PR #578, opened 2022-11-18
Remove the unique index from `User#ecf_id`

peteryates:
## Context

I don't think this column needs to be unique. It tells us which user in the ECF application this record belongs to, but ECF users can have multiple email addresses (via their participant identities).

It causes problems when a person applies more than once with different email addresses. When submitting the NPQ application to ECF we hit an error caused the unique index on `users.ecf_id`, which was assigned to their original user record when they completed their first application.

The current solution is to manually 'move' the applications to their original user account and re-trigger the submission manually.

### Pros

* users don't need to worry about which of their email addresses they use when applying for an NPQ
* it simplifies our submission process and avoids any manual intervention

#### Cons

* we'll end up with some users in the NPQ database that share an `ecf_id`


leoapost (approved):
Sounds sensible

Lockyy (approved):

PR #579, opened 2022-11-18
Add an unsynced user list to the admin section

peteryates:
### Jira Ticket

https://dfedigital.atlassian.net/browse/CPDNPQ-656

### Context and changes

Unsynced users are users who have completed their application but we failed to link their record with an ECF user. This usually happens because they've used more than one email address and we already have their `ecf_id` on record.

| List of unsynced applications | Viewing an unsynced application |
| ----------- | ----------------- |
| image | image |

### DIFF
--- a/app/views/admin/unsynced_applications/show.html.erb
+++ b/app/views/admin/unsynced_applications/show.html.erb
@@ -1,13 +1,13 @@
 <% content_for :before_content do %>
   <%= render GovukComponent::BackLinkComponent.new(
     text: "Back",
-    href: admin_applications_path
+    href: admin_unsynced_applications_path
   ) %>
 <% end %>
 
 <div class="govuk-grid-row">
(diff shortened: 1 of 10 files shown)


app/views/admin/unsynced_applications/show.html.erb
```diff
@@ -0,0 +1,118 @@
...
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h1 class="govuk-heading-xl">Unsynced application</h1>
+
+    <%=
+      govuk_summary_list do |sl|
+
+        sl.row do |row|
```
peteryates, line 15: This currently duplicates what's in the regular application view. I just replicated it for now to demo, I'll probably combine them into a single partial

leoapost (approved):

PR #582, opened 2022-11-23
Fund teachers doing NPQEYL course in independent schools

leoapost:
### Context

Ticket: [CPDNPQ-646](https://dfedigital.atlassian.net/browse/CPDNPQ-646)

Teachers should be marked as eligible for funding if they select the NPQEYL course with certain schools.

- Add the `eyl_eligible` flag in schools table to indicate they offer funding for the NPQEYL course
- Update the funding eligibility service object to consider the eligibility of applications with such schools and NPQEYL course
- Add a rake task to mass-update the `eyl_eligible` flag in the schools we want to make them eligible for NPQEYL course. It takes a CSV path as argument pointing to a csv file with the urns (`gias_urn` column) of the schools to be updated.


db/schema.rb
```diff
@@ -184,6 +184,7 @@
     t.boolean "high_pupil_premium", default: false, null: false
     t.text "postcode_without_spaces"
     t.integer "number_of_pupils"
+    t.boolean "eyl_eligible", default: false
```
Lockyy, line 187: What do you think about including something related to funding in the name of this column? Something like `eyl_funding_eligible`. That'd make it very explicit that it's funding related rather than what courses you can apply for.

lib/tasks/eyl_eligible_schools.rake
```diff
@@ -0,0 +1,35 @@
+require "csv"
+
+namespace :eyl_eligible_schools do
+  desc "Make schools eligible for NPQEYL courses"
+  task :update, %i[file_name] => :environment do |_t, args|
+    updated_records = 0
+    update_errors = 0
+    file_name = args.file_name
+
+    raise "File not found: #{file_name}" unless File.exist?(file_name)
```
Lockyy, line 10: Does the file need to be included in the repo?

db/schema.rb
leoapost: Indeed that should provide better context. I've pushed the changes

lib/tasks/eyl_eligible_schools.rake
leoapost: The file is included

PR #589, opened 2022-11-28
expand on how we use your data

EmmaFrith:
- Add 2 paras to the 'How we use your data' section to better encompass what we do with peoples data 
- Be clearer about what we mean re. 'informing government policy'
- Include 'run our services' in the 'improve our services' section - i.e. we need your information to make our services work (e.g. sharing information across services etc)
- be more user centred in the opening para - e.g. this is what happens when **you** register for an NPQ

Review the content: https://npq-registration-review-app-589.london.cloudapps.digital/privacy-policy

### DIFF
--- a/app/views/pages/privacy_policy.html.erb
+++ b/app/views/pages/privacy_policy.html.erb
@@ -89,10 +89,10 @@
       We’ll also look at any feedback you share - for example, through website surveys.
     </p>
 
-    <h3 class="govuk-heading-m">Getting insight to make government policy better</h3>
+    <h3 class="govuk-heading-m">Informing government policy</h3>
 
     <p class="govuk-body">
-      We may use your data to help inform government policy around teacher recruitment and retention.
+      We may use your data to help inform government policy.
     </p>
     
     <h3 class="govuk-heading-m">Identifying you so that you can access DfE services</h3>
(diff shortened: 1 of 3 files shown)


app/views/pages/privacy_policy.html.erb
```diff
@@ -89,11 +93,22 @@
...
 
     <p class="govuk-body">
-      We may use your data to help inform government policy around teacher recruitment and retention.
+      Your data contributes to our understanding education, training and care in England, helping us to inform government policy.
     </p>
+    
+    <h3 class="govuk-heading-m">Identifying you so that you can access DfE services</h3>
+    
+    <p class="govuk-body">
+      If we do not already hold your data on our records, we’ll create a professional record for you.
```
peteryates, line 105: a _professional_ record?

EmmaFrith: e.g. the DQT

EmmaFrith: I was debating whether to drop that bit and go straight in with the 'identifying you' stuff. Thoughts?

EmmaFrith, line 105:
```suggestion
      If we do not already hold your data on our records, we’ll add you to our database of educational professionals.
```

peteryates (approved):
Looks good!

PR #600, opened 2022-12-01
[CPDNPQ-723] Only allow targeted delivery funding once

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CPDNPQ-723

`targeted_delivery_funding_eligibility` is now filtered through a check to the ECF DB for whether the user has already been received targeted delivery funding for the course they have selected on an accepted application. If they have, they won't be funded.

I've moved the check for this into Services::FundingEligibility to make sure we don't make multiple queries to the ECF API.

### DIFF
--- a/app/lib/services/funding_eligibility.rb
+++ b/app/lib/services/funding_eligibility.rb
@@ -63,11 +63,23 @@ module Services
       end
     end
 
+    def eligible_for_targeted_delivery_funding?
+      !previously_received_targeted_funding_support? &&
+        course_and_institution_eligible_for_targeted_delivery_funding?
+    end
+
     def ineligible_institution_type?
       [NO_INSTITUTION, INELIGIBLE_INSTITUTION_TYPE].include?(funding_eligiblity_status_code)
     end
 
   private
+
+    def course_and_institution_eligible_for_targeted_delivery_funding?
+      @course_and_institution_eligible_for_targeted_delivery_funding ||= Services::Eligibility::TargetedDeliveryFunding.new(
+        institution:,
+        course:,
+      ).call
+    end
 
     def inside_catchment?
       @inside_catchment
@@ -120,10 +132,16 @@ module Services
       ]
     end
 
-    def previously_funded?
-      results = EcfApi::NpqFunding.with_params(npq_course_identifier: course.identifier).find(@trn)
+    def ecf_api_funding_lookup
+      @ecf_api_funding_lookup = EcfApi::NpqFunding.with_params(npq_course_identifier: course.identifier).find(@trn)
+    end
 
-      results["previously_funded"] == true
+    def previously_funded?
+      ecf_api_funding_lookup["previously_funded"] == true
+    end
+
+    def previously_received_targeted_funding_support?
+      ecf_api_funding_lookup["previously_received_targeted_funding_support"] == true
     end
   end
 end
--- a/app/lib/services/funding_eligibility.rb
+++ b/app/lib/services/funding_eligibility.rb
@@ -64,8 +64,8 @@ module Services
     end
 
     def eligible_for_targeted_delivery_funding?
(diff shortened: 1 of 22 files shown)


app/lib/services/handle_submission_for_store.rb
```diff
@@ -188,17 +188,15 @@ def funding_eligibility
       eligible_for_funding?
     end
```
shjohnson, line 190: What happens if the transaction is rolled back and the job is still enqueued?

shjohnson (commented):
few questions but maybe not all blockers

app/lib/services/funding_eligibility.rb
```diff
@@ -63,12 +63,24 @@ def funding_eligiblity_status_code
...
+        !previously_received_targeted_funding_support? # Check last as it involves an API call
+    end
+
     def ineligible_institution_type?
       [NO_INSTITUTION, INELIGIBLE_INSTITUTION_TYPE].include?(funding_eligiblity_status_code)
     end
 
   private
 
+    def course_and_institution_eligible_for_targeted_delivery_funding?
```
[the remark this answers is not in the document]
Lockyy: Ah no it's a series of checks, there's the various safety checks but the core check is if a User's school is a valid type of school and has less than 600 pupils.

This is the step before the API call to see if the user is already unable to receive funding without the check for if they're previously funded applying.

app/lib/services/handle_submission_for_store.rb
Lockyy: In that situation then the Job wouldn't succeed, though with the Job being enqueued as the final step it's only a failure within the enqueuing that would prevent the transaction finalising.

app/lib/services/funding_eligibility.rb
```diff
@@ -63,12 +63,24 @@ def funding_eligiblity_status_code
       end
```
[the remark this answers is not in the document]
Lockyy: This may be a solution that satisfies the competing desires of only querying the API once during all calculations and maintaining the services as having a single responsibility.

```
class FundingCalculator
  def initialize(...)
    @funding_calculation_data = FundingCalculationData.new(...)
  end

  def funding_eligibility
    FundingEligibilityCalculator.new(@funding_calculation_data).call
  end

  def targeted_delivery_funding_eligibility
    TargetedDeliveryFundingEligibilityCalculator.new(@funding_calculation_data).call
  end
end

class FundingCalculationData
  ...
    
  def api_call
    @api_call ||= ...
  end
end
```

This would move all the logic related to each type of eligibility into a single class with each being given a data store containing everything they need which would allow us to memoize the api call response.

app/lib/services/handle_submission_for_store.rb
shjohnson: ok cool, was just checking 👍

Lockyy: To be safer and to protect from future changes, we should move the enqueue outside the transaction 🙂 That way we don't accidentally add things after that in the block and cause issues.

app/lib/services/funding_eligibility.rb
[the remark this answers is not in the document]
Lockyy: Overall however, whether anything needs changing or refactoring in this class is outside the scope of this PR and we'd need to have a large discussion about working in time to do this sort of refactor along with deciding what the correct approach is. 

Moving this into the model directory may be the simplest solution since this is clearly a model and not a service.

Lockyy:
@shjohnson Have adjusted the PR with a couple of your notes. Other changes are more related to refactoring existing structures which are out of scope so I won't be making any of those changes here. Unless there's any blockers I think we're good to proceed.

shjohnson (approved):
🟢

PR #611, opened 2022-12-08
Downcase all emails from the TraOpenidConnect provider

Lockyy:
### Context

Emails are stored in the database as downcased strings, this downcasing occurs after validation. Emails coming back from the TRA are not downcased meaning that a find_or_initialize_by using this data without downcasing would lead to a new record with the non-downcased string being created. When this goes to be persisted it would trigger a validation error.

- Downcase all emails coming from the TRA integration.


shjohnson (approved):
LGTM Nice Catch 🍰

PR #614, opened 2022-12-12
CPDNPQ480 ITT lead mentor registration

shjohnson:
### Context

This ticket is to deliver a new registration flow for ITT Lead Mentors to tell us which accredited ITT provider they work for so that we can determine eligibility for funding for the NPQLTD course. 

Jira: https://dfedigital.atlassian.net/browse/CPDNPQ-480


spec/lib/services/approved_itt_providers/update_spec.rb
```diff
@@ -0,0 +1,89 @@
+require "rails_helper"
+
+RSpec.describe Services::ApprovedIttProviders::Update do
+  let(:now) { "2022-12-12 18:00:00" }
+
+  around do |example|
+    Timecop.freeze(now) do
```
Lockyy, line 7: We can do this with built in functionality now, we don't need Timecop fortunately. https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html

This can be written as:
```
travel_to(Time.parse(now)) do
  example.run
end
```

app/lib/forms/itt_provider.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def question
+      Forms::QuestionTypes::AutoCompleteIttProvider.new(name: QUESTION_NAME)
+    end
+
+    def next_step
+      if approved_itt_provider?
+        :choose_your_npq
+      end
```
Lockyy, line 20: Where would it go otherwise?

app/models/registration_wizard.rb
```diff
@@ -213,6 +219,10 @@ def institution_from_store
     institution(source: store["institution_identifier"])
   end
 
+  def lead_mentor_for_accredited_itt_provider?
```
Lockyy, line 222: Can this go on Services::QueryStore?

app/lib/forms/choose_your_npq.rb
```diff
@@ -2,15 +2,24 @@ module Forms
   class ChooseYourNpq < Base
     include Helpers::Institution
 
-    attr_accessor :course_id
+    QUESTION_NAME = :course_id
```
Lockyy, line 5: Will this be used for setting up a question and moving to the shared template for the question page? Like in Forms::IttProvider?

app/views/registration_wizard/work_setting.html.erb
```diff
@@ -1,34 +1,7 @@
...
-    <% end %>
-  </div>
-</div>
+<%=
+  render(
+    'registration_wizard/shared/question_page',
+    form: @form,
+    wizard: @wizard,
+  )
+%>
```
Lockyy, line 7: Love it, this is so nice to see spreading around the other questions. 😁

db/migrate/20221212192830_create_itt_providers.rb
```diff
@@ -0,0 +1,15 @@
+class CreateIttProviders < ActiveRecord::Migration[6.1]
+  def change
+    create_table :itt_providers do |t|
+      t.text :legal_name, unique: true
+      t.text :operating_name
+      t.date :removed
+      t.date :added
```
Lockyy, line 7: Do we need added when we already have created_at?

db/migrate/20221212192830_create_itt_providers.rb
```diff
@@ -0,0 +1,15 @@
+class CreateIttProviders < ActiveRecord::Migration[6.1]
+  def change
+    create_table :itt_providers do |t|
+      t.text :legal_name, unique: true
+      t.text :operating_name
+      t.date :removed
```
Lockyy, line 6: It would fit more in with the rest of the schema for this to be removed_at

lib/tasks/approved_itt_providers.rake
```diff
@@ -0,0 +1,23 @@
...
+    Services::ApprovedIttProviders::Update.call(file_name:)
+
+    Rails.logger.info("Import finished")
+  end
+end
+
+# add test file
+
+# how to add the dropdown
+# run tests and see flow break
```
Lockyy, line 23: Leftover notes

app/lib/forms/choose_your_npq.rb
```diff
@@ -41,6 +50,8 @@ def next_step
         end
       elsif course.ehco?
         :about_ehco
+      elsif approved_itt_provider? && lead_mentor_course?
+        :possible_funding
```
Lockyy, line 54: Does this mean that this always leads to be a user being funded?

Lockyy (commented):
With a new path through the questionnaire we'll need to add in a new feature spec file that follows that path.

app/lib/forms/itt_provider.rb
shjohnson: ahh yes i need to confirm this but the flow diagram didnt have any other option?

app/lib/forms/choose_your_npq.rb
Lockyy: If so, we need to adjust the Services::FundingEligibility to update the user's funding status based on this information. Once that's done the next if statement, ` if eligible_for_funding?` would navigate the user accordingly.

app/models/registration_wizard.rb
shjohnson: il have a look

app/lib/forms/choose_your_npq.rb
shjohnson: Yes, it was left over from another branch i have which i removed for this bit of work

app/views/registration_wizard/work_setting.html.erb
shjohnson: Yeah will try to do as many as possible although still few bits that im confused with how the magic sets up everything.

app/lib/forms/choose_your_npq.rb
Lockyy: If we don't update Services::FundingEligibility then we'll be directing the user down the "you are expected to be funded" path without adjusting how the eligible_for_funding or funding_eligiblity_status_code data stored against their application is calculated.

db/migrate/20221212192830_create_itt_providers.rb
shjohnson: possibly... i did think this then thought of reason why not i can have another look

db/migrate/20221212192830_create_itt_providers.rb
shjohnson: ok i can change that 👍

shjohnson: think i had it this way at first to be honest

app/lib/forms/choose_your_npq.rb
shjohnson: As i understand yes? although also the symbol :possible_funding seems a bit misleading! I guess the provider still has to accept the application

Lockyy: Provider acceptance is unrelated and happens after funding eligibility is calculated. 

There's an element of backwards logic going on here. You go to possible_funding because your eligible_for_funding value was true. That's what the next line is doing. Rather what we have here is doing that data check then sending the user on without adjusting how funding eligibility is actually calculated.

If a user selecting an approved ITTProvider and having a lead mentor course should change the result of eligible_for_funding when the application is saved and sent to the ECF DB then you need to update Services::FundingEligibility to correctly calculate their funding, as it stands the above change only updates what information the user is shown and doesn't change the result of their application.

app/lib/forms/itt_provider.rb
Lockyy: Will definitely need to figure that out, I think if you don't provide a fallback it'll trigger a 500. Unless it's meant to just raise an error in which case `approved_itt_provider?` should go in the validations on the form and then :choose_your_npq is always the next step if validations pass.

spec/features/happy_journeys/js/when_choose_lead_mentor_course.rb
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    allow(ApplicationSubmissionJob).to receive(:perform_later).with(anything)
+
+    expect_page_to_have(path: "/registration/check-answers", submit_button_text: "Submit", submit_form: true) do
+      expect_check_answers_page_to_have_answers(
+        {
+
+          "Course" => "NPQ for Leading Teacher Development (NPQLTD)",
+          "Employer" => "",
```
Lockyy, line 88: Do we want empty answers here? We should be omitting these answers if the user didn't get asked the question.

spec/features/happy_journeys/js/when_choose_lead_mentor_course.rb
```diff
@@ -0,0 +1,193 @@
...
+      "trn_auto_verified" => false,
+      "trn_verified" => true,
+      "uid" => user_uid,
+    )
+
+    expect(retrieve_latest_application_data).to eq(
+      "cohort" => 2022,
+      "course_id" => Course.find_by_code(code: :NPQLTD).id,
+      "ecf_id" => nil,
+      "eligible_for_funding" => false,
```
Lockyy, line 156: Is this correct? The user picked NPQLTD and selected an approved ITT Provider. They should be funded shouldn't they? It seems based on the funding_eligiblity_status_code below that they were marked ineligible because the first line of the calculator kicked in. The ITT check should happen before the institution check.

spec/lib/services/funding_eligibility_spec.rb
```diff
@@ -164,6 +164,8 @@
       end
     end
 
+    context "when the "
```
Lockyy, line 167: ?

peteryates: ...moon hits your eye like a big pizza pie? 😂

app/views/registration_wizard/choose_your_npq.html.erb
```diff
@@ -35,13 +33,11 @@
...
         <% @form.options.each do |struct| %>
           <%= f.govuk_radio_button :course_id, struct.value, label: { text: struct.text }, hint: { text: struct.hint }, link_errors: struct.link_errors %>
         <% end %>
       <% end %>
-
       <%= f.govuk_submit %>
     <% end %>
   </div>
-</div>
+</div>
```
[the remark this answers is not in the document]
peteryates: Isn't this governed by the `.editorconfig`?

spec/features/happy_journeys/js/when_choose_lead_mentor_course.rb
shjohnson: Same now changed from testing

spec/lib/services/funding_eligibility_spec.rb
shjohnson: I have no words

spec/factories/itt_provider.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :itt_provider do
+    sequence(:legal_name) { |n| "private childcare provider #{n}" }
```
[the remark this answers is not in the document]
shjohnson: good point and already changed this

app/views/registration_wizard/choose_your_npq.html.erb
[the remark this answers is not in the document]
shjohnson: This was created through the removal of the new question format into new file but will check out editor as something changed recently

app/lib/services/eligibility/targeted_delivery_funding.rb
```diff
@@ -18,6 +19,8 @@ def call
         return false if institution.is_a?(PrivateChildcareProvider)
         return false if institution.number_of_pupils.nil?
         return false if institution.number_of_pupils.zero?
+
+        return false if employment_role == "lead_mentor_for_accredited_itt_provider"
```
Lockyy, line 23: Not a blocker on merging.

This string is being repeated in various places around the app and for checking. It'd be nice to have this moved into a constant somewhere, perhaps on Forms::YourEmployment and then referred to elsewhere via that constant. To guard against typos and changes in the future.

app/lib/services/funding_eligibility.rb
```diff
@@ -35,9 +48,11 @@ def funded?
 
     def funding_eligiblity_status_code
       @funding_eligiblity_status_code ||= begin
-        return NO_INSTITUTION if institution.nil?
+        return NO_INSTITUTION if institution.nil? && !lead_mentor_course? && !approved_itt_provider
         return PREVIOUSLY_FUNDED if previously_funded?
-        return FUNDED_ELIGIBILITY_RESULT if eligible_urns.include?(institution.urn)
+        return FUNDED_ELIGIBILITY_RESULT if eligible_urns.include?(institution.try(:urn))
+        return FUNDED_ELIGIBILITY_RESULT if lead_mentor_course? && approved_itt_provider && lead_mentor
+        return NOT_LEAD_MENTOR_COURSE if approved_itt_provider && lead_mentor
```
Lockyy, line 55: This section has become confusing to me as the guard clauses are relying on each others results and the lead mentor checks are weaving into non-lead mentor results. It works, but it's harder to parse than it has to be.

It might be clearer to rework all the above into something like this so that if you have chosen a lead mentor, you get diverted to a whole set of separate status code calculations that can be reasoned about. It also means that the rest of the guard clauses remain untouched.

```
def funding_eligiblity_status_code
  @funding_eligiblity_status_code ||= begin
    return lead_mentor_eligibility_status_code if lead_provider_status_code_calculations_active?

    return NO_INSTITUTION if institution.nil?
    return PREVIOUSLY_FUNDED if previously_funded?
    return FUNDED_ELIGIBILITY_RESULT if eligible_urns.include?(institution.try(:urn))

    ...
  end
end

def lead_provider_status_code_calculations_active?
  # I'm not sure the exact trigger for this, I think it would be:
  lead_mentor
end

def lead_mentor_eligibility_status_code
  return NOT_LEAD_MENTOR_COURSE unless lead_mentor_course? # Don't do a check to the API unless necessary
  return PREVIOUSLY_FUNDED if previously_funded?
  return FUNDED_ELIGIBILITY_RESULT if approved_itt_provider

  # This is a new status code, currently the code reverts to INELIGIBLE_INSTITUTION_TYPE which is less specific and clear what happened
  return NOT_APPROVED_IIT_PROVIDER_RESULT
end
```

The above rules and checks may not be exactly what they need to be but I think it's important to not weave the two sets of checks together as it feels like the beginning of the road of making this section hard to keep straight in your head.

spec/lib/services/funding_eligibility_spec.rb
```diff
@@ -164,6 +173,32 @@
...
+      context "and the course is not NPQLTD" do
+        Course.all.reject(&:npqltd?).each do |course|
+          let(:course) { course }
+
+          it "is not eligible for #{course.name}" do
+            expect(subject.funded?).to be_falsey
+            expect(subject.funding_eligiblity_status_code).to eq :not_lead_mentor_course
+          end
+        end
+      end
```
Lockyy, line 199: I think we should have a spec here for if approved_itt_provider is false.

app/lib/services/eligibility/targeted_delivery_funding.rb
shjohnson: agreed 👍

spec/lib/services/funding_eligibility_spec.rb
shjohnson: Thats fair although in the feature flow it cant really get here because of the validations but from a spec level test it could happen.

app/lib/services/funding_eligibility.rb
shjohnson: I agree and wanted to suggest this code needs more attention, do you feel this is a blocker for this release?

app/lib/services/funding_eligibility.rb
```diff
@@ -35,9 +48,13 @@ def funded?
...
     def funding_eligiblity_status_code
       @funding_eligiblity_status_code ||= begin
+        if approved_itt_provider && lead_mentor
+          return lead_mentor_eligibility_status
+        end
+
         return NO_INSTITUTION if institution.nil?
         return PREVIOUSLY_FUNDED if previously_funded?
-        return FUNDED_ELIGIBILITY_RESULT if eligible_urns.include?(institution.urn)
+        return FUNDED_ELIGIBILITY_RESULT if eligible_urns.include?(institution.try(:urn))
```
Lockyy, line 57: I don't think we need this check anymore 🙂

app/lib/services/funding_eligibility.rb
```diff
@@ -35,9 +48,13 @@ def funded?
 
     def funding_eligiblity_status_code
       @funding_eligiblity_status_code ||= begin
+        if approved_itt_provider && lead_mentor
+          return lead_mentor_eligibility_status
+        end
```
Lockyy, line 53: Have spoken to Sam and understand that the approved_itt_provider being false if you're a lead mentor is an edge-case that shouldn't occur since proceeding past that point in the journey would require you to force your way past. These users receiving a NO_INSTITUTION error is correct since they *should* have an institution (an approved_itt_provider) but don't.

app/lib/services/funding_eligibility.rb
```diff
@@ -74,10 +91,19 @@ def ineligible_institution_type?
 
   private
 
+    def lead_mentor_eligibility_status
+      if lead_mentor_course?
+        FUNDED_ELIGIBILITY_RESULT
```
Lockyy, line 96: Oh I was just about to approve but noticed an issue. If you have an approved ITT Provider and are a lead mentor, and pick the right course, you can be funded multiple times. We just need to add the previously funded check in here to make sure we still check for previous funding:

```suggestion
        return PREVIOUSLY_FUNDED if previously_funded? 
        
        FUNDED_ELIGIBILITY_RESULT
```

With that in place we'll be good to go

Lockyy (approved):

Lockyy:
Ship it

PR #619, opened 2022-12-14
[CPDNPQ-655] Record logs to the DB of sync requests to the ECF DB

Lockyy:
### Context

We are building out a set of tools in the admin interface for making sure users are properly synced to the ECF DB. Right now there isn't explicit logging of the results of syncing, making it harder to see when and why a sync attempt failed.

- Add in DB logging for sync requests


peteryates (approved):
Nice!

PR #624, opened 2022-12-22
Remove mention of 'cost' from post submission content

EmmaFrith:
- Remove mention of 'costs' from post submission copy - most people do not need to pay
- Make it more obvious you need to apply with the provider

| Before  | After |
| Screenshot 2022-12-28 at 10 18 42      | Screenshot 2022-12-28 at 10 13 07     |

### DIFF
--- a/app/views/registration_wizard/confirmation.html.erb
+++ b/app/views/registration_wizard/confirmation.html.erb
@@ -14,11 +14,11 @@
       }
     ) %>
 
-    <h2 class="govuk-heading-m">What happens next</h2>
+    <h2 class="govuk-heading-m">Apply through your provider</h2>
 
-    <p class="govuk-body"><%= @form.lead_provider.name %> will contact you to complete your registration.</p>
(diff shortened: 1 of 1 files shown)


FMooi:
The text to tell users they still have a next step in the application process is really clear. 
I just had a comment about whether the name of the NPQ should be: Leading Teaching NPQ (NPQLT). Not a deal breaker but to be consistant with the PDfT&L pages and earlier in the flow?

FMooi (commented):
Sorry think I left my comment in the wrong place. All the next steps to apply text looks good. Just querying the course naming, but I'd be open to it staying as is as it's the only time in the flow the acronym is used so might be helpful to see as providers also use the acronym.

EmmaFrith:
> The text to tell users they still have a next step in the application process is really clear. I just had a comment about whether the name of the NPQ should be: Leading Teaching NPQ (NPQLT). Not a deal breaker but to be consistant with the PDfT&L pages and earlier in the flow?

Good point about the inconsistency with our naming conventions. I'll look at this in a separate PR!

FMooi (approved):

PR #628, opened 2022-12-28
link to PDTL site from home page

EmmaFrith:
## What

Update this link destination. 

Screenshot 2022-12-28 at 12 57 16

| Current destination | Proposed destination |
| Screenshot 2022-12-28 at 12 57 42      | Screenshot 2022-12-28 at 13 01 56     |

## Why?

The 'Choose an NPQ and provider' page duplicates content on the PDTL site. The PDTL now lists providers. 

## Other issues 

The 'Choose and NPQ and provider' page also shows if you answer 'No' to the question 'Have you already chosen an NPQ and provider?' 

Screenshot 2022-12-28 at 14 39 11

I've proposed removing the provider list on this page for now, to avoid duplication with the PDTL site.

| Current content | Proposed content |
| Screenshot 2022-12-28 at 14 40 20      | Screenshot 2022-12-28 at 14 41 37     |

## Next steps 

We should review whether the 'Choose an NPQ and provider' page and the question 'Have you already chosen an NPQ and provider?' is needed at all. 

We do already inform people that you need to choose an NPQ and provider on the start page. 

Is there evidence to suggest that people come to the NPQ service without already knowing their NPQ and provider?

### DIFF
--- a/app/views/registration_wizard/start.html.erb
+++ b/app/views/registration_wizard/start.html.erb
@@ -8,7 +8,7 @@
 
     <h2 class="govuk-heading-m">Before you start</h2>
 
-    <p class="govuk-body"><%= govuk_link_to("Choose an NPQ and provider", root_path) %>.</p>
+    <p class="govuk-body"><a href="https://professional-development-for-teachers-leaders.education.gov.uk/" class="govuk-link">Choose an NPQ and provider</a>.</p>
 
     <p class="govuk-body">If you work in early years or childcare, <%= govuk_link_to("check if your workplace is registered with Ofsted", "https://reports.ofsted.gov.uk/childcare") %> and get their Ofsted unique reference number (URN) if they have one.</p>
 
--- a/app/views/registration_wizard/start.html.erb
+++ b/app/views/registration_wizard/start.html.erb
@@ -8,7 +8,7 @@
 
     <h2 class="govuk-heading-m">Before you start</h2>
 
-    <p class="govuk-body"><%= govuk_link_to("Choose an NPQ and provider", choose_an_npq_and_provider_path) %>.</p>
+    <p class="govuk-body"><a href="https://professional-development-for-teachers-leaders.education.gov.uk/" class="govuk-link">Choose an NPQ and provider</a>.</p>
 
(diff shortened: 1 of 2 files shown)


app/views/registration_wizard/start.html.erb
```diff
@@ -8,7 +8,7 @@
 
     <h2 class="govuk-heading-m">Before you start</h2>
 
-    <p class="govuk-body"><%= govuk_link_to("Choose an NPQ and provider", choose_an_npq_and_provider_path) %>.</p>
+    <p class="govuk-body"><a href="https://professional-development-for-teachers-leaders.education.gov.uk/" class="govuk-link">Choose an NPQ and provider</a>.</p>
```
peteryates, line 11: I believe this will give you the same result but lets us keep using the link helper:

```suggestion
    <p class="govuk-body"><%= govuk_link_to("Choose an NPQ and provider", root_path) %>.</p>
```

peteryates (approved):
Looks good to me, left one suggestion on the link.

EmmaFrith:
> Looks good to me, left one suggestion on the link.

Thanks Pete! I tried the link change u suggested, but then realised that's the link to the reg service start page. I want it to go here: https://professional-development-for-teachers-leaders.education.gov.uk/

Diana-Mundo (commented):
We had some issues during research where users were confused about the wording on this link "Choose an NPQ and provider". They thought this was a step they had to take in order to register. Is it worth reviewing this copy as well?

Also, is it possible to just re-direct users to the PDfTL site without that intermediate page?

EmmaFrith:
> We had some issues during research where users were confused about the wording on this link "Choose an NPQ and provider". They thought this was a step they had to take in order to register. Is it worth reviewing this copy as well?
> 
> Also, is it possible to just re-direct users to the PDfTL site without that intermediate page?

Definitely worth reviewing the copy and the question in general. Shall we treat that as a separate issue and look at it in a separate PR?

FMooi (approved):
Moving to the PDfT&L site makes sense. Should the PDfT&L site link in the second instance open in a new tab so you can easily return to the reg flow?
Agreed to raise a separate PR on the copy.

EmmaFrith:
thanks @FMooi - happy to look at the link issue in a separate PR, along with the copy

PR #636, opened 2023-01-05
[CPDNPQ-656] [CPDNPQ-657] Improve user and application management in admin interfaces

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CPDNPQ-656
https://dfedigital.atlassian.net/browse/CPDNPQ-657

#### Changes
 
- Add Users index and show to admin interface. 
  This allows us to view users and see a list of their applications and ECF sync attempts.
- Add ECF sync log table to admin applications interface
- Add list of unsynced users to admin interface 
- Add buttons to user and applications admin show pages to allow rescheduling ECF sync if records are unsynced and next sync is over 5 minutes in the future.
- Add functionality specs for admin interfaces

#### Minor changes

- Merge unsynced_applications#show into applications#show.
- Update admin/schools#show and admin/schools#index to display `-` instead of whitespace for missing data.
- Show newest applications first on admin/unsynced_applications#index

#### Bugfixes

- Fix issue where pagination broke tab highlighting in admin interface.
- Fix classname capitalisation issues on admin/school#show
- ApplicationSubmissionJob now fails when sync to ECF of User or Application records fails, allowing job to retry at a later time


app/components/subnav_component/nav_item_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="app-subnav__list-item<%= current_page?(path, check_parameters: true) ? " app-subnav__list-item--selected" : "" %>">
```
Lockyy, line 1: This was preventing the tab from highlighting if there were any query params, for example /admin/users?page=2 would stop highlighting the users tab because the params didn't match the URL on the button.

app/controllers/admin/application_submissions_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def create
+    user = User.find(params[:user_id])
+
+    if user.synced_to_ecf? && user.applications_synced_to_ecf?
+      flash[:error] = "User has already been synced to ECF"
+    else
+      existing_ecf_sync_jobs = user.ecf_sync_jobs.to_a
+
+      if ApplicationSubmissionJob.perform_later(user:)
+        existing_ecf_sync_jobs.each(&:delete)
```
Lockyy, line 11: We only want to have one job in the queue

app/controllers/admin/unsynced_applications_controller.rb
```diff
@@ -5,13 +5,9 @@ def index
     @pagy, @applications = pagy(scope)
   end
 
-  def show
-    @application = Application.unsynced.includes(:user).find(params[:id])
-  end
```
Lockyy, line 10: This is merged into admin/applications#show.

The pages were identical and merging them together simplifies the code.

app/lib/services/ecf/ecf_user_creator.rb
```diff
@@ -32,6 +32,9 @@ def call
         Sentry.with_scope do |scope|
           scope.set_context("User", { id: user.id })
           Sentry.capture_exception(e)
+
+          # Re-raise to fail the sync, we'll want to retry again later
+          raise e
```
Lockyy, line 37: This is how the class used to function before I added in the exception capturing.
 
Previously if a sync failed in a way that raised an exception (which includes failed saves since JsonApiClient::Resource uses errors for flow control) then the ApplicationSubmissionJob would fail and be retried later, which may allow the error to resolve itself. Without it the job would succeed and no retry would occur.

app/lib/services/ecf/npq_profile_creator.rb
```diff
@@ -64,6 +64,9 @@ def call
         Sentry.with_scope do |scope|
           scope.set_context("Application", { id: application.id })
           Sentry.capture_exception(e)
+
+          # Re-raise to fail the sync, we'll want to retry again later
+          raise e
```
Lockyy, line 69: See comment on app/lib/services/ecf/ecf_user_creator.rb.

app/views/admin/applications/show.html.erb
```diff
@@ -1,14 +1,36 @@
 <% content_for :before_content do %>
   <%= render GovukComponent::BackLinkComponent.new(
     text: "Back",
-    href: admin_applications_path
+    href: url_for(:back)
```
Lockyy, line 4: This page can be reached from one of two indexes, this will take the user back to the correct one.

app/components/subnav_component/nav_item_component.html.erb
```diff
@@ -1,4 +1,4 @@
...
+<li class="app-subnav__list-item<%= current_page?(path) ? " app-subnav__list-item--selected" : "" %>">
```
peteryates, line 1: This might be easier to read as a method in the `NavItemComponent`.

```def
def subnav_list_item_classes
  class_names(
    "app-subnav__list-item"
    "app-subnav__list-item--selected" => current_page?(path)
  )
end
```

```
tag.li(class: subnav_list_item_classes) do
  ...
end
```

peteryates (approved):
Looks great.

PR #641, opened 2023-01-11
Remove default_cohort override for course seeds

Lockyy:
This was matching the old data, removing the override makes the seeding fallback to the DB default, which is 2022.


shjohnson (commented):
Wunderbar 🙌 

Approving now but realise few tests need fixing just as referencing the old date.

Lockyy:
@shjohnson I've fixed those tests 🙂

PR #643, opened 2023-01-16
Copy feature flags to all users on login

Lockyy:
## Background 

The spec in this PR was created to replicate a bug that’s triggered 80 times over about 3 months ([viewable on Sentry here](https://sentry.io/organizations/dfe-teacher-services/issues/3792837223/?)) and leads to a full error page for users who encounter it that blocks progress. This flow seems like such an extreme edge-case that I'm not convinced that this is the trigger that lead to all 80 of those errors, but it is definitely *a* path to it triggering, so closing it is a positive change.

## Recreation steps

1. Go through the GAI flow without a verified TRN, return and be removed from the pilot as no TRN is returned.
2. Go through the non-GAI flow until you reach the email input, enter a different email to the one you used on GAI for a user that already exists that doesn’t have a TRN on the record. You will be logged in as this user.
3. Enter your TRN on the TRN step, it is required and you cannot proceed without one.
4. Proceed and complete the Choose your NPQ page.
5. Encounter a 500 error.

You encounter a 500 error here because after the Choose your NPQ page a funding eligibility check is performed. This check requires your TRN.

Since your email was already on a user record when you entered it on the email input page you were logged in as that user but importantly that user did not have your feature flag ID copied across (since it’s only copied for new users) and so your feature flags reset to the flags for that user. This user’s flags more than likely would leave you as a part of the pilot. 

Because you were still considered a part of the pilot during the eligibility check the TRN for the user would be looked up on the user record in the wrong place looking where it would be for GAI users instead of for non-GAI users. Then it tries to use nil when constructing the URL and fails. 

## Fix 

The solution I found was to just copy feature flags during log in for new and existing uses. This overall is a better approach IMO because it’s not a great idea to have any user suddenly jump from one set of flags to another. The solution was to copy the feature flags from the session ID to the user even if it isn't a new user being created.

Alongside changing the feature flag ID copying for the non-pilot journey I have updated it during the Get an Identity journey too. This way no matter what you'll always end up maintaining the feature flag state that you began your journey in, even if you have previously gone through the journey differently. The possible combinations of going from one set of flags to another is essentially untestable and will lead to a large amount of edge cases.

### DIFF
--- a/app/lib/forms/confirm_email.rb
+++ b/app/lib/forms/confirm_email.rb
@@ -24,11 +24,13 @@ module Forms
     end
 
     def after_save
-      user = User.find_or_create_by!(email: wizard.store["email"]) do |new_user|
-        # Transfer feature flag from NullUser to User when created, if the user
-        # already exists then the user will revert to their original set of flags.
-        new_user.feature_flag_id = wizard.session["feature_flag_id"]
-      end
+      user = User.find_or_create_by!(email: wizard.store["email"])
+
+      # Set the feature flag ID on the logged in user to match the session's existing ID.
+      # This way the user has a smooth journey through the app and isn't suddenly swapped
+      # from one A/B tested feature to another.
+      user.update!(feature_flag_id: wizard.session["feature_flag_id"])
+
       # TODO: protect against session fixation
       wizard.session["user_id"] = user.id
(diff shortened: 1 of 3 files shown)


shjohnson (approved):
LGTM

PR #648, opened 2023-01-17
Use safe navigation when logging errors

Lockyy:
Unsafe navigation here was triggering data being logged that shouldn't have been.


peteryates (approved):
Looks good. I suspect a rebase will exlcude the `deploy_to_review_app.yml` change.

PR #656, opened 2023-01-20
Automatically update github-actions with dependabot

tvararu:
Dependabot can manage bumping e.g. `actions/checkout@v2` to `actions/checkout@v3` on our behalf, saving us time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,8 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -15,3 +15,8 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: github-actions
```
shjohnson, line 18: Does this automatically add us as reviewers and do we need to ad any labels for the PR? 

Also wonder if there is a set time for dependabot checking for updates? i know you can set per package but wondered if we can set for everything? something like 8am?

shjohnson: otherwise this looks great thanks 👍

shjohnson (approved):

.github/dependabot.yml
tvararu: Settings are the same as the other package ecosystems in the file, so based on a recent PR like https://github.com/DFE-Digital/npq-registration/pull/645, it does look like reviewers are automatically added :+1: 

Set time would be probably same as the other ones. I'm not sure about further settings.

PR #662, opened 2023-01-23
Admin panel style improvement

Lockyy:
Bools in the admin panel should be styled with red/green indicators


peteryates (approved):
Looks good.

There's the [tag component](https://govuk-components.netlify.app/components/tag/#coloured-tags) too which might be useful for tag generation.

Lockyy:
That'd be nice to integrate with to replace this implementation.

PR #663, opened 2023-01-23
Edit start page in line with prototype

EmmaFrith:
In user research, a user clicked on the 'choose an NPQ and provider' link when she had already chosen an NPQ and provider. 

Hypothesis:

Editing the link text and placement may help prevent people clicking the link when they don't need to. 

| Before  | After |
| Screenshot 2023-01-23 at 16 21 18     | Screenshot 2023-01-23 at 16 21 45     |

### DIFF
--- a/app/views/registration_wizard/start.html.erb
+++ b/app/views/registration_wizard/start.html.erb
@@ -8,11 +8,18 @@
 
     <h2 class="govuk-heading-m">Before you start</h2>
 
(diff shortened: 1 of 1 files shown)


Lockyy (approved):
Much nicer, I like the change from saying “choose your npq and provider”. Once that was pointed out as looking like a start button it really stood out as a potential confusion point.

PR #666, opened 2023-01-24
Use new NPQ API for syncing users

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-713

We need to sync GAI IDs across to the ECF DB during user creation. https://github.com/DFE-Digital/early-careers-framework/pull/2919 has added in a new API that replaces the old one, accepting the same params as the previous API along with get_an_identity_id.

Use the above mentioned new API when syncing users, sending a get_an_identity_id when available.

### Guidance to review

Flows we need to test: 
- Regular flow where you return from the Get an Identity interface with a TRN. This will leave your NPQ user account with a get an identity ID which should then get synced when you complete your application.
- Alterntive flow where you return from the Get an Identity interface without a TRN. This leaves your NPQ user account without a GAI ID, which should *still* have your account synced when you complete your application. Only this time without an ID.

### DIFF
--- a/app/lib/ecf_api/npq/base.rb
+++ b/app/lib/ecf_api/npq/base.rb
@@ -1,0 +1,7 @@
+module EcfApi
+  module Npq
+    class Base < ::EcfApi::Base
+      self.site = "#{ENV['ECF_APP_BASE_URL']}/api/v1/npq"
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


shjohnson (approved):
Had to get my head round a couple of things but otherwise all good 💚

PR #667, opened 2023-01-24
Remove School-Led Network as a provider option

Lockyy:
This will leave the provider in the DB but not assigned to any courses or visible to users.

### Context

- https://dfedigital.atlassian.net/browse/CPDNPQ-743

Remove School-Led Network Lead Provider from list of providers for all courses. This will leave the provider in the DB but not assigned to any courses or visible to users.


shjohnson (approved):
Code wise seems fine although surprised its not removed from any tests? Good to go once the review app has had a once over.

Lockyy:
Removed from the specs that directly test which are displayed but it wasn't used in any of the end to end specs.

EmmaFrith (approved):

PR #676, opened 2023-01-30
Use new question format for the choose npq page

shjohnson:
Reformat the 'What are you applying for?' page

Only change the UI and not the value sent during submission

### DIFF
--- a/app/lib/forms/aso_new_headteacher.rb
+++ b/app/lib/forms/aso_new_headteacher.rb
@@ -43,7 +43,7 @@ module Forms
   private
 
     def course
-      Course.find_by(id: wizard.store["course_id"])
+      Course.find_by(name: ::Course::LEGACY_NAME_MAPPING[store["choose_your_npq"]])
     end
 
     def funding_eligiblity_status_code
--- a/app/lib/forms/aso_new_headteacher.rb
+++ b/app/lib/forms/aso_new_headteacher.rb
@@ -43,7 +43,7 @@ module Forms
   private
 
     def course
-      Course.find(wizard.store["course_id"])
+      Course.find_by(id: wizard.store["course_id"])
     end
 
     def funding_eligiblity_status_code
--- a/app/lib/forms/aso_new_headteacher.rb
+++ b/app/lib/forms/aso_new_headteacher.rb
@@ -43,7 +43,7 @@ module Forms
   private
 
     def course
-      Course.find_by(name: ::Course::LEGACY_NAME_MAPPING[store["choose_your_npq"]])
+      Course.find(wizard.store["course_id"])
     end
 
     def funding_eligiblity_status_code
--- a/app/lib/forms/choose_school.rb
+++ b/app/lib/forms/choose_school.rb
@@ -54,7 +54,7 @@ module Forms
   private
 
     def course
-      @course ||= Course.find_by(name: ::Course::LEGACY_NAME_MAPPING[wizard.store["choose_your_npq"]])
+      @course ||= Course.find(wizard.store["course_id"])
     end
 
     def institution_location
--- a/app/lib/forms/choose_school.rb
+++ b/app/lib/forms/choose_school.rb
@@ -54,7 +54,7 @@ module Forms
   private
 
     def course
-      @course ||= Course.find(wizard.store["course_id"])
+      @course ||= Course.find_by(name: ::Course::LEGACY_NAME_MAPPING[wizard.store["choose_your_npq"]])
     end
 
     def institution_location
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -17,6 +17,7 @@ module Forms
       Forms::QuestionTypes::RadioButtonGroup.new(
         name: :choose_your_npq,
         options:,
+        opts: { fieldset: { legend: { size: "m", tag: "h1" } } },
       )
     end
 
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -10,7 +10,7 @@ module Forms
     validate :validate_course_exists
 
     def self.permitted_params
-      [QUESTION_NAME.to_sym]
+      [QUESTION_NAME]
     end
 
     def question
@@ -110,7 +110,7 @@ module Forms
     end
 
     def previous_course
-      Course.find_by(id: wizard.store["choose_your_npq"])
+      Course.find_by(id: wizard.store["course_id"])
     end
 
     def previously_eligible_for_funding?
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -10,7 +10,7 @@ module Forms
     validate :validate_course_exists
 
     def self.permitted_params
-      [QUESTION_NAME]
+      [QUESTION_NAME.to_sym]
     end
 
     def question
@@ -35,6 +35,8 @@ module Forms
     end
 
     def after_save
+      wizard.store["course_id"] = course.id.to_s
+
       return if lead_provider_valid?
 
       wizard.store["lead_provider_id"] = nil
@@ -90,7 +92,7 @@ module Forms
     end
 
     def course
-      courses.find_by(id: course_id)
+      courses.find_by(name: ::Course::LEGACY_NAME_MAPPING[choose_your_npq])
     end
 
   private
@@ -108,7 +110,7 @@ module Forms
     end
 
     def previous_course
-      Course.find_by(id: wizard.store["course_id"])
+      Course.find_by(id: wizard.store["choose_your_npq"])
     end
 
     def previously_eligible_for_funding?
@@ -140,7 +142,7 @@ module Forms
 
     def validate_course_exists
       if course.blank?
-        errors.add(:course_id, :invalid)
+        errors.add(:choose_your_npq, :invalid)
       end
     end
   end
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -92,7 +92,7 @@ module Forms
     end
 
     def course
-      courses.find_by(name: ::Course::LEGACY_NAME_MAPPING[choose_your_npq])
+      courses.find_by(name: ::Course::DISPLAY_NAME_MAPPING[choose_your_npq])
     end
 
   private
--- a/app/lib/forms/choose_your_npq.rb
+++ b/app/lib/forms/choose_your_npq.rb
@@ -2,7 +2,7 @@ module Forms
   class ChooseYourNpq < Base
     include Helpers::Institution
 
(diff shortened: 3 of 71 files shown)


app/lib/forms/choose_school.rb
```diff
@@ -54,7 +54,7 @@ def possible_institutions
   private
 
     def course
-      @course ||= Course.find(wizard.store["course_id"])
+      @course ||= Course.find_by(name: ::Course::LEGACY_NAME_MAPPING[wizard.store["choose_your_npq"]])
```
Lockyy, line 57: We're repeating this in a lot of different places, is there any way to move this name to ID translation inside the question and leave all the other lookups the same. Holding this secondary primary key makes things more complex than it has to be.

Lockyy: We are actually still setting course_id in that question, so I'm not sure if we need to change these other lookups? We still have course_id.

shjohnson: Sorry only just saw this, easier option would be to have in the DB as new name or UI name

shjohnson: I think its also because havent really know how important this new text really is

app/lib/forms/choose_your_npq.rb
```diff
@@ -2,18 +2,41 @@ module Forms
   class ChooseYourNpq < Base
     include Helpers::Institution
 
-    attr_accessor :course_id
+    QUESTION_NAME = :choose_your_npq
```
Lockyy, line 5: At it's core, I'm not sure why we're making this change and adding in the LEGACY_NAME_MAPPING system. What does that accomplish?

We're not sending this across to ECF, we're not using it anywhere. Why are we not able to continue using the result from `courses` to populate the list of options and use the ID->Name pairing from the record for the radio value->name combo?

app/lib/services/handle_submission_for_store.rb
```diff
@@ -67,7 +67,7 @@ def raw_application_data
       # Cutting out confirmation keys since that is not application related data
       # Though I recognise that this means that even though this is meant to be raw
       # it still has a small layer of processing
-      store.except("generated_confirmation_code")
+      store.except("generated_confirmation_code", "choose_your_npq")
```
Lockyy, line 70: I'm not sure why this would be exempt from storage, this is still an answer the user entered and the goal of `raw_application_data` is to store a raw recording of exactly what happened in the users application, which would include this.

EmmaFrith:
Thanks @shjohnson this is looking good. Could you add the following hint text under this radio option please? 

'Structured support and networking opportunities if you’re in your first 5 years of head teaching.' 

Screenshot 2023-02-06 at 13 34 12

EmmaFrith:
Is it possible to reduce the size of this legend and add a bit more space between the sentence above and the legend @shjohnson?

Screenshot 2023-02-06 at 13 39 19

so that it looks like this: 
Screenshot 2023-02-06 at 13 41 07

app/views/registration_wizard/choose_your_npq.html.erb
```diff
@@ -1,47 +1,10 @@
...
-        <% end %>
-      <% end %>
-
-      <%= f.govuk_submit %>
-    <% end %>
-  </div>
-</div>
+<h1 class="govuk-heading-xl"><%= t("helpers.title.registration_wizard.choose_your_npq") %></h1>
+
+<p class="govuk-body">To register for noth the headship NPQ and the early headship coaching offer you need to submit 2 separate registrations. </p>
```
EmmaFrith, line 3:
```suggestion
<p class="govuk-body">To register for an NPQ and the <a href="https://professional-development-for-teachers-leaders.education.gov.uk/early-headship-coaching-offer" class="govuk-link">early headship coaching offer</a>, submit 2 separate registrations.</p>
```

app/lib/services/handle_submission_for_store.rb
shjohnson: I guess this makes a difference depending on the other comment.

app/lib/forms/aso_new_headteacher.rb
```diff
@@ -43,7 +43,7 @@ def options
   private
 
     def course
-      Course.find_by(id: wizard.store["course_id"])
+      Course.find(wizard.store["course_id"])
```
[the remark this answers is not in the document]
shjohnson: This is true and no it wasnt on purpose it was due to a refactor and not returning it to the original code

app/lib/forms/choose_your_npq.rb
shjohnson: Maybe im missing something but i had great difficulty with the auto-magic of the question generation and layout. IE this is more to do with headings layout of the question.

shjohnson: Going to PM you about this

shjohnson: I think i misunderstood what you meant when i was fixing it;

So issue we have is we need to identify the course and the question button only allows us to set one value on the radio button? previously we had the text value sent across aswell which was using the course name. So we cant change the course name in the DB as i understand but have to change the value which is currently linked to the course through the UI and ID? 

Going forward with this question format and how we are storing values/ answers etc i feel we need to make a decision to how this is achieved or maybe it already has. @Lockyy your raw application data makes a lot of sense but then we need to agree what is stored there? Does it matter that the ui is displaying different information to the actual course name? 

Im not sure using course ID limited to the DB is the best option as felt a code smell linking the code to a set order in the db? I feel the course name is unlikely to change that is offered by the provider and even if it does a little this should then be added as a new course so we keep a record of the previous choices rather than changing the existing ones.

shjohnson:
@EmmaFrith New screenshot by the way
Screenshot 2023-02-14 at 16 13 19

config/locales/en.yml
```diff
@@ -223,6 +225,8 @@ en:
           another: "For example, I am sharing the costs with my workplace"
         funding_options:
           another: "For example, I am sharing the costs with my workplace"
+        choose_your_npq_options:
+          early_headship_coaching_offer: "Structured support and networking opportunities if you’re in your first 5 years of head teaching."
```
EmmaFrith, line 229:
```suggestion
          early_headship_coaching_offer: "Structured support and networking opportunities if you’re a headteacher in your first 5 years of headship."
```

EmmaFrith (approved):

app/lib/forms/choose_your_npq.rb
Lockyy: I think we've discussed this separately but wanted to note, the raw_application_data is just the store as it was at the point of submission to act as an audit trail for what the user entered. It is not intended to be anything more complicated than an exact record of the store.

app/lib/forms/choose_your_npq.rb
```diff
@@ -2,18 +2,42 @@ module Forms
...
+      [
+        build_option_struct(value: "leading_behaviour_and_culture", link_errors: true),
+        build_option_struct(value: "leading_literacy"),
+        build_option_struct(value: "leading_teaching"),
+        build_option_struct(value: "leading_teacher_development"),
+        build_option_struct(value: "senior_leadership"),
+        build_option_struct(value: "headship"),
+        build_option_struct(value: "executive_leadership"),
+        build_option_struct(value: "early_years_leadership"),
+        build_option_struct(value: "early_headship_coaching_offer", divider: true),
```
Lockyy, line 34: It still feels weird to have the move to a system where we have the keys for courses stored in two locations. Why can't we use a map of courses? Like we previously had, or a map or something on the course model. 

As it stands we now have DISPLAY_NAME_MAPPING and here defining the available courses and the display column on the Course model is now ignored.

Lockyy (commented):
I still have questions about whether this is the best way to store/render out courses

app/lib/forms/choose_your_npq.rb
shjohnson: Well i changed it away from the ID mapping to the keys and unless that key is in the DB there is no way of linking up. 
Think im not understanding what you mean but your suggesting change it back to have id's in the values which then gets the content to be displayed in the `en.yml`.  Which would allow you to look up in the DB as used through most the tests and finding the course etc..

The code has keys for finding the data which arent added in the DB. I just felt linking id's in the code and having them map to courses in the db was a bit of a code smell. Ideally is more readable if the key means something and this is how its found in the db.

app/lib/forms/question_types/base.rb
```diff
@@ -3,9 +3,10 @@ module QuestionTypes
     class Base
       attr_reader :name, :options
 
-      def initialize(name:, options: [])
+      def initialize(name:, options: [], opts: {})
         @name = name
         @options = options
```
Lockyy, line 8: Would `opts = {}` be better than `opts: {}`? That way we aren't declaring `opts:..., options:...` and opts is purely an internal name.

app/views/registration_wizard/shared/_question_page.html.erb
```diff
@@ -19,7 +19,8 @@
         render(
           "registration_wizard/shared/questions/#{form.question.type}",
           form: f,
-          question: form.question
+          question: form.question,
+          fieldset_legend_attributes: form.question.try(:fieldset_legend_attributes)
```
Lockyy, line 23: `form.question` is already being passed in, so we don't need to pass this in separately and we'd be able to grab this only where we know it's needed, i.e. inside `_radio_button_group.html.erb`, and not need to use try, might be a little cleaner to only reference this where necessary

config/locales/en.yml
```diff
@@ -238,6 +242,17 @@ en:
...
+        course_id_options:
+          2: "Leading behaviour and culture"
+          3: "Leading literacy"
+          4: "Leading teaching"
+          5: "Leading teacher development"
+          6: "Senior leadership"
+          7: "Headship"
+          8: "Executive leadership"
+          9: "Early years leadership"
+          10: "Early headship coaching offer"
```
Lockyy, line 254: You're right that using IDs isn't ideal, we should return to this and implement your course.key system so that we can stop using primary keys anywhere. It'd also let us stop referencing ECF UUIDs within the NPQ app at all if we were just passing keys over to ECF and it used that to look up records. 

It'd allow us to cut that hard link between database records!

Lockyy (approved):
Ship it

EmmaFrith:
Hey @shjohnson @Lockyy, something weird has happened to the H1....

Screenshot 2023-02-17 at 11 57 24

It should be this:


Screenshot 2023-02-17 at 11 58 05

EmmaFrith (changes requested):
something weird going on with the H1, see my comment in thread

EmmaFrith:
Hmm, the page where you choose a provider appears to have disappeared - I'm looking in the review app btw. Normally, once you select an NPQ, you're asked to select a provider. I'm not getting that question in the review app @shjohnson

EmmaFrith (approved):

PR #677, opened 2023-01-30
Remove uckfield from targeted delivery funding

shjohnson:
### Context
Ensure Uckfield college does not get targeted delivery funding


shjohnson:
Closing as needs further discussion about delivery vs support funding

PR #678, opened 2023-01-31
Remove uckfield from ukprn list 

shjohnson:
Remove Uckfield from list and ensure old column is removed

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_01_06_114221) do
+ActiveRecord::Schema.define(version: 2023_01_31_142734) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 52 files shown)


db/schema.rb
```diff
@@ -105,6 +106,16 @@
...
+  create_table "itt_providers", force: :cascade do |t|
+    t.text "legal_name"
+    t.text "operating_name"
+    t.datetime "removed_at"
+    t.boolean "approved"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["legal_name"], name: "index_itt_providers_on_legal_name", unique: true
+  end
+
```
peteryates, line 118: I don't see a corresponding migration for this change in the PR.

shjohnson: oh wow this is a weird commit that came across somewhow!

shjohnson: rebasing issue 👍

Lockyy (approved):

PR #688, opened 2023-02-08
[CPDNPQ-881] Add job and tasks for syncing get_an_identity_ids

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-881

We need to sync historical GAI IDs over to the ECF DB.

Adds three jobs:

- users:get_an_identity_id_sync:stats 
  Output to the log stats on total users marked as synced and total users marked as unsynced
- users:get_an_identity_id_sync:all 
  Enqueues background jobs to sync get_an_identity_ids for all users with ecf_ids, get_an_identity_ids, and that have not previously been synced
- users:get_an_identity_id_sync:user 
  Enqueues a background job to sync the user with the provided ID, assuming the record has an ecf_id and a get_an_identity_id


Gemfile
```diff
@@ -6,6 +6,7 @@ ruby File.read(".ruby-version").chomp
 gem "activerecord-session_store"
 gem "bootsnap", ">= 1.1.0", require: false
 gem "canonical-rails"
+gem "delayed_job"
```
Lockyy, line 9: Already included, this just adds it explicitly into the gem file rather than inferring it from "sentry-delayed_job"

db/migrate/20230206153448_add_get_an_identity_id_sync_bool_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddGetAnIdentityIdSyncBoolToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :get_an_identity_id_synced_to_ecf, :boolean, default: false
```
Lockyy, line 3: Once the sync is completed and the data is flowing across properly this will be removed

Gemfile
shjohnson: can we lock to a version though?

shjohnson: `"~> 4.1"` or something?

config/initializers/delayed_job.rb
```diff
@@ -0,0 +1,4 @@
+Delayed::Worker.queue_attributes = {
+  default: { priority: 0 },
+  low_priority: { priority: 10 },
+}
```
shjohnson, line 4: Nice!

app/jobs/get_an_identity_id_sync_job.rb
```diff
@@ -0,0 +1,67 @@
...
+      create_successful_ecf_sync_request_log(user:, message: "User's get_an_identity_id already synced to ECF")
+    else
+      # Sync get_an_identity_id to ECF
+      ecf_user.update!(get_an_identity_id: user.get_an_identity_id)
+
+      create_successful_ecf_sync_request_log(user:, message: "User's get_an_identity_id synced to ECF")
+    end
+
+    # Mark user as synced so that this user gets skipped in future sync runs
+    user.update_column(:get_an_identity_id_synced_to_ecf, true)
```
shjohnson, line 23: I know we have in the rake task but do we want to add protection in this class to skip if a user is passed in with column true?

shjohnson (approved):
One comment non-blocker LGTM

app/jobs/get_an_identity_id_sync_job.rb
Lockyy: I considered this, I decided to leave it without protections because in the case we want to manually kick off this job it would be a blocker to have it auto-skip if this was true. There's no *real* downside to letting it run, it's purely used in the rake task for speeding things up.

PR #692, opened 2023-02-16
[CPDNPQ-806] Add interface for managing admins

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-806

Rather than requiring DB access for managing regular admins it would be better to have an interface for managing them. This would also allow us to proactively monitor who has admin access.

- [x] Replace `flipper_admin_access` with `super_admin`
- [x] Add admin interface for managing admins
- [x] Add ability to create new admins as a super admin
- [x] Add ability to remove admins as a super admin
- [x] Add ability to create new super admins as a super admin


app/views/admin/schools/index.html.erb
```diff
@@ -1,13 +1,6 @@
-<% content_for :before_content do %>
```
shjohnson, line 1: What was the reason to remove this?

db/migrate/20230105160357_rename_flipper_admin_access_to_superadmin.rb
```diff
@@ -0,0 +1,7 @@
+class RenameFlipperAdminAccessToSuperadmin < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :super_admin, :boolean, default: false, null: false
+
+    User.where(flipper_admin_access: true).update_all(super_admin: true)
```
shjohnson, line 5: Is flipper access used anywhere else? if not maybe we can delete the old column in this migration?

spec/lib/services/handle_submission_for_store_spec.rb
```diff
@@ -44,7 +44,7 @@
 
     describe "#call" do
       def stable_as_json(record)
-        record.as_json(except: %i[id created_at updated_at])
+        record.as_json(except: %i[id created_at updated_at flipper_admin_access])
```
shjohnson, line 47: If we delete this field then wouldnt have to do this aswell

shjohnson (commented):
Great to get this up!

app/views/admin/schools/index.html.erb
Lockyy: I've moved this interface to use the header bar tabs so there's no need for a back button when this is a top level page within the nav bar that's always present.

db/migrate/20230105160357_rename_flipper_admin_access_to_superadmin.rb
Lockyy: I've got a follow up PR to do that, the migration runs before the deployment so we'd have a moment where a column may be accessed that no longer exists. This PR deprecates then a follow up removes the no longer used column.

spec/lib/services/handle_submission_for_store_spec.rb
Lockyy: responded above

app/controllers/admin/admins_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  def index
+    @pagy, @users = pagy(scope)
+  end
+
+  def new
+    @user = User.new
+  end
+
+  def create
+    @user = User.find_or_initialize_by(email: user_params[:email])
```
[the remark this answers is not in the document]
Lockyy: We don't, and I should've removed that from the user_params since it's not a part of the form. I should go back and add that into the form.

shjohnson (approved):
LGTM

PR #696, opened 2023-02-20
[CPDNPQ-808] Reapply changes from #676 and move to using course identifiers instead of primary keys

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-808

#676 introduced issues due to using DB primary keys for locale lookups. Because of this we had to revert the PR.

We still want those changes but we need to sort the fact we're using DB specific IDs for locale strings.

This PR puts the #676 code back along with a set of changes to move from using `courses.id` around the app for question values to instead using `courses.identifier`.

`courses.identifier` is a new column populated using `COURSE_ECF_ID_TO_IDENTIFIER_MAPPING`. The identifiers that are populated are the same identifiers that are in the ECF database. By moving to this for the unique ID, we can begin the process of moving to no longer having to care about ECF_IDs and specific foreign/primary keys for course identification and instead use these unique string based identifiers.

Most importantly here, this means that within the NPQ app we are always referencing the same set of identifiers in the HTML which then get used for locale population.


db/migrate/20230106114221_update_default_cohort_for_ehco.rb
```diff
@@ -1,6 +1,6 @@
 class UpdateDefaultCohortForEhco < ActiveRecord::Migration[6.1]
   def change
-    ehco_course = Course.find_by_code(code: :EHCO)
+    ehco_course = Course.find_by(identifier: "npq-early-headship-coaching-offer")
```
Lockyy, line 3: This has already been run, but find_by_code has been removed so I had to update this.

config/initializers/courses.rb
```diff
@@ -0,0 +1,76 @@
...
+      position: 0,
+      name: "Additional Support Offer for new headteachers",
+      ecf_id: "7fbefdd4-dd2d-4a4f-8995-d59e525124b7",
+      identifier: "npq-additional-support-offer",
+      description: "The Additional Support Offer is a targeted support package for new headteachers.",
+      display: false,
+    },
+    {
+      position: 1,
+      name: "NPQ for Leading Behaviour and Culture (NPQLBC)",
```
Lockyy, line 13: These names are all still the old names. They aren't used anywhere in the interface however they are used in emails and the admin interface. To avoid making unexpected changes I've left the names as is, then we can return after and replace those calls to `.name` with separate localisations and remove the name/description columns entirely.

app/lib/forms/choose_your_npq.rb
```diff
@@ -2,15 +2,33 @@ module Forms
...
+    def question
+      Forms::QuestionTypes::RadioButtonGroup.new(
+        name: :course_identifier,
+        options:,
+        style_options: { fieldset: { legend: { size: "m", tag: "h1" } } },
+      )
+    end
+
+    def options
+      divider_index = courses.length - 1 # Place the "Or" divider before the last course
```
Lockyy, line 25: This still isn't great because it implicitly includes the assumption that the EHCO course is last, since the logic is that it's placed before EHCO (which is last) rather than before whichever NPQ is last.

spec/features/disabled_get_an_identity_integration/sad_journeys/works_in_childcare_but_not_in_england_spec.rb
```diff
@@ -63,8 +63,8 @@
     School.create!(urn: 100_000, name: "open manchester school", address_1: "street 1", town: "manchester", establishment_status_code: "1")
 
     expect_page_to_have(path: "/registration/choose-your-npq", submit_form: true) do
-      expect(page).to have_text("What are you applying for?")
-      page.choose("NPQ for Senior Leadership (NPQSL)", visible: :all) # Needs changing to an early years course once added
+      expect(page).to have_text("Which NPQ do you want to do?")
+      page.choose("Senior leadership", visible: :all) # Needs changing to an early years course once added
```
Lockyy, line 67: We should probably do what this comment tells us to? 👀

spec/lib/forms/choose_your_provider_spec.rb
```diff
@@ -41,28 +41,26 @@
     end
 
     npqeyl_and_npqll_codes = %w[
-      NPQEYL
```
Lockyy, line 44: Looking back over this makes it clear we had identifiers all along, they were just in the code and stored in COURSE_NAMES. Which wasn't clear.

Lockyy: The new system is a far better identifier system since it uses the correct identifier codes everywhere.

spec/rails_helper.rb
```diff
@@ -92,16 +92,6 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
-
-  config.before(:suite) do
-    Rails.application.load_seed
```
Lockyy, line 97: Took me too long to find this so I moved it, and the feature flag setup below, into their own support files. Plus decoupled from the actual seeds file and instead call the specific seeders we want.

app/lib/forms/aso_new_headteacher.rb
```diff
@@ -43,7 +43,7 @@ def options
   private
 
     def course
-      Course.find_by(id: wizard.store["course_id"])
+      wizard.query_store.course
```
Lockyy, line 46: As the first step to changing how we locate courses I replaced all manual lookups for courses using the course_id into the query_store. That way, once the user is selecting course identifiers (e.g. "npq-executive-leadership") instead of course IDs (e.g. 3) we only need to change one lookup location.

app/lib/forms/choose_your_npq.rb
```diff
@@ -2,15 +2,33 @@ module Forms
   class ChooseYourNpq < Base
     include Helpers::Institution
 
-    attr_accessor :course_id
+    QUESTION_NAME = :course_identifier
```
Lockyy, line 5: The course_id question goes away, replaced with course_identifier. Now users will select from the same options as previously except the value on each is a unique identifier that is consistent across databases and uncoupled from primary keys. You can see a list of courses and their identifiers in config/initializers/courses.rb.

These identifiers can then be used in l18n files reliably since they are always the same. The unreliability of primary keys is what lead to previous bugs in #676.

app/models/course.rb
```diff
@@ -1,63 +1,29 @@
 class Course < ApplicationRecord
```
Lockyy, line 1: We have now entirely removed references to EECF UUIDs from the model, these were previously in the code to act as a translation layer to obtain identifiers which were used to link to LeadProviders. Now that identifiers are right on the model, that can go.

Along with that COURSE_NAMES can go. That was being used as a pseudo-identifier as a way to link course records into the codebase. If you look you'll see lines like `ASO: "Additional Support Offer for new headteachers`. This allowed us to do things like `Course.find_by(code: :ASO)` to pull specific records out. Now that identifiers are properly in place in the DB we don't need to use this code->name translation layer and can instead query directly on identifiers. Like we are now doing in the scopes below.

Lockyy: COURSE_ECF_ID_TO_IDENTIFIER_MAPPING was used in the LeadProviders model but now instead we are using the identifiers as a key for filtering available providers.

It would still be ideal to remove that and replace it with a proper relationship between the two tables.

app/models/registration_wizard.rb
```diff
@@ -150,7 +150,7 @@ def answers
     end
 
     array << OpenStruct.new(key: "Course",
-                            value: query_store.course.name,
+                            value: I18n.t(query_store.course.identifier, scope: "helpers.label.registration_wizard.course_identifier_options"),
```
Lockyy, line 153: Makes sure that the option the user sees on the CYA page matches the option they selected.

app/lib/forms/choose_your_npq.rb
peteryates: Is there a way of passing the 'or' option in separately? I think that'd make it easier to follow. Maybe as `options:, alternative_option:` or something?

shjohnson: I understand this option makes it easy to change order through the DB rather than a code change although as i understand this doesnt really happen that often. I guess i still would have been a fan of having all the radio options listed here unless we think it looks messy? Its just trying to control things like the divider and anything else that might happen seems too complex for a simple list? I might be in the minority camp here though?

shjohnson (approved):
I have approved but is one comment although i think should also have Peter go over because im not in tomorrow and Naomi and myself have both worked on this so good to get another eyes if possible?

app/lib/forms/aso_new_headteacher.rb
peteryates: This is much nicer.

app/lib/forms/choose_your_npq.rb
peteryates: I don't think it being a code change is necessarily a bad thing, so long as a new dev in a year's time can easily work out what's going on. However, there's no reason that can't be addressed as a follow up.

config/locales/en.yml
```diff
@@ -238,6 +242,18 @@ en:
...
+          "npq-leading-behaviour-culture": "Leading behaviour and culture"
+          "npq-leading-teaching-development": "Leading teacher development"
+          "npq-senior-leadership": "Senior leadership"
+          "npq-headship": "Headship"
+          "npq-executive-leadership": "Executive leadership"
+          "npq-additional-support-offer": "Additional Support Offer"
+          "npq-early-headship-coaching-offer": "Early headship coaching offer"
+          "npq-early-years-leadership": "Early years leadership"
+          "npq-leading-literacy": "Leading literacy"
+
```
peteryates, line 256: Is there a benefit to these options being here rather than in `Courses::DEFINITIONS`?

peteryates (approved):
Looks good overall, happy for it to go in if it helps solve the bug.

It does feel like we have information about courses spread across a few places; the `DEFINITIONS` constant, the courses table and the locales. If I was coming at this without prior knowledge I think that would take a bit of time to work out.

config/locales/en.yml
Lockyy: I was trying to keep localisations distinct from the data definitions, data layer vs. view layer. The names/descriptions in Courses::Definitions atm is going to go away once the name/description column are dropped from the DB.

I can certainly see the benefit of it being fully contained, but the identifier->localised string will become more complex once they're used for rendering out into emails and the admin console as well. It would get quite complicated in the definitions once we have to have multiple localised strings.

config/initializers/courses.rb
```diff
@@ -0,0 +1,76 @@
+module Courses
```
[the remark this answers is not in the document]
Lockyy: That feels like it would be slightly neater yeah.

PR #707, opened 2023-02-28
Combine journey specs

peteryates:
### Context

This change is an attempt to remove duplication from the features. We currently have features split in the following ways:

  * JS/non-JS
  * happy/sad
  * feature flag enabled/disabled

The majority of these features overlap so it would make sense for us to remove duplication by adding a little logic to the specs themselves. We can also lean on RSpec's tags to run them in batches.

- Move basic reg journey to generic journeys dir
- Make the generic journey specs pass


spec/features/journeys/basic_registration_journey_spec.rb
```diff
@@ -3,12 +3,22 @@
...
 
-  scenario "registration journey" do
+  context "when JavaScript is enabled", :js do
+    scenario("registration journey (with JS)") { basic_registration_journey(js: true) }
+  end
+
+  context "when JavaScript is disabled", :no_js do
+    include_context "use rack_test driver"
+    scenario("registration journey (without JS)") { basic_registration_journey(js: false) }
+  end
```
peteryates, line 19: Use RSpec's tags to be able to easily target the ones we want. I'll probably disable `:js` ones locally most of the time as they take hundreds of times longer than the `:no_js` ones

spec/features/journeys/basic_registration_journey_spec.rb
```diff
@@ -200,7 +195,7 @@
         "funding" => "trust",
         "institution_identifier" => "School-100000",
         "institution_location" => "manchester",
-        "institution_name" => "",
+        "institution_name" => js ? "" : "open", # FIXME: i have no idea why this is different
```
peteryates, line 198: We appear to write an extra attribute when picking a school without JS, I guess we submit a search rather than instigating a JS autocompletion.

spec/features/journeys/basic_registration_journey_spec.rb
```diff
@@ -30,6 +40,10 @@
       page.choose("Yes", visible: :all)
     end
 
+    unless js
+      expect_page_to_have(path: "/registration/get-an-identity", submit_form: true)
+    end
```
peteryates, line 45: Extra non-JS only step. Not a fan of a leading `unless` but when it's inline it might get lost over there on the right!

spec/features/journeys/basic_registration_journey_spec.rb
```diff
@@ -3,12 +3,22 @@
...
+  context "when JavaScript is enabled", :js do
+    scenario("registration journey (with JS)") { basic_registration_journey(js: true) }
+  end
+
+  context "when JavaScript is disabled", :no_js do
+    include_context "use rack_test driver"
+    scenario("registration journey (without JS)") { basic_registration_journey(js: false) }
+  end
+
+  def basic_registration_journey(js:)
```
peteryates, line 21: Probably would make more sense to call this `javascript_enabled` or something.

spec/support/helpers/journey_step_helper.rb
```diff
@@ -63,6 +63,38 @@ def choose_a_childcare_provider(js:, location:, name:)
...
+          within ".npq-js-reveal" do
+            page.fill_in "private-childcare-provider-picker", with: provider.urn
+          end
+
+          expect(page).to have_content("#{provider.urn} - #{provider.name} - #{provider.address_1}, #{provider.town}")
+
+          page.find("#private-childcare-provider-picker__option--0").click
+        end
+      else
+        # FIXME: This doesn't work, I think the real journey might be broken in prod. There's
```
peteryates, line 88: cc @Lockyy - would you mind taking a look at this with @shjohnson if you get a few spare minutes? Got a bit stuck here last week.

peteryates:
I don't have the capacity to finish this off. @Lockyy @shjohnson - if you think it's worth pursuing please try and factor it in, I think it'll make every subsequent journey change miles faster. If not, I'll close it down.

spec/features/journeys/basic_registration_journey_spec.rb
BroiSatse: The comment is correct - institution_name is populated via a no-js fallback search

spec/support/helpers/journey_step_helper.rb
BroiSatse: It is now fixed :)

app/lib/forms/choose_private_childcare_provider.rb
```diff
@@ -48,7 +48,7 @@ def no_js_fallback_search_loop?
 
     def validate_private_childcare_provider_name_returns_results
       if display_no_javascript_fallback_form? && possible_institutions.blank?
-        errors.add(:institution_name, :no_results, location: institution_location, name: institution_name)
+        errors.add(:institution_name, :no_results, urn: institution_name)
```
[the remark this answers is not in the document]
shjohnson: Or actually maybe it is coming through as urn

shjohnson (approved):
Seems good though!

app/lib/forms/choose_private_childcare_provider.rb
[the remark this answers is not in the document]
BroiSatse: Yeah, the naming there was quite a bit confusing, I assume for historical reason. I'd love to revisit that step at some point for some major refactor/cleanup as there are quite a few edgy cases we don't handle well (for example in the js journey there is no presence validation errors displayed on institution select even though we prevent user from progressing - this seems intentional and requires some more digging in)

Lockyy (approved):

PR #710, opened 2023-03-03
[CPDNPQ-883] Integrate new previously funded API to send GAI to ECF for previously funded checks

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-883

We need to prepare for a system where the Get an Identity service does not have to send us TRNs for every user. To support this the ECF app now has an API for checking previous funding that doesn't require a TRN. We need to integrate with that and send TRNs/Get an Identity IDs when available to support the differing availability of each.

### DIFF
--- a/app/lib/forms/choose_your_provider.rb
+++ b/app/lib/forms/choose_your_provider.rb
@@ -58,6 +58,7 @@ module Forms
         inside_catchment: inside_catchment?,
         new_headteacher: new_headteacher?,
         trn:,
+        get_an_identity_id:,
       ).funded?
     end
 
@@ -74,6 +75,7 @@ module Forms
              :inside_catchment?,
              :new_headteacher?,
              :trn,
+             :get_an_identity_id,
              to: :query_store
 
     def validate_lead_provider_exists
--- a/app/lib/services/funding_eligibility.rb
+++ b/app/lib/services/funding_eligibility.rb
@@ -22,12 +22,19 @@ module Services
     # Lead Mentor
     NOT_LEAD_MENTOR_COURSE = :not_lead_mentor_course
 
-    attr_reader :institution, :course, :trn, :approved_itt_provider, :lead_mentor, :employment_role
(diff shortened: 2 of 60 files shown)


app/lib/services/funding_eligibility.rb
```diff
@@ -133,7 +134,11 @@ def eligible_urns
     end
 
     def ecf_api_funding_lookup
-      @ecf_api_funding_lookup = EcfApi::NpqFunding.with_params(npq_course_identifier: course.identifier).find(trn)
+      @ecf_api_funding_lookup = EcfApi::Npq::PreviousFunding.find_for(
+        trn:,
+        get_an_identity_id:,
+        npq_course_identifier: course.identifier,
+      )
```
Lockyy, line 141: I am planning on moving all this into the query store and pass the info into this service by just giving it the query store. This is a bigger refactor though so I'll be doing a spike for that separately.

spec/features/admin_spec.rb
```diff
@@ -187,14 +187,16 @@ def sign_in_as_admin
...
     # Test application pagination
-    users[0..2].each do |user|
+    display_users[0..2].each do |user|
       expect(page).to have_content(user.email)
     end
 
     page.find("[aria-label=next]").click
 
-    users[3..].each do |user|
+    display_users[3..].each do |user|
```
Lockyy, line 199: I think this should solve the flakey failure we've got.

app/lib/forms/choose_your_provider.rb
```diff
@@ -58,6 +58,7 @@ def eligible_for_funding?
         inside_catchment: inside_catchment?,
         new_headteacher: new_headteacher?,
         trn:,
+        get_an_identity_id:,
```
shjohnson, line 61: whats the reason not getting this from the `query_store` maybe im missing something

shjohnson (approved):
One question but otherwise makes sense

PR #711, opened 2023-03-08
Enable Codespaces for NPQ

peteryates:
Enabled Codespaces for the app.

This should let content/interaction designers run the application without having to go through the process of getting Ruby, Node, PostgreSQL etc running locally. I've tested it and it works with a single click. Managed to get this going while on a call!

Code/approach borrowed from [this starter repo](https://github.com/luizkowalski/devcontainer-rails).

image


shjohnson (approved):
LGTM 🍏

PR #718, opened 2023-03-20
Update the FE list

shjohnson:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-956

Update the FE list after policies requirement


Lockyy (approved):
Jake has signed off on the updated list being correct.

PR #720, opened 2023-03-24
[CPDNPQ-860] Change TSF in logic

shjohnson:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-860

This change is to include changes regarding the new `tsf_primary_eligibility` and `tsf_primary_plus_eligibility` outcomes for the application. 

Addition of the following fields to the application model:
- `primary_establishment`
- `number_of_pupils`
- `tsf_primary_eligibility`
- `tsf_primary_plus_eligibility`
 
Addition of the following fields to the school model:
- `phase_type`
- `phase_name`

Update the GIAS import script to update the new school fields when importing data from GIAS although the default will be set as `"Not applicable"` before these are updated

Display the new `school_phase_name` in the school admin view: `app/views/admin/schools/show.html.erb`

Create a new centre point for all the `targeted support funding calculations` (TSF) `Services::Eligibility::TSFCalculator` which will return the results in a hash format like so:
```
{
    tsf_primary_eligibility: true/false,
    tsf_primary_plus_eligibility: true/false,
    previously_received_targeted_funding_support: true/false,
    targeted_delivery_funding: true/false
}
```
These will only be resolved when required to reduce any un-needed calculations

Details fo the new flow:
https://lucid.app/lucidspark/323f8cf1-e51e-4992-a68b-6dfdc71bdd91/edit?invitationId=inv_d04d827e-52de-4656-a90f-3eaf8b75f76f#

Things to be aware and possibly discuss;
 - Moved the establishment type checking into the `school` model as this is used in two places now
 - If an institution is eligible and primary it will always return true for `tsf_primary_eligibility` and `targeted_delivery_funding`
 - We now are using both `targeted_delivery_funding` in a class name and then `tsf_` as a prefix for the new checking, might be better to unify these.
 - Need to confirm the field names coming from GIAS data for phase name, ECF has been checked. 
 - double check fields sent to ECF and also import school data from ECF for school phase


app/models/school.rb
```diff
@@ -42,4 +49,45 @@ def in_england?
...
     "School-#{urn}"
   end
+
+  def tsf_eligible_establishment?
+    tsf_eligible_establishment_type_codes.include?(establishment_type_code)
+  end
+
+private
+
+  def tsf_eligible_establishment_type_codes
```
Lockyy, line 59: Would love to move this into the DB so we can let Jake manage it via the admin console.

app/models/school.rb
```diff
@@ -42,4 +49,45 @@ def in_england?
   def identifier
     "School-#{urn}"
   end
+
+  def tsf_eligible_establishment?
```
Lockyy, line 53: Much better place for this 🙌

app/lib/services/eligibility/tsf_calculator.rb
```diff
@@ -0,0 +1,61 @@
+module Services
+  module Eligibility
+    class TsfCalculator
```
Lockyy, line 3: With the two types of funding coming through this would a more top level name make it clearer that it provides info on both targeted support funding and targeted delivery funding?

`Services::Eligibility::TargetedFundingCalculator`? 

Initially I spent a bit thinking this one worked exclusively on targeted support funding and was confused about the presence of targeted_delivery_funding as a key.

Also, I'm a bit unsure of introducing a new acronym here (tsf), is there a reason not to spell it fully out as targeted_support_funding like how targeted_delivery_funding is fully spelled out?

app/lib/services/funding_eligibility.rb
```diff
@@ -88,9 +88,24 @@ def funding_eligiblity_status_code
       end
     end
 
-    def eligible_for_targeted_delivery_funding?
-      course_and_institution_eligible_for_targeted_delivery_funding? &&
-        !previously_received_targeted_funding_support? # Check last as it involves an API call
+    def tsf_eligibility_calculator
+      # This returns the tsf calculation plus the previously received funding result from ECF:
```
Lockyy, line 92: I don't think this is accurate since previously_received_targeted_funding_support? is taken into account outside this class now.

app/lib/services/handle_submission_for_store.rb
```diff
@@ -206,12 +218,33 @@ def funding_eligibility
...
+      tsf_elgibility[:tsf_primary_eligibility]
+    end
+
+    def tsf_primary_plus_eligibility
+      tsf_elgibility[:tsf_primary_plus_eligibility]
+    end
+
+    def tsf_elgibility
+      # idea is to return a result with the results of the eligibility
+      # result will be (maybe with the previous funding result):
```
Lockyy, line 234: Is this out of date (since we don't include it)

spec/support/helpers/journey_helper.rb
```diff
@@ -26,5 +26,19 @@ def stub_participant_validation_request(trn: "1234567", date_of_birth: "1980-12-
         )
         .to_return(status: 200, body: participant_validator_response(**response), headers: {})
     end
+
+    def stub_previously_funded_request(get_an_identity_id:, npq_course_identifier:, trn: "1234567")
```
Lockyy, line 30: There may already be a helper in place that covers this, I added it in a recent PR, `mock_previous_funding_api_request` stubs this and lets you set a body if you want one.

shjohnson: Some reason thought this wasnt available but will double check

app/lib/services/funding_eligibility.rb
shjohnson: yeah need to sort the comment i guess is what you mean?

Lockyy (approved):

PR #721, opened 2023-03-24
Improve seeds

shjohnson:
### Context

Its very annoying having to log in to the admin panel without an admin user so this creates an admin user plus `PrivateChildcareProvider` users. There are lots other things we can add to seeds but this provides benefit with getting easy access to the backend through the `admin@example.com` user which will only work in development at the moment. 

[What has been changed?]

### DIFF
--- a/app/lib/forms/sign_in.rb
+++ b/app/lib/forms/sign_in.rb
@@ -21,6 +21,8 @@ module Forms
     end
 
     def after_save
+      return if skip_sending_otp
+
       user = User.find_by(email:)
 
       if user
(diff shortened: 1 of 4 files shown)


app/lib/forms/sign_in.rb
```diff
@@ -30,5 +32,9 @@ def after_save
         ConfirmEmailMailer.confirmation_code_mail(to: email, code:).deliver_now
       end
     end
+
+    def skip_sending_otp
+      Rails.env.development? && @email == "admin@example.com"
```
shjohnson, line 37: Realise this isnt ideal and maybe we can find a better way but thought simple enough for now

shjohnson:
Closing in favour of changes here: https://github.com/DFE-Digital/npq-registration/pull/789

PR #723, opened 2023-03-29
[CPDNPQ-903] Add rake task for syncing self-serve GAI data changes through to NPQ and ECF

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-903

The Get an Identity service provides a self-serve system for updating your name and email. Until we implement webhooks however, this data won't sync through to the NPQ database. And even once that is implemented, historical changes won't pull through. 

We need something that will pull data from GAI, save it to NPQ, and then push it to ECF.

- Adds models for interacting with GAI API
- Adds rake tasks for performing the GAI -> NPQ -> ECF sync on:
  - All GAI users
  - All GAI users that have never been synced
  - A specific user


app/models/get_an_identity/user.rb
```diff
@@ -0,0 +1,54 @@
...
+      self.load
+    end
+
+  private
+
+    def request_user
+      url = "#{ENV.fetch('TRA_OIDC_DOMAIN')}/api/v1/users/#{id}"
+      headers = { "Authorization" => "Bearer #{GetAnIdentity::AccessToken.new}" }
+
+      HTTParty.get(url, headers:).tap do |response|
```
Lockyy, line 31: I was going to use JsonApiClient like the ECF integration but I had to start building a parser since GAI doesn't use the `{ "data": { "type"/"attributes"/"id": .. }` formatting for their responses so I went with just hitting the API with HTTParty and parsing manually in load like this since overall it led to code that was easier to understand.

lib/tasks/users/get_an_identity_syncing.rake
```diff
@@ -44,4 +44,46 @@ namespace :users do
       end
     end
   end
+
+  namespace :get_an_identity_data_sync do
```
shjohnson, line 48: Will this get used that often? as its only for one user right?

app/lib/services/get_an_identity/user_updater.rb
```diff
@@ -0,0 +1,34 @@
...
+      def call
+        user = User.first
+        get_an_identity_user = ::GetAnIdentity::User.find(user.get_an_identity_id)
+
+        # GetAnIdentity::User#uid may not be the same as User#get_an_identity_id
+        # This is because while those two values both refer to the same thing,
+        # in cases of deduping we may get a new uid from GetAnIdentity when the user we
+        # looked up has been merged into another.
+        # By setting that new UID on the user, we can replicate that result within NPQ
+        user.update!(
```
shjohnson, line 23: Just a thought is there any chance of us ever needing a record of the previous data? Should we create a history record for that data?

app/lib/services/ecf/ecf_user_updater.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      def call
+        remote = user.ecf_user
+
+        # JsonApiClient::Resource uses errors for flow control, so failed saves
+        # will divert to the rescue block below
+        # I'd prefer to use the return value of save, but that's not possible
+        remote.email = user.email
+        remote.full_name = user.full_name
+        remote.get_an_identity_id = user.get_an_identity_id
```
shjohnson, line 22: Would this record the errors if the remote is not about to save just one of the fields and not the last one?

app/models/get_an_identity/access_token.rb
```diff
@@ -0,0 +1,21 @@
+module GetAnIdentity
+  class AccessToken
+    attr_reader :token
+
+    def initialize
+      client_id = ENV.fetch("TRA_OIDC_CLIENT_ID")
```
shjohnson, line 6: How do we normally handle env variables like this? ie storing them on server more for my information

lib/tasks/users/get_an_identity_syncing.rake
shjohnson: Also think description could be updated

app/models/get_an_identity/user.rb
```diff
@@ -0,0 +1,54 @@
+module GetAnIdentity
```
shjohnson, line 1: Im not sure if this should be a model? feels like this should be a service as your fetching a user? Your not creating etc..

.github/workflows/get_an_identity_data_sync_all.yml
```diff
@@ -0,0 +1,31 @@
+name: Resync User data for all GAI users
```
shjohnson, line 1: How do we normally test this configuration? or other github actions? once on dev or something?

app/models/get_an_identity/user.rb
shjohnson: Also confusing as this is called from the services folder right?

spec/lib/services/ecf/ecf_user_updater_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+          data: {
+            type: "user",
+            id: "123",
+          },
+        }
+      end
+
+      it "updates the user" do
+        subject.call
+        expect(a_request(:patch, "https://ecf-app.gov.uk/api/v1/npq/users/#{ecf_id}")).to have_been_made.once
```
shjohnson, line 90: 👍

spec/models/get_an_identity/user_spec.rb
```diff
@@ -0,0 +1,112 @@
+require "rails_helper"
+
+RSpec.describe GetAnIdentity::User do
+  describe "#find" do
+    # specs for find method
```
shjohnson, line 5: Are these comments meant to be here?

spec/models/get_an_identity/user_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        expect(loaded_user.email).to eq "mail@example.com"
+        expect(loaded_user.full_name).to eq "Jane Doe"
+        expect(loaded_user.date_of_birth).to eq Date.new(1990, 1, 1)
+        expect(loaded_user.trn).to eq "1234567"
+        expect(loaded_user.mobile_number).to eq ""
+        expect(loaded_user.merged_user_ids).to eq []
+        expect(loaded_user.raw).to eq response_body
+      end
+
+      context "when uid from response does not match the requested uid" do
```
shjohnson, line 68: Does test actually test anything as couldnt work it out from the description and your stubbing the response so the response_uid is always going to be whatever you set it as right?

shjohnson (commented):
Few comments but otherwise i think nearly there 🍏

Gemfile
```diff
@@ -37,6 +37,8 @@ gem "sentry-ruby"
 gem "stimulus-rails"
 gem "webpacker"
 
+gem "httparty"
```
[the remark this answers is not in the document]
Lockyy: Yeah good point, fixed at ~> 0.21

app/lib/services/get_an_identity/user_updater.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      attr_reader :user
+
+      def initialize(user:)
+        @user = user
+      end
+
+      def call
+        user = User.first
```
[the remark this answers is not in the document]
Lockyy: Whoops yes, this was from some manual testing to ensure the code worked between local instances. Thanks for catching this.

lib/tasks/users/get_an_identity_syncing.rake
Lockyy: I added it mainly as a backup, in case we get a few individual failures or we get asked to push through changes for a specific user we can use this to sort it immediately rather than needing to use the console at all. Might not be used often but it'll be really helpful if we ever need it.

app/lib/services/get_an_identity/user_updater.rb
Lockyy: Getting a gem installed so we can have an audit trail of record changes would be a really good idea, ECF has that I think. We should pitch it and see if there're any implications re: data protection.

app/lib/services/ecf/ecf_user_updater.rb
Lockyy: Can you clarify what you mean? The API is an all or nothing update, it won't update anything unless the record is valid.

app/models/get_an_identity/access_token.rb
Lockyy: They are stored in Github secrets then the Github Actions pass them into the `cf push` call in each deployment using the --var parameter. This then gets picked up in the manifests and set as an env variable on the server. Which is picked up here. 

These ones are already all set up. :)

app/models/get_an_identity/user.rb
Lockyy: Also, models don't need to create things. User.find doesn't create anything either.

.github/workflows/get_an_identity_data_sync_all.yml
Lockyy: Hm. They're tied tightly to the environment so we couldn't actually test them on a non-prod environment. This is a carbon copy of .github/workflows/update_all_schools.yml with the rake task swapped out which runs daily, so I would be confident this structure works.

I'm going to manually run the rake task itself in dev to test that there though.

spec/models/get_an_identity/user_spec.rb
Lockyy: whoops, planning notes 😅

spec/models/get_an_identity/user_spec.rb
Lockyy: It's stubbed yes, but it's making sure that if they do differ they get loaded appropriately, rather than one or the other being used for both. It'll allow you from the GetAnIdentity::User object you get back to see the ID you requested and the one you got back, which you could compare. Perhaps that's not the best approach though. It feels a little iffy to not have them be the same.

What do you think? Overwrite the ID with whatever you get back and leave the one requested to appear in the merged_user_ids array?

app/lib/services/ecf/ecf_user_updater.rb
shjohnson: Yeah i think i miss read this as was trying to workout when the update happens. this made it look like it updates each field at a time. so i think this is all good

app/lib/services/get_an_identity/user_updater.rb
shjohnson: Yeah or other option short term is create a small history table to store the data before its changed. to be honest if this data is already captured maybe we dont need to worry?

lib/tasks/users/get_an_identity_syncing.rake
shjohnson: ok no worries yeah i got confused and thought youd just use the other one sending an array with one user

.github/workflows/get_an_identity_data_sync_all.yml
shjohnson: Yeah sounds good, more an observation i guess

spec/models/get_an_identity/user_spec.rb
shjohnson: Ok i think ive worked out my gripe with this... 😂 In the code your asking for a user with a particular ID which is used as the params so i was confused why the response userID would be any different? 

In the test your not really testing this as an edge case as such because the response will just come back with whatever your setting it as? ie its not like the functionality would change etc i guess.

app/lib/services/ecf/ecf_user_updater.rb
Lockyy: Ah, it originally did them all together in an update() call but if nothing changed it'd hit the ECF API and cause an error since the JSON-API client would send an empty body. I had to move to this so I could check if an update was required before sending one.

app/models/get_an_identity/user.rb
Lockyy: What you're saying there is how I'm viewing them. They're objects representing data from another DB, so they're models in the same way as our regular models they just have a bit of distance between them and their data store. What I mean by other models is what's in lib/ecf_api, they're models they're just not in the model directory.

I could move them under lib, but it feels odd to have a generic "lib" directory containing code, the ecf_api classes feel better suited to move over into models to fit them into the core Rails MVC structure. It's not really particularly important, one or the other they're accessed and represented the same sort of thing.

spec/models/get_an_identity/user_spec.rb
Lockyy: It is testing things, just because the API response is stubbed doesn't mean it isn't testing the code. If I go in and break my actual code, this test would fail.

For example if I change the code inside GetAnIdentity::User from `@uid = response["userId"]` to `@uid = @id` the spec goes red. 

It's an explicit test that we are exposing data from the API and not the data the user initially passed in. I agree that it's essentially a meaningless test since we know it works, but it makes it explicit in the specs that we know that these two can differ and that is expected. But I guess in the end, it's a relatively meaningless spec.

app/models/get_an_identity/user.rb
shjohnson: I feel its a bit of a code smell to be honest, its obvious not our fault how the data model is setup and requiring to access this data in ECF. Although in a pure rails model youl would be able to create/delete/update all through the model class. I see this as a read only action in finding out the state of the data and although i actually dont agree it should sit in lib either.

i would rather we had something like `app/services/external/ecf/user_details.rb` or `..find_user.rb` / `retrieve_user.rb` Ideally we would remove the `app/lib` folder completely as i think all those folders could be in better places. By having services like this that have one task which is getting the latest details from ecf app keeps it very separate and unlikely cause future confusion. 

Also this is all for data-syncing between apps right? i do worry how much this could grow in future but so far we have limited models in NPQ so i doubt will become massive concern. Possibly as just us two happy to have someone else weigh in on the proposal?

app/models/get_an_identity/user.rb
shjohnson: I think going forward if this was expanded at all we would have a base class for all the configuration related stuff plus use a decorator/presenter/serialiser for the object maybe?

shjohnson: So you would have a very small command class like `user_details` that inherits from the base class with all configuration and then the command class which basically describes what the api call is doing would have serialised decorator for how the data should be presented?

Lockyy: Yeah that would make sense when things expand. I'm hopeful that with the webhook integration we won't need to do any more integration with APIs.

app/models/get_an_identity/user.rb
Lockyy: I can see that as an alternative, I disagree it's a code smell and would argue that pushing more into services isn't a great move since services are themselves non-standard Rails and turn into a catch-all "things that don't fit neatly into MVC." This does, to me, fit neatly into MVC as it fits pretty cleanly as a data-model.

@peteryates I think Sam has asked for a third opinion on this one.

peteryates: This is an interesting one and there are pros and cons to both approaches. I'm not against the idea of API-backed models but as @shjohnson says, it can lead to confusion when the model doesn't feel/act quite like an ActiveModel. 

The conventional way to do this in Rails would be via [ActiveResource](https://github.com/rails/activeresource), which does give you a lot of the Railsisms for 'free'. Perhaps that pattern could be applied here? That would make it more transparent and resemble the sibling classes better - providing the remote API will play nicely with it.

ActiveResource supports creating/updating/deleting too, and from what I can see this class just retrieves and makes available data - so perhaps a service class makes more sense.

Regardless, I'd probably go with either:

*  if it's only ever going to retrieve one user record at once change it to a service class `GetAnIdentityAPI::RetrieveUser` - so it's really obvious what it's doing
* leave it as a model but namespace it so it's clear it uses an API and not the local db

Lockyy: I definitely like the idea of ActiveResource, though frankly having used it in the past I'm wary of getting stuck in with it again as it ended up being very slow compared to alternatives. This may have changed though.

With this one running over quite a bit I'm not keen to refactor into a `GetAnIdentityAPI::RetrieveUser` style service, though I can see that approach working. It doesn't quite match the pattern we have with the EcfApi namespaced models though.

With this already being namedspaced under GetAnIdentity, which should make it clear it's representing GetAnIdentity models, do you think that matches what you mean by the second point? Or do you think a specific `External` namespace?

spec/models/get_an_identity/user_spec.rb
Lockyy: I've gotten rid of the extra spec :) Probably right that it's not really testing much

app/models/get_an_identity/user.rb
peteryates: `External` works for me, it makes it clear there's a distinction between it and the other models. @shjohnson?

Lockyy: I've pushed up a commit that puts them under `::GetAnIdentity::API`, so `::GetAnIdentity::API::AccessToken` and `::GetAnIdentity::API::User`, this feels the most descriptive and fits in with the naming pattern of `EcfApi::User` etc.

app/models/get_an_identity/external/user.rb
```diff
@@ -0,0 +1,56 @@
...
+                  :raw
+
+      def initialize(id:)
+        @id = id
+        self.load
+      end
+
+    private
+
+      def request_user
```
shjohnson, line 29: Im not sure if we will need many more of these services going forward but i think if that happened we should re factor some of this work to have a base class that would contain a lot of this type of code?

I think can let it slide for this PR but as you mentioned this would be more than just grabbing data in the future hence keeping as a model. 

so refactor things like the token fetching and maybe just having a structure to pass into the load class etc

app/lib/services/get_an_identity/user_updater.rb
```diff
@@ -12,9 +12,9 @@ def initialize(user:)
       end
 
       def call
-        get_an_identity_user = ::GetAnIdentity::User.find(user.get_an_identity_id)
+        get_an_identity_user = ::GetAnIdentity::External::User.find(user.get_an_identity_id)
```
shjohnson: Thinking about this now i think we should have external outside the GAI namespace? That way keeps it very clear and seperate like Peter suggested.

app/lib/services/get_an_identity/user_updater.rb
```diff
@@ -25,6 +25,7 @@ def call
           trn: get_an_identity_user.trn,
           uid: get_an_identity_user.uid,
           email: get_an_identity_user.email,
+          trn_lookup_status: get_an_identity_user.trn_lookup_status,
```
shjohnson, line 28: 👍

app/models/get_an_identity/user.rb
shjohnson: Sorry i never responded to this but discussed with @Lockyy on Thursday. I agree to make things super clear about it being separate and external works fine. Honestly i would be cautious about this pattern and if this is the only task that stays in that folder i feel a top level service folder would be better than being placed in `app/lib`.

Then following that we could move the other items in lib into better folder structures and get rid of `app/lib`?

app/lib/services/admin/users_search.rb
```diff
@@ -25,6 +25,6 @@ def call
 private
 
   def default_scope
-    User.left_joins(:applications)
+    User.left_joins(:applications).order(email: :asc)
```
Lockyy, line 28: This change is out of scope, but for some reason the intermittent failure on the admin specs had become persistent and I had to get this and the specs both ordering in the same way to guarantee proper testing. This'll also make it so the interface is consistent for admin users.

spec/features/disabled_get_an_identity_integration/happy_journeys/js/changing_do_you_work_in_a_school_from_no_to_yes_spec.rb
```diff
@@ -200,7 +200,7 @@
       expect(page).not_to have_text("The Early Headship Coaching Offer is a package of structured face-to-face support for new headteachers.")
     end
 
-    expect(retrieve_latest_application_user_data).to eq(
+    expect(retrieve_latest_application_user_data).to match(
```
Lockyy, line 203: Makes it easier to compare the good and bad on a failure.

shjohnson (approved):

app/lib/services/get_an_identity/user_updater.rb
Lockyy: Good point, swapped around

PR #732, opened 2023-04-11
[CPDNPQ-908] Implement Get an Identity webhook message API

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-908

Get an Identity is going to be handling user updates/merging going forwards, to communicate that down to us they are going to be providing a webhook that you can subscribe to. We need to implement an API so we can subscribe it as a webhook URL within Get an Identity and start receiving messages.

As a first step we need to begin accepting messages so that we can implement handlers for each message type further down the line.

- Implement POST /api/v1/get_an_identity/webhook_messages endpoint
- Add admin interface for viewing Webhook Messages and requeueing the process job

#### How the Webhook receiving process works, step by step

1. Get An Identity service sends message to NPQ API
2. Message signature is verified
3. Message is stored in the database
4. Background job is enqueued to process message
5. Get an Identity is informed of successful message receipt
6. Background Job processed separately.

The actual processing occurs separately so that failures at this stage do not impact informing GAI that we have successfully received the message


spec/controllers/api/v1/get_an_identity/webhook_messages_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+        before do
+          allow(ENV).to receive(:[]).with("GET_AN_IDENTITY_WEBHOOK_SECRET").and_return(stubbed_secret)
+        end
+
+        it "responds with a success response code" do
+          send_request
+          expect(response).to be_successful
+        end
+
+        it "creates a ::GetAnIdentity::WebhookMessage record with thesend_request body parsed into the record" do
```
shjohnson, line 31: just add a space in the description 🖊️

spec/controllers/api/v1/get_an_identity/webhook_messages_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+
+        context "when a message is received without the required fields" do
+          let(:raw_message) { "{}" }
+
+          it "returns a 200" do
+            send_request
+            expect(response).to have_http_status(:ok)
+          end
+
+          it "creates a ::GetAnIdentity::WebhookMessage record with thesend_request body parsed into the record" do
```
shjohnson, line 61: same as above

spec/controllers/api/v1/get_an_identity/webhook_messages_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+          end
+
+          it "does not enqueue a job to process the webhook" do
+            expect {
+              send_request
+            }.not_to have_enqueued_job(::GetAnIdentity::ProcessWebhookMessageJob)
+          end
+        end
+
+        context "when an already received message is repeated but the event type is different" do
```
shjohnson, line 123: Unless im understanding wrong but should this be:

`'when a duplicate message_id is received but event type is different'` Feels like using the phrase `already received message doesnt quite make sense.`

shjohnson (approved):

PR #746, opened 2023-04-20
[CPDNPQ-903] Use find_each for iterating over large numbers of records

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-903

There are too many records to load into memory all at once

Use find_each instead of each so that we only load a small number (1000) into memory at at time

### DIFF
--- a/lib/tasks/users/get_an_identity_syncing.rake
+++ b/lib/tasks/users/get_an_identity_syncing.rake
@@ -21,7 +21,7 @@ namespace :users do
                   .with_get_an_identity_id # Must have a get_an_identity_id
                   .where(get_an_identity_id_synced_to_ecf: false) # Must not have already been synced
 
-      users.each do |user|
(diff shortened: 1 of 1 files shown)


leoapost (approved):

lib/tasks/users/get_an_identity_syncing.rake
```diff
@@ -21,7 +21,7 @@ namespace :users do
                   .with_get_an_identity_id # Must have a get_an_identity_id
                   .where(get_an_identity_id_synced_to_ecf: false) # Must not have already been synced
 
-      users.each do |user|
+      users.find_each(batch_size: 200) do |user|
```
Lockyy, line 24: Reduced from the default 1000 to be certain we don't run into memory issues, this specific number isn't particularly meaningful we just wanted to be careful

PR #758, opened 2023-05-05
[CPDNPQ-1088] Refactor ChooseChildcareProvider form

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1088

This area of the app is complicated and difficult to understand. Refactoring and documenting the code is crucial to handover.

- Implement Forms::QuestionTypes::AutoCompleteInstitution that abstracts the logic behind the questions on Forms::ChooseChildcareProvider, Forms::ChooseSchool, and Forms::ChoosePrivateChildcareProvider.
- Implement Forms::QuestionTypes::AutoCompleteChildcareProvider that subclasses from Forms::QuestionTypes::AutoCompleteInstitution.
- Move Forms::ChooseChildcareProvider to use Forms::QuestionTypes::AutoCompleteChildcareProvider format.


app/lib/forms/question_types/base.rb
```diff
@@ -1,14 +1,18 @@
...
-      attr_reader :name, :options
+      attr_reader :name, :options, :locale_keys, :question_data
 
-      def initialize(name:, options: [], style_options: {})
+      def initialize(name:, options: [], style_options: {}, locale_keys: {}, **question_data)
         @name = name
         @options = options
         @style_options = style_options # Freeform optional parameters that can differ for each subclass
+        @locale_keys = locale_keys
+        @question_data = OpenStruct.new(question_data)
```
Lockyy, line 11: Allows delegate_missing_to to work properly

app/lib/forms/question_types/base.rb
```diff
@@ -1,12 +1,13 @@
 module Forms
   module QuestionTypes
     class Base
-      attr_reader :name, :options
+      attr_reader :name, :options, :locale_name, :question_data
```
BroiSatse, line 4: `question_data` is not assignable via initializer. Is that on purpose?

app/models/school.rb
```diff
@@ -22,6 +22,10 @@ class School < ApplicationRecord
 
   scope :open, -> { where(establishment_status_code: %w[1 3 4]) }
 
+  def display_name
+    name
+  end
```
BroiSatse, line 27: Are these for active admin?

config/locales/en.yml
```diff
@@ -360,3 +363,14 @@ en:
...
+        choose_childcare_provider_other: "Workplace not shown above"
+        choose_childcare_provider_search: "Please choose from workplaces located in %{institution_location}"
+
+        choose_school: "What’s the name of your workplace?"
+        choose_school_other: "School or college not shown above"
+        choose_school_search: "Please choose from schools or 16 to 19 educational settings located in %{institution_location}."
+
+        choose_private_childcare_provider: "Enter your or your employer’s URN"
+        choose_private_childcare_provider_other: "My employer is not shown above"
+        choose_private_childcare_provider_search: "Please choose your employer"
```
BroiSatse, line 376: OOi: Why the quotes? Is this the convention we're using here? :)

app/views/registration_wizard/shared/_question_page.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :title do %>
   <%= form.errors.present? ? "Error: " : nil %>
-  <%= t("helpers.#{form.question.title_locale_type}.registration_wizard.#{form.question.name}") %>
+  <%= t("helpers.#{form.question.title_locale_type}.registration_wizard.#{form.question.name_locale_key}") %>
```
BroiSatse, line 3: Locales within questions are an interesting problem. I definitely agree those should follow form names rather than field name, but would be great to just make them available via `t('.key')`. I'd probably move them out of `helpers` into their own `questions` scope?

BroiSatse (approved):

app/models/school.rb
Lockyy: PrivateChildcareProviders sometimes have a blank name and so end up displaying "institution_identifier" as the radio text. To fix this I added display_name as a method on those for use in the no_js scenario. Since School and LocalAuthority also get rendered out in that view they needed to respond to the same interface.

app/lib/forms/question_types/base.rb
Lockyy: That is a mistake and should've been removed when I reverted a possible avenue for implementing something. Removing this now.

app/views/registration_wizard/shared/_question_page.html.erb
Lockyy: I have a second PR coming up to change FindChildcareProvider that moves all these question/hint locale strings into a method on the question. That way we can do this sort of refactor to move the strings somewhere else.

EmmaFrith (approved):

PR #761, opened 2023-05-05
[CPDNPQ-1088] Update childcare-provider-not-in-engalnd to use copy_page template

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1088

We are standardising on one page format across the app.

- Forms::ChildcareProviderNotInEngland moved to use copy_page template

### DIFF
--- a/app/models/registration_wizard.rb
+++ b/app/models/registration_wizard.rb
@@ -307,6 +307,7 @@ private
       your_role
       your_employer
       school_not_in_england
+      childcare_provider_not_in_england
       possible_funding
       ineligible_for_funding
       funding_your_npq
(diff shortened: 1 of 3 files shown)


app/models/registration_wizard.rb
```diff
@@ -307,6 +307,7 @@ def steps
       your_role
       your_employer
       school_not_in_england
+      childcare_provider_not_in_england
```
Lockyy, line 310: This page is pretty much inaccessible in actuality, since in_england? is hard coded to true for ChildcareProviders. However, better to have the page actually working rather than not.

shjohnson (approved):

PR #762, opened 2023-05-09
[CPDNPQ-1047] Add env examples to .env.template

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1047

Codebase changes for improved handover.

Adds examples to .env.template required for app functionality.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,6 +1,16 @@
+RAISE_ON_MISSING_TRANSLATIONS=false
+
 GOVUK_NOTIFY_API_KEY=
 
 ECF_APP_BASE_URL="https://ecf-dev.london.cloudapps.digital"
 ECF_APP_BASE_URL="http://localhost:3001"
+ECF_APP_BEARER_TOKEN=
 
-ECF_APP_BEARER_TOKEN=""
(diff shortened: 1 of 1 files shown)


.env.template
```diff
@@ -1,6 +1,16 @@
+RAISE_ON_MISSING_TRANSLATIONS=false
```
BroiSatse, line 1: I'd vote for this to be true by default - we hit a number of issue in TV when missing translation were causing havoc in production and were missed completely in test and development.

BroiSatse (approved):

.env.template
Lockyy: I'd also vote for that, however I didn't want to turn it on by default as that would be a behavioural change. Which is why it falls back to false in the actual config file.

However, considering this is just the example .env file and it'd only turn it on for people newly setting up the project, I think now would be a good time to make that change.

@BroiSatse @shjohnson What do you think about:
a) Setting this to true in .env.example, so people setting up the project from scratch have it on by default? This is backwards compatible with existing setups.
and b) Setting this to true in development.rb so you have to turn it off rather than turning it on? This is not backwards compatible and would change this for anyone with an existing setup.

PR #764, opened 2023-05-09
Move share_provider to question_page system

BroiSatse:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1104

Just another part of the question_page migration

- Replaced the share_provider view with a common question partial.
- Created new type of question - check box

### Notes:

I was playing with an idea of CheckBoxGroup as well, however it seems there are some significant conceptual differences between a group and a single checkbox and in the end decided to write CheckBox as a standalone type of question rather than a very specific type of CheckBoxGroup.

### DIFF
--- a/app/views/registration_wizard/shared/questions/_check_box.html.erb
+++ b/app/views/registration_wizard/shared/questions/_check_box.html.erb
@@ -1,0 +1,6 @@
+<%= form.govuk_check_boxes_fieldset(question.name, multiple: false, **question.style_options) do %>
+  <% question.body.each do |body_text| %>
+    <p class="govuk-body"><%= body_text %></p>
+  <% end %>
+  <%= form.govuk_check_box question.name, question.checked_value, question.unchecked_value, multiple: false, required: question.required %>
+<% end  %>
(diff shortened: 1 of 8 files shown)


app/views/registration_wizard/shared/questions/_check_box.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= form.govuk_check_boxes_fieldset(question.name, multiple: false, **question.style_options) do %>
+  <% question.body.each do |body_text| %>
+    <p class="govuk-body"><%= body_text %></p>
+  <% end %>
```
Lockyy, line 4: Oh I like this, would be good to implement in other question types too.

BroiSatse: It is a bit limited though - we have very little control over where that text is displayed, how it is formatted and only handle simple text - no bullets nor images... I was thinking about passing a block which would allow modification of given question layout, but then we're sort-of entering view-component zone and I ma not sure if that's what we're going for here. Another option would be to write a helper responsible for rendering a question (which could take a block and pass it to the partial)

Lockyy: That's fair, it's at least a start of how to get something more flexible in place.

Lockyy (approved):

PR #766, opened 2023-05-10
[CPDNPQ-1080] Use localised course names instead of course.name where possible

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1080

`course.name` contains old versions of NPQ names, we should be moving towards only using the localised strings everywhere.

These names are used in most locations around the app, this cleans up remaining uses of the old names.

### Changes proposed in this pull request

- Use new course names on /registration/ineligible-for-funding
- Use new course names on /registration/choose-your-provider
- Use new course names on /registration/possible-funding
- Use new course names on /registration/confirmation
- Use new course names in application submitted emails 
  Note that though we are sending the new course name this isn't actually used in the template.
- Use new course names on /admin/users
- Use new course names on /admin/unsynced_applications

| Old | New|
| - | - | 
| NPQ for Leading Teaching (NPQLT) | Leading teaching |
| NPQ for Leading Behaviour and Culture (NPQLBC) | Leading behaviour and culture |
| NPQ for Leading Teacher Development (NPQLTD) | Leading teacher development |
| NPQ for Senior Leadership (NPQSL) | Senior leadership |
| NPQ for Headship (NPQH) | Headship |
| NPQ for Executive Leadership (NPQEL) | Executive leadership |
| Additional Support Offer for new headteachers | Additional Support Offer |
| Early Headship Coaching Offer | Early headship coaching offer |
| NPQ for Early Years Leadership (NPQEYL) | Early years leadership |
| NPQ for Leading Literacy (NPQLL) | Leading literacy |

| Before | After| 
| - | - |
| choose-your-provider | choose-your-provider |
| confirmation | confirmation |
| ineligible-for-funding | ineligible-for-funding |
| possible-funding | possible-funding |

PR #769, opened 2023-05-12
Updated the content of the confimration page

BroiSatse:
### Context

[Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-000
](https://dfedigital.atlassian.net/browse/CPDNPQ-992)

Simple update of some content on confirmation page as per new designs


javier-npq (approved):

config/locales/en.yml
```diff
@@ -2,6 +2,19 @@ en:
   time:
     formats:
       admin: "%R on %d/%m/%Y"
+  courses:
+    alternative_names:
```
Lockyy, line 6: Would it be possible to change the key to identifier? To match this https://github.com/DFE-Digital/npq-registration/pull/766/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R236? 

It'd work into the consistent naming system that Emma and I are introducing through the app.

BroiSatse: I actually wondered whether this is something that should be part of the database model? It feels a bit odd to have a part of naming in the database and the rest in translations...

Lockyy: This is actually something we're specifically moving away from, the names being in the DB was causing several issues. Moving to just holding the identifier in the DB and then names in the locale file has made things easier. I've taken what you've said on board in that PR and swapped course.identifier with course.name.

Lockyy: The issues were down to the DB values causing inconsistencies across this and the ECF app and needing to refresh the DB to make changes. Having the identifier in the DB and the locales handling how that is converted into what the user sees makes deploying and managing things easier.

Lockyy (approved):

PR #771, opened 2023-05-15
Migrated choose-your-provider page to the question-page style

BroiSatse:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1108

Migration to question-page


app/lib/forms/question_types/radio_button_group.rb
```diff
@@ -1,14 +1,19 @@
 module Forms
   module QuestionTypes
     class RadioButtonGroup < Base
+      DEFAULT_STYLES = {
+        legend: {
+          size: "xl",
+          tag: "h1",
+        }.freeze
+      }.freeze
```
Lockyy, line 9: Can these be dynamic? It gets into needing a wrapper around DEFAULT_STYLES otherwise to add label strings as a default. See here: https://github.com/DFE-Digital/npq-registration/pull/773/commits/eb71ac8e7b19e2c7c09888bde8fa0bd049a25e93#diff-c2e5b256743b09aebd787c9aca59a0b1877d0a854d38e32c27e8b55f0dc80e44R28

peteryates: These [can be set in the form builder config too](https://github.com/DFE-Digital/govuk-formbuilder/blob/main/lib/govuk_design_system_formbuilder.rb#L91)

shjohnson (approved):

PR #775, opened 2023-05-16
Question page as a view component

BroiSatse:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1109

Apart from the refactor onto a question page, this PR also converts QuestionType objects into a standalone ViewComponents. Transofrmation is seamlessly invisible and QuestionType objects can still be used both via a question-page partial or be rendered explicitly as `render form.question`.

View components adds a bit of more power to previously plain object as we can now define various slots in our question pages as we won't be able to replace all the pages with a plain objects without defining a completely new DSL to mimic html - for example, in this PR we need to add an additional link below the text box.

Ideally, we would come up with some universal configuration option for all the question pages. Naturally some of the pages will have their own additional attributes, which is why I also introduced `ActiveModel::Attributes` to QuestionType - this way we can avoid the long list of keyword arguments and overriding initialise function.

### DIFF
--- a/app/lib/forms/confirm_email.rb
+++ b/app/lib/forms/confirm_email.rb
@@ -11,6 +11,18 @@ module Forms
       ]
     end
 
+    def question
+      @question ||= QuestionTypes::TextField.new(
+        name: :confirmation_code,
+        form: self,
+        header: "Confirm your email address",
+        style_options: {
+          width: 10,
+          label: { text: "Enter your code" },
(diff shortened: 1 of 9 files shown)


BroiSatse:
I am closing this PR as the email-confirmation page it is affecting has been now permanently removed. The changes to base view will be hopefully merged in another PR

PR #776, opened 2023-05-17
[CPDNPQ-1092] Move Get an Identity login to beginning of NPQ flow

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1092

To support the account management interface we need to move the Get an Identity login flow (and backup TRN question) to the beginning of the flow. 

- Moves GAI to beginning of flow
- Remove the concept of the “pilot” for GAI, meaning all users will be tied to a GAI account regardless of TRN status
- Move TRN fallback flow to be a single question (qualified_teacher_check) after the GAI flow.
- Remove email questions (no longer needed since we can rely on emails from GAI being present).


app/controllers/omniauth_controller.rb
```diff
@@ -1,6 +1,5 @@
 class OmniauthController < Devise::OmniauthCallbacksController
   skip_before_action :verify_authenticity_token, only: [:tra_openid_connect]
-  before_action :remove_from_get_an_identity_pilot, only: :tra_openid_connect, if: :tra_response_excludes_trn?
```
Lockyy, line 3: The concept of the pilot is entirely removed in this PR, users are always tied to the GAI account they logged in via. The fallback for no TRN being present is to now introduce the qualified_teacher_check after the oauth login as the first question, then move right into the rest of the flow.

app/helpers/forms/flow_helper.rb
```diff
@@ -0,0 +1,9 @@
+module Forms::FlowHelper
+  def after_login_next_step
+    if Services::Feature.trn_required? && query_store.current_user.trn.blank?
```
Lockyy, line 3: To allow TRN-free users now all we'll need to do is toggle Services::Feature.trn_required? so that it returns false and users without TRNs will continue on with the flow as any other user would, eventually reaching lead providers. For now this is hardcoded to true and all users without TRNs go to the qualified_teacher_check question.

app/controllers/omniauth_controller.rb
```diff
@@ -67,12 +65,8 @@ def failure_message
     "There was an error. Please try again in a few moments. If this problem persists, contact us at continuing-professional-development@digital.education.gov.uk"
   end
 
-  def registration_wizard_omniauth_step
-    :start
-  end
-
   def failed_sign_in_path
-    registration_wizard_show_path(:get_an_identity)
+    registration_wizard_show_path(:start)
```
Lockyy, line 69: With the basic oauth kick off button being on the start page we no longer need the get_an_identity interstitial page as a non-JS fallback, the start page functions as that now.

app/lib/forms/qualified_teacher_check.rb
```diff
@@ -120,12 +113,33 @@ def active_alert?
...
+        trn: trn_to_store,
+        full_name:,
+        date_of_birth:,
+        national_insurance_number: ni_number_to_store,
+        trn_verified: trn_verified?,
+        trn_lookup_status:,
+        trn_auto_verified: !!trn_auto_verified?,
+        active_alert: active_alert?,
+      )
+      wizard.store["trn_set_via_fallback_verification_question"] = true
```
Lockyy, line 142: This controls whether to show the name/email/nino/trn fields on the check_answers page, since we need to display those when the user has to fallback to this page.

Lockyy: It's also useful to just have the flag stored so we can identify these users.

app/lib/forms/qualified_teacher_check.rb
```diff
@@ -120,12 +113,33 @@ def active_alert?
...
+      query_store.current_user.update!(
+        trn: trn_to_store,
+        full_name:,
+        date_of_birth:,
+        national_insurance_number: ni_number_to_store,
+        trn_verified: trn_verified?,
+        trn_lookup_status:,
+        trn_auto_verified: !!trn_auto_verified?,
+        active_alert: active_alert?,
+      )
```
Lockyy, line 141: Setting the info here immediately rather than at the end of the flow in handle_submission_for_store means that we can use a single place to find TRNs now, `current_user.trn`.

app/lib/services/feature.rb
```diff
@@ -2,14 +2,10 @@ module Services
   class Feature
     REGISTRATION_OPEN_DATE = Time.zone.parse("6 June 2022 12:00")
 
-    GAI_INTEGRATION_KEY = "Get an Identity integration".freeze
```
Lockyy, line 5: This is already always at 100%, this change is here since the code now always has it turned on and users can't be "removed from the pilot" anymore since the non-pilot journey has been removed and replaced with the signle qualified_teacher_check question.

app/lib/services/handle_submission_for_store.rb
```diff
@@ -10,8 +10,6 @@ def initialize(store:)
 
     def call
       ActiveRecord::Base.transaction do
-        user.update!(user_update_data)
```
Lockyy, line 13: This happens in the after_save on qualified_teacher_check now.

spec/support/helpers/journey_helper.rb
```diff
@@ -16,6 +16,14 @@ def retrieve_latest_application_data
       latest_application&.as_json(except: %i[id created_at updated_at user_id])
     end
 
+    def deep_compare_application_data(expected_data)
```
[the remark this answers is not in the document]
shjohnson: Alt Text
Alt Text
Alt Text
Alt Text

spec/support/shared_contexts/stub_previously_funding_check_for_all_courses.rb
```diff
@@ -1,13 +1,12 @@
...
-  let(:api_call_get_an_identity_id) { raise NotImplementedError }
   let(:api_call_trn) { raise NotImplementedError }
 
   before do
     Course.pluck(:identifier).each do |course_identifier|
       mock_previous_funding_api_request(
         course_identifier:,
         trn: api_call_trn,
-        get_an_identity_id: api_call_get_an_identity_id,
+        get_an_identity_id: user_uid,
```
shjohnson, line 9: Is user_uid always going to be set where this is used?

shjohnson:
Have we had a few runs through the UI and reieiw app for this? already i know there will be a few clashes with my PR but just got to get one of them in asap

shjohnson:
Seems good though and great change to get out!

app/lib/services/handle_submission_for_store.rb
Lockyy: All users get data set via the GAI login, then any users that would've been updated by this go through qualified_teacher check. Anything that could be set on the user has already been set previously.

app/lib/forms/qualified_teacher_check.rb
```diff
@@ -136,6 +155,18 @@ def ni_number_required?
 
     attr_reader :verified_trn
 
+    def trn_to_store
+      if trn_verified?
```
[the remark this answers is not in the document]
Lockyy: As I recall some Lead providers need it this way

app/lib/forms/qualified_teacher_check.rb
```diff
@@ -70,14 +75,7 @@ def next_step
...
 
-        if wizard.query_store.inside_catchment?
-          return :find_school if wizard.query_store.works_in_school?
-          return :kind_of_nursery if wizard.query_store.works_in_childcare?
-
-          return :your_employment
-        end
-
-        :choose_your_npq
+        :provider_check
```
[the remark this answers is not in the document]
Lockyy: 100%

The saves etc. all feel like they should be after_save/before_save steps, not in the next step. I didn't do any major refactoring because this question is going to be entirely deleted and I wasn't sure if it was worth getting into the weeds on it based on that.

app/lib/services/feature.rb
```diff
@@ -19,17 +15,8 @@ def initialize_feature_flags
...
-      end
-
-      def enroll_user_in_get_an_identity_pilot(user)
-        Flipper.enable_actor(Services::Feature::GAI_INTEGRATION_KEY, user)
-      end
-
-      def remove_user_from_get_an_identity_pilot(user)
-        Flipper.enable_actor(Services::Feature::REMOVE_USER_FROM_GAI_PILOT_KEY, user)
+      def trn_required?
+        true
```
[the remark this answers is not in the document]
Lockyy: Good call

app/lib/forms/qualified_teacher_check.rb
```diff
@@ -136,6 +155,18 @@ def ni_number_required?
...
+    def trn_to_store
+      if trn_verified?
+        verified_trn.rjust(7, "0")
+      else
+        trn.rjust(7, "0")
+      end
+    end
+
+    def ni_number_to_store
+      national_insurance_number unless trn_verified?
```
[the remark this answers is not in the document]
Lockyy: It is odd. I'm unsure why that's the case but that is how handle_submission_for_store did this so I have replicated the functionality here.

spec/support/helpers/journey_helper.rb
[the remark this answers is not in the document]
Lockyy: Yeah as soon as I realised I could split it up I was so pleased to get it in place and make this quicker.

spec/support/shared_contexts/stub_previously_funding_check_for_all_courses.rb
Lockyy: Yeah, it's set as a part of "Stub Get An Identity Omniauth Responses" which is required for all flows now since all users hit the get an identity flow.

shjohnson (approved):

PR #777, opened 2023-05-19
Cpdnpq 1112 find school migration

BroiSatse:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1112

This tickets builds on top of the #775 and is blocked until it is merged.


app/lib/forms/question_types/base.rb
```diff
@@ -32,9 +50,9 @@ def question_text
         I18n.t("helpers.#{title_locale_type}.registration_wizard.#{name_locale_key}")
       end
 
-    private
-
-      attr_reader :style_options
+      def body=(value)
+        super Array.wrap(value)
+      end
```
peteryates, line 55: Not entirely sure reaching into ViewComponent is a good idea here.

peteryates: If I replace the method above with this the whole test suite passes:

```ruby
def body
  []
end
```

BroiSatse: @peteryates - Body is not a part of a ViewComponent - it's just one of the attributes defined on the base question_type. It defined a list of paragraphs to be displayed between the h1 and the question - they are not part of the tests, so forcing them to empty array should not cause any test failures (unless they explicitly test for content).

app/lib/forms/question_types/check_box.rb
```diff
@@ -1,30 +1,22 @@
...
-
-      def initialize(*args, checked_value: "1", unchecked_value: "0", required: false, **opts)
-        @checked_value = checked_value
-        @unchecked_value = unchecked_value
-        @required = required
-
-        super(*args, **opts)
-      end
+      attribute :checked_value, default: "1".freeze
+      attribute :unchecked_value, default: "0".freeze
```
javier-npq, line 5: Just my 2 cents but maybe it would be better to use constants or symbols here?

Something like


```
CHECKED_VALUE = "1".freeze
  UNCHECKED_VALUE = "0".freeze

  attribute :checked_value, default: CHECKED_VALUE
  attribute :unchecked_value, default: UNCHECKED_VALUE
```

Or

`attribute :checked_value, default: :checked`
`attribute :unchecked_value, default: :unchecked`

javier-npq (changes requested):

app/lib/forms/question_types/check_box.rb
BroiSatse: OOI - why would it be better? Those are only the default value of dynamic state. The values are set to match default values of gov.uk component that is being used. I don't really see any benefit of using constants here as those values should not be used anywhere without the instances. Changing default values to :checked and :unchecked will break rails params casting.

javier-npq (approved):
Looks good to me, the comment made was addressed via chat.

PR #784, opened 2023-05-23
[CPDNPQ-1092] Add teacher reference number question back in

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1092

Faisal highlighted that the flow isn't fully clear without this and Jackie and Diana agreed. 

This PR adds the TRN question back in after the GAI flow and before the qualified_teacher_check question.

### DIFF
--- a/app/lib/forms/teacher_reference_number.rb
+++ b/app/lib/forms/teacher_reference_number.rb
@@ -1,0 +1,51 @@
+module Forms
+  class TeacherReferenceNumber < Base
+    VALID_TRN_KNOWLEDGE_OPTIONS = %w[yes no-dont-have].freeze
+
+    attr_accessor :trn_knowledge
+
+    validates :trn_knowledge, presence: true, inclusion: { in: VALID_TRN_KNOWLEDGE_OPTIONS }
+
+    def self.permitted_params
+      %i[
+        trn_knowledge
+      ]
+    end
+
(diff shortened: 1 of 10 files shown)


app/lib/forms/teacher_reference_number.rb
```diff
@@ -0,0 +1,51 @@
...
+      Forms::QuestionTypes::RadioButtonGroup.new(
+        name: :trn_knowledge,
+        options:,
+        style_options: { fieldset: { legend: { size: "m", tag: "h1" } } },
+      )
+    end
+
+    def options
+      [
+        build_option_struct(value: :yes, link_errors: true),
```
shjohnson, line 46: Why is one a string and the other a symbol?

shjohnson (approved):

app/lib/forms/teacher_reference_number.rb
Lockyy: Oh yeah 🙂 These are actually both symbols. I had to use quotes because the other has hyphens in it and the interpreter doesn’t see those as part of the symbol name unless wrapped to escape . Wrapping in quotes, aka. :”foo-bar”, allows you to use hyphens in symbol names. 

I’d prefer to use just :no but this matches the value stored in the DB prior to this question being removed.

PR #787, opened 2023-05-24
[Do not merge] [CPDNPQ-1078] [SPIKE] Implement geocoded local authority search

Lockyy:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1078

This is not to be merged into NPQ, this code will be used in another repository that will be set up later. That repo will use this code for searching local authorities.

This spike uses activerecord-postgis-adapter, geocoder, and rgeo to translate free text user location search into a list of the nearest local authorities.

The GeoJSON file available from [the GOV Open Geography Portal](https://geoportal.statistics.gov.uk/datasets/ons::local-authority-districts-december-2022-boundaries-uk-bfc/explore) is loaded into the DB by `Services::GeojsonLoader` using the PostGIS adapter provided by `activerecord-postgis-adapter`. 

Geocoder is used to convert entered user searches into lon/lat coordinates.

Rgeo is then used to transform those lon/lat coordinates into [EPSG:27700 British National Grid projection](https://epsg.io/27700-1197) coordinates. These coordinates once transformed line up to the projection stored in the database which is also the 27700 projection. 

The transformed coordinates are then used in a distance query using <-> to find the nearest local authority geometries.

Note that the database adaptor had to be swapped over from postgres to postgis to enable this functionality.


app/lib/services/geojson_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+    def self.reload_geojsons(geojson_path)
+      empty_database
+      load_geojsons(geojson_path)
+    end
+
+    def self.empty_database
+      GeoLocalAuthority.delete_all
+    end
+
+    def self.read_file(geojson_path)
```
slawosz, line 12: Would move this method below `load_geojsons` method as its helper.

app/lib/services/geojson_loader.rb
```diff
@@ -0,0 +1,45 @@
+module Services
+  class GeojsonLoader
+    def self.reload_geojsons(geojson_path)
+      empty_database
+      load_geojsons(geojson_path)
+    end
+
+    def self.empty_database
+      GeoLocalAuthority.delete_all
```
slawosz, line 9: I would consider inlining instead of separate method.

app/lib/services/geojson_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+      file = File.read(geojson_path)
+      features = RGeo::GeoJSON.decode(file, json_parser: :json)
+    end
+
+    def self.load_geojsons(geojson_path)
+      Rails.logger.debug("Loading #{geojson_path}")
+
+      features = read_file(geojson_path)
+
+      features.each do |feature|
```
slawosz, line 22: Would it make sense to wrap it into transaction?

app/models/geo_local_authority.rb
```diff
@@ -0,0 +1,31 @@
+class GeoLocalAuthority < ApplicationRecord
+  include RGeo::ActiveRecord::GeometryMixin
+
+  SRID = 27_700
+
+  def self.nearest_three_to(string)
```
slawosz, line 6: Maybe `self.nearest_to(string, amount: 3)` instead hardcoding argument in method name?

app/models/geo_local_authority.rb
```diff
@@ -0,0 +1,31 @@
...
+    location = Geocoder.search(string).first
+    by_distance(location.longitude.to_f, location.latitude.to_f).limit(3)
+  end
+
+  def self.by_distance(longitude, latitude)
```
slawosz, line 11: How many results does it returns by default? Could it turn out to be expensive as do extensive search?

BroiSatse: It doesn't really return any results - it returns AR scope. The scope itslef uses some of the postgis magic which is rathere quite performant. I wouldn't expect any delays based on the size of the database in this service

PR #789, opened 2023-05-25
Create view page for user account information

shjohnson:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1086

This change will allow users to view their past applications in the account page and see what information they select per registration.

- Upgrade `govuk-components`, `govuk-frontend` and `view_component` to the latest version as utilising summary cards where syntax has changed.
- Create new view page for accounts to show all registrations
- Improve seed data

To do- 
Sort out status values
Change the otp to be a static value rather than in the seeds
Improve seeds for the staging env
Fix flow from GAI
Double check test coverage


app/helpers/application_helper.rb
```diff
@@ -8,6 +8,25 @@ def boolean_red_green_tag(bool, text = nil)
     content_tag(:strong, text, class: "govuk-tag govuk-tag--#{colour}")
   end
 
+  def boolean_submitted_passed_tag(status, text = nil)
+    case status.downcase
```
slawosz, line 12: Cosmetic:
```
color = case status.downcase
when "foo"
  "bar"
when "baz"
  "foobar"
end
```
Or
```
{ 
  "foo" => "bar",
  ...
}[status]
````

File: app/lib/forms/sign_in.rb [hunk 2]
```diff
@@ -30,5 +32,9 @@ def after_save
         ConfirmEmailMailer.confirmation_code_mail(to: email, code:).deliver_now
       end
     end
+
+    def skip_sending_otp
```
slawosz, line 36: Cosmetic: `skip_sending_otp?` or maybe `omit_otp?`. Besides, maybe instead of doing it on per step basis, it could be left as it is and on development otp could accept static code like `12345`. But its of course out of scope for and leave it as topic for discussion - especially I am not familiar with this functionality yet.

shjohnson: 100% this needs some love and could be added to the backlog for changes!

shjohnson: to be honest in dev or even testing i dont see why its used; static number actually better your right though! might include

app/helpers/application_helper.rb
peteryates: `#fetch`? 😅

peteryates: Also as we're already using govuk-components it might make sense to use `govuk_tag` here, it supports [all the colours](https://govuk-components.netlify.app/components/tag/#coloured-tags).

app/views/accounts/_provider_details.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= govuk_summary_card(title: "Provider details") do |card|
+
+card.with_summary_list(rows: [
+    {
+      key: { text: "Provider" },
```
shjohnson, line 5: @dalcJ @EmmaFrith move into course details

app/views/accounts/show.html.erb
```diff
@@ -1,19 +1,28 @@
...
+  <% courses = current_user.applications.map { |app| app.course }.uniq %>
+
+  <% courses.each do |course| %>
+    <%= render "application_table", course: course %>
+  <% end %>
+<% end %>
+
+<h1 class="govuk-heading-l">Register for an NPQ</h1>
+
+<p class="govuk-body">User this service to submit a new registration</p>
```
EmmaFrith, line 25:
```suggestion
<p class="govuk-body">Use this service to submit a new registration.</p>
```

app/views/accounts/_work_details.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    {
+      key: { text: "Do you work in England?" },
+      value: { text: boolean_tag(application.teacher_catchment?) }
+    },
+    {
+      key: { text: "Work setting" },
+      value: { text: I18n.t(application.work_setting, scope: "helpers.label.registration_wizard.work_setting_options") },
+    },
+    {
+      key: { text: "Employer" },
```
EmmaFrith, line 13:
```suggestion
      key: { text: "Employer name or unique reference number" },
```

slawosz (approved):

PR #800, opened 2023-06-02
Add Flow Visualiser to README.md and cleanup invalid questions

Lockyy:
### Context

The flow as it actually is in the app can be hard to follow without a way to see an overview of the whole flow. Adding a generated diagram to the readme that updates off of the code itself would allow us to easily view and share the app's real flow.

- Adds rake task for generating PNG diagram of questionnaire flow
- Removes questions that are no longer in use
- Sets next_step on end state questions
- Removes hardcoded funding eligibility symbols from next_steps and moves to using constants

The changes outside of the visualiser are changes prompted by the visualiser displaying unused questions or from the flow not making sense because the next_step method on questions was not set up correctly.

I've made sure the commits are split into "visualiser" changes and "code" changes.


.github/workflows/build_flow_visualisation.yml
```diff
@@ -0,0 +1,52 @@
...
+          path: markdown
+
+      - name: Push updated visualisation to repo Wiki
+        run: |
+          cp tmp/visualisations/registration_wizard_visualisation.png markdown/visualisations/registration_wizard_visualisation.png
+          cd markdown
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add visualisations/registration_wizard_visualisation.png
+          git diff-index --quiet HEAD || git commit -m "Update flow visualisation" && git push
```
peteryates, line 52: Great idea to use the Wiki's repo for this 🧠

peteryates (approved):
I love it, really easy to read and understand what's going on.

.github/workflows/build_flow_visualisation.yml
Lockyy: I was trying to use github artefact saving for a while and it just wasn't working neatly. Once this occurred to me it fit together really smoothly. 🤩

PR #801, opened 2023-06-05
Update Have Ofsted URN to use QuestionType system

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1113

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -67,11 +67,35 @@ jobs:
       - name: Install node.js dependencies
         run: yarn install
 
+      - name: Create screenshots directory
+        run: mkdir -p tmp/capybara
+
       - name: Set up test database
         run: bin/rails parallel:setup
 
       - name: Run tests
         run: bundle exec rake parallel:spec
+
+      - name: Checkout wiki code
+        uses: actions/checkout@v3
+        if: failure()
+        with:
+          repository: ${{github.repository}}.wiki
+          path: markdown
+
+      - name: Push screenshots to repo Wiki
+        if: failure()
+        run: |
+          rm -rf "markdown/capybara-screenshots/$GITHUB_REF_NAME"
+          mkdir -p "markdown/capybara-screenshots/$GITHUB_REF_NAME"
+          cp tmp/capybara/*.png "markdown/capybara-screenshots/$GITHUB_REF_NAME/"
+          cd markdown
+          ruby generate-screenshots-list.rb
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add "capybara-screenshots/$GITHUB_REF_NAME/*"
+          git diff-index --quiet HEAD || git commit -m "Update screenshots for pr ${{ github.event.number }}" && git push
+
   linting:
     name: Run Rubocop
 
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -45,6 +45,7 @@ gem "net-smtp", require: false
 group :development, :test do
   gem "byebug", platforms: %i[mri mingw x64_mingw]
   gem "capybara"
+  gem "capybara-screenshot"
```
slawosz, line 48: Its useful tool that makes screenshot (both html and image) of the page that failed so it is easily visible what is wrong.

app/views/registration_wizard/have_ofsted_urn.html.erb
```diff
@@ -1,44 +1,16 @@
-<% content_for :title do %>
-  <%= @form.errors.present? ? "Error: " : nil %>Do you or your employer have an Ofsted unique reference number (URN)?
+<% content_for :additional_info do %>
```
slawosz, line 1: I am happy to rename `:additional_info` for more meaningful name.

Gemfile
Lockyy: Very nice, do these appear as artefacts in the Github Action?

app/views/registration_wizard/have_ofsted_urn.html.erb
Lockyy: @BroiSatse I recall you implementing something similar to this?

Gemfile
Lockyy: If not, we should get that in along side this so you can see the screenshots in failed Github Action test runs. We can add this to the test.yml after the spec runs to store these artefacts so that you could view them:

```
- name: Archive test failure screenshots
  uses: actions/upload-artifact@v3
  with:
    name: test-failure-screenshots
    path: path/to/screenshots/**/*
```

Then it'd be available to download like in this test run: https://github.com/DFE-Digital/npq-registration/actions/runs/5155930311

I found out how to do this here: https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

app/views/registration_wizard/have_ofsted_urn.html.erb
BroiSatse: Yeah I did. There are two similiar concepts in already - one is `body` attribute (which is unable to handle this case) and the other one is a slot on Question as a ViewComponent. I think I'm ok with this bein done as is but I'd like to o over all the steps at one point and clean it all up.

BroiSatse (approved):

.github/workflows/specs.yml
```diff
@@ -67,11 +67,44 @@ jobs:
...
+
+      - name: Remove screenshots from wiki repo
+        run: |
+          rm -rf "markdown/capybara-screenshots/$GITHUB_REF_NAME"
+          cd markdown
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git rm "capybara-screenshots/$GITHUB_REF_NAME"
+          git diff-index --quiet HEAD || git commit -m "Remove screenshots for pr ${{ github.event.number }}" && git push
+
```
BroiSatse, line 107: Out of interest - why are we pushing screenshots to the wiki repo instead of saving them as build artifacts? (Sorry if this was already discussed!)

PR #802, opened 2023-06-05
Parallelise specs/linting/brakeman on github

Lockyy:
### Context

Should be a small speed improvement

Runs rspec, rubocop, and brakeman in parallel rather than sequentially, allowing them to fail/pass separately


shjohnson (approved):
LGTM, defo good thing to have because few times we fail on something where we could actually also solve something else rather than having to push again.

PR #818, opened 2023-06-21
ITT picker aliases

BroiSatse:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1159


app/lib/services/autocomplete_itt_providers.rb
```diff
@@ -5,7 +5,18 @@ def self.as_open_structs
     end
   end
 
+  def self.as_autocomplete_options
+    all_names.map do |legal, operating|
```
slawosz, line 9: The `all_names` method name is little misleading, I would simply use query here:
```
IttProvider.pluck(:legal_name, :operating_name).map do ...
```

app/webpacker/packs/itt-provider-picker.js
```diff
@@ -4,8 +4,31 @@ import openregisterLocationPicker from 'govuk-country-and-territory-autocomplete
...
-    url: ''
+    url: '',
+    source (query, callback) {
+      const distinct = {}
+
+      const filtered = optionsMap.filter(option => option.search.includes(query.toLowerCase()))
+      filtered.forEach(option => {
+        distinct[option.value] = distinct[option.value] || { path: option.text !== option.value && option.text , name: option.value }
+      })
+      console.log(Object.values(distinct))
```
slawosz, line 30: We probably don't want `console.log` in production?

slawosz (approved):
LGTM, minor comments.

PR #819, opened 2023-06-22
CPDNPQ-1188 Background Job to Update Status Fields

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1188

- have a rake job created which will periodically update the status field (lead_rovider_approval_status and participant declaration outcomes) in NPQ app to match any updates made in the ECF app.


app/lib/services/ecf/ecf_application_synchronization_service.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    private
+
+      def build_uri
+        URI.parse("#{ENV['ECF_APP_BASE_URL']}/api/v1/npq/application_synchronizations#index")
+      end
+
+      def fetch_required_application_ids
+        Application.where(lead_provider_approval_status: nil).limit(1000).pluck(:ecf_id)
+      end
```
javier-npq, line 23: Sending parameterized ids based on empty `lead_provider_approval_statuses` for testing purposes.

config/schedule.rb
```diff
@@ -0,0 +1,5 @@
+require File.expand_path(File.join(File.dirname(__FILE__), "..", "config", "environment"))
+
+every 1.day, at: "00:00" do
```
slawosz, line 3: Have you tested if its working - by running it on review app every 10 minutes or so? Ideally changing manually `ecf` DB?

javier-npq: Yea, used the above mentioned approach and the logs stated that the job ran Ok. I just merged the refactored ECF code changes and I'm going to run a sanity on this again just to be sure.

app/lib/services/ecf/ecf_application_synchronization_service.rb
```diff
@@ -0,0 +1,51 @@
...
+        Net::HTTP.start(uri.hostname, uri.port, use_ssl: use_ssl?, read_timeout: 30) do |http|
+          http.request(request)
+        end
+      end
+
+      def handle_response(response)
+        if response.is_a?(Net::HTTPSuccess)
+          data = response_data(response)
+          data["data"].each do |record|
+            SyncApplicationStatusJob.perform_now(record)
```
BroiSatse, line 35: A bit of a nitpick, but I wonder if we could some additional data processing here. At the moment, this Job is performing http call, but it is also SyncApplicationStatus job responsibility to understand the structure of the response.

Could we create a generic Struct object like `Application::SynchronizationData` with clearly defined fields, parse the response onto collection fo those structs and pass it instead?

config/schedule.rb
```diff
@@ -0,0 +1,5 @@
+require File.expand_path(File.join(File.dirname(__FILE__), "..", "config", "environment"))
+
+every 10.minutes do
+  runner "ApplicationSynchronizationJob.perform_now", environment: Rails.env
```
BroiSatse, line 4: Could we do `perform_later` here instead? I think it would be better for the job to be performed by the worker rather than ctron process (I am not sure how the logs would work for cron worker)

slawosz: That makes sense!

BroiSatse:
There's one potential issue with scalability here. If we enable cron on our docker, all the deployed instances will run cron in the background causing multiple execution of the same job. Also, running things in the background is a bit against of cloud architecture. In ideal world, we should have a single dedicated container responsible for running cron - this would also resolve issues with invisible cron logs.

app/lib/services/ecf/application_updater.rb
```diff
@@ -0,0 +1,18 @@
...
+      def initialize(data)
+        @data = data
+      end
+
+      def call
+        id = @data["attributes"]["id"]
+        lead_provider_approval_status = @data["attributes"]["lead_provider_approval_status"]
+        participant_outcome_state = @data["attributes"]["participant_outcome_state"]
+
+        application = Application.find_by(ecf_id: id)
```
slawosz, line 13: Here, I would rather prefer to have `ActiveRecord::RecordNotFound` exception as it should never happen. If happens, there is something wrong with the data and we will investigate.

app/lib/services/ecf/ecf_application_synchronization_service.rb
slawosz: And I would take another approach here - I personally like to have simple to follow code. Inside the loop you could have such code [inlined](https://refactoring.com/catalog/inlineFunction.html) to prevent extra layers of abstraction, as is really calling a very simple function with no logic. Also, at current stage tracking single jobs is more complicated.
```
application = record["attributes"]
lead_provider_approval_status = application["lead_provider_approval_status"]
participant_outcome_state = application["participant_outcome_state"]
application = Application.find_by!(ecf_id: application["id"])
application.update!(lead_provider_approval_status:, participant_outcome_state:)
```

config/schedule.rb
javier-npq: Yea, will definitely be changing it back to `perform_later` . Once everything's tested and sorted

.github/workflows/update_application_statuses.yml
```diff
@@ -0,0 +1,82 @@
+name: Schedule Update Application Statuses
```
javier-npq, line 1: @slawosz @BroiSatse, I moved out the cron job into a separate container.

javier-npq: Not quite sure if I need to do anything else to add this workflow to Github actions and make it a part of the deployment process?

app/lib/services/ecf/ecf_application_synchronization.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      def handle_response(response)
+        if response.is_a?(Net::HTTPSuccess)
+          data = response_data(response)
+          data.each do |record|
+            id = record.id
+            lead_provider_approval_status = record.lead_provider_approval_status
+            participant_outcome_state = record.participant_outcome_state
+
+            application = Application.find_by(ecf_id: id)
```
slawosz, line 39: How about using `find_by!` as it throws proper excetion by default?

slawosz (approved):

slawosz:
Approved. It would be worth it to verify manually few days after deployment, if the applications are properly being updated.

PR #837, opened 2023-07-12
CPDNPQ-1188: Update only synced application with ecf

javier-npq:
### Context
update only synced applications with ecf

### Ticket/PR: 
https://github.com/DFE-Digital/npq-registration/pull/819

We are not raising an exception, we will be updating any ID's that are present in the npq space.

```
For example, application A and B got updated in the ECF. But in the NPQ, we have only application A and application B has not synced yet.

It should update only application A since it is synced with Ecf.
```

### DIFF
--- a/app/lib/services/ecf/ecf_application_synchronization.rb
+++ b/app/lib/services/ecf/ecf_application_synchronization.rb
@@ -36,8 +36,12 @@ module Services
             lead_provider_approval_status = record.lead_provider_approval_status
             participant_outcome_state = record.participant_outcome_state
 
-            application = Application.find_by!(ecf_id: id)
-            application.update!(lead_provider_approval_status:, participant_outcome_state:)
+            application = Application.find_by(ecf_id: id)
+            if application.present?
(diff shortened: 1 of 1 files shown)


app/lib/services/ecf/ecf_application_synchronization.rb
```diff
@@ -36,8 +36,12 @@ def handle_response(response)
             lead_provider_approval_status = record.lead_provider_approval_status
             participant_outcome_state = record.participant_outcome_state
 
-            application = Application.find_by!(ecf_id: id)
-            application.update!(lead_provider_approval_status:, participant_outcome_state:)
+            application = Application.find_by(ecf_id: id)
+            if application.present?
+              application.update!(lead_provider_approval_status:, participant_outcome_state:)
+            else
+              Rails.logger.info("Application #{id} is not synced yet")
```
slawosz, line 43: Maybe ` Rails.logger.info("Application where ecf_id=#{id} is not synced yet")`

slawosz (approved):

PR #857, opened 2023-08-17
Remove confirmation page: redirect to registration

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1230

Remove the confirmation page and redirects the user directly to the user registration.
The existing implementation did not support the automation of the landing page by
defining a new step, so I have decided to [add a new method to the defines if the step is 
the last one](https://github.com/DFE-Digital/npq-registration/blob/0093b392f5ec0e9883de212750a76b45bf743ff0/app/lib/forms/check_answers.rb#L11-L13).

```ruby
def last_step?
  true
end
```

Using this method, the controller will be able to check if the step is last and take 
the proper actions.

You may notice many tests changing, this has happened for two reasons:

1. The assertions for the last step were inlined in all the tests, so I had to change all
of them. To prevent this from happening again [I have extracted a common step](https://github.com/DFE-Digital/npq-registration/blob/c9d9855ec8e79c44abc1ec9e394159ecca8e35d8/spec/support/helpers/journey_assertion_helper.rb#L34-L42).
2. I have included the two attributes below in the `check-answers` step because
they were already set on the `confirmation` step.

```ruby
  wizard.store["submitted"] = true
  wizard.session["clear_tra_login"] = true
```

image

image

### DIFF
--- a/app/views/accounts/user_registrations/show.html.erb
+++ b/app/views/accounts/user_registrations/show.html.erb
@@ -12,7 +12,7 @@
 <% content_for :before_content do %>
   <% if current_user.applications.count > 1 %>
     <%= render GovukComponent::BackLinkComponent.new(
-      text: "Back",
+      text: "Back to your registrations",
       href: account_path
     ) %>
   <% end %>


javier-npq (approved):
Looks good to me, checked on review app as well. The flow looks good.

dalcJ:
@pmanrubia 

The back button on the registration page should only show if the user has >1 registrations. Currently it is showing for any number of registrations. 

The for the back button should be 'Back to your registrations', like this: 

Screenshot 2023-08-24 at 12 46 02

dalcJ:
@pmanrubia 

The block at the bottom for Register for an NPQ, can we add extra 30px of space between the heading and table. 


Screenshot 2023-08-24 at 12 38 43

dalcJ:
@pmanrubia I also noticed on the 'choose NPQ' page that Maths is showing up - should it be, as it's not part of this PR? 

Screenshot 2023-08-24 at 14 10 48

javier-npq:
Yea it might be appearing in your pr because the db is shared and we've created the course.. the flow will only be available once the maths npq stuff gets merged

dalcJ:
Ah, thanks for clarifying about the maths option. That makes sense

dalcJ:
@pmanrubia thanks. Just one thing left is the text on the back button. It should be 'Back to your registrations' 

Screenshot 2023-08-25 at 13 13 48

dalcJ:
@pmanrubia thanks - all looks good :)

dalcJ (approved):

PR #873, opened 2023-09-05
Log users with duplicated Get-an-identity-id

pmanrubia:
Related to: #869 and #862 

https://dfedigital.atlassian.net/browse/CPDNPQ-1269

### Context

Some users have two accounts created in NPQ, one of them has an UID 
assigned and the other account it doesn't.

This PR, logs them by checking if the user if there is an existing user with 
the same UID that we are about to assign.

These cases must be checked individually and an action should be taken.

### DIFF
--- a/lib/tasks/one_off/populate_get_identity.rake
+++ b/lib/tasks/one_off/populate_get_identity.rake
@@ -38,10 +38,10 @@ namespace :get_identity_id do
 
         get_identity_id = attrs.fetch(:user_id)
         if user && user.uid.present? && user.uid != get_identity_id
-          Rails.logger.error("User #{user.id} #{user.email} with existing GIA? different -> (#{user.uid != get_identity_id})")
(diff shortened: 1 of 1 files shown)


lib/tasks/one_off/populate_get_identity.rake
```diff
@@ -35,7 +35,14 @@ namespace :get_identity_id do
         user = application.user
 
         Rails.logger.error("User not found") if user.nil?
-        Rails.logger.error("User #{user.id} with existing GIA? different -> (#{user.uid != attrs.fetch(:user_id)})") if user && user.uid.present?
+
+        get_identity_id = attrs.fetch(:user_id)
+        if user && user.uid.present? && user.uid != get_identity_id
+          Rails.logger.error("User #{user.id} #{user.email} with existing GIA? different -> (#{user.uid != get_identity_id})")
```
slawosz, line 41: We should not log user emails. User id id enought.

slawosz (approved):

PR #874, opened 2023-09-05
Enabled feature flag for sandbox environment 

javier-npq:
### Context

Enabled feature flag for sandbox environment

### Ticket

https://dfedigital.atlassian.net/browse/CPDNPQ-1257

`SERVICE_ENV` is an environment variable that used to obtain the correct environment. Enable the feature flag as a result.

### DIFF
--- a/app/views/accounts/_application_table.html.erb
+++ b/app/views/accounts/_application_table.html.erb
@@ -25,18 +25,18 @@
           Provider application
         </dt>
         <dd class="govuk-summary-list__value">
-            <% if application.lead_provider_approval_status.blank? || application.lead_provider_approval_status&.capitalize == "Pending" %>
+            <% if application.lead_provider_approval_status.blank? || pending?(application) %>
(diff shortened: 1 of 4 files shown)


app/views/accounts/_application_table.html.erb
```diff
@@ -36,7 +36,7 @@
         </dd>
       </div>
 
-      <% if application.participant_outcome_state.present? && application.lead_provider_approval_status.capitalize == "Accepted" %>
+      <% if application.participant_outcome_state.present? && application.lead_provider_approval_status&.capitalize == "Accepted" %>
```
slawosz, line 39: Would be nice to have a nicer way of checking this, like
```
class Application
  def accepted?
    lead_provider_approval_status&.capitalize == "Accepted"
  end
end
```

slawosz (approved):

app/views/accounts/_application_table.html.erb
javier-npq: Yeah, nice catch. Updated the code likewise. 👍

PR #893, opened 2023-09-19
Introduce the review environment

peteryates:
### Context

When we move the NPQ reg application to Azure we'll be moving to the model used in ECF where we have a Rails environment per actual (deployed) environment, plus a `development` one for working locally.

I also cleaned up the database config by removing the unnecessary comments.

- Add the review env config
- Remove the noise from the database config
- Add review database config

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,24 +1,6 @@
-# PostgreSQL. Versions 9.1 and up are supported.
-#
-# Install the pg driver:
-#   gem install pg
-# On OS X with Homebrew:
-#   gem install pg -- --with-pg-config=/usr/local/bin/pg_config
-# On OS X with MacPorts:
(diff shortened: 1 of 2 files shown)


config/database.yml
```diff
@@ -1,84 +1,20 @@
...
-  # Schema search path. The server defaults to $user,public
-  #schema_search_path: myapp,sharedapp,public
-
-  # Minimum log levels, in increasing order:
-  #   debug5, debug4, debug3, debug2, debug1,
-  #   log, notice, warning, error, fatal, and panic
-  # Defaults to warning.
-  #min_messages: notice
+review:
+  <<: *default
```
peteryates, line 12: no `database:` needed here because in review we'll always be supplying a `DATABASE_URL`

javier-npq (approved):

PR #901, opened 2023-09-22
Remove .js dependency from Registrations with GAI

pmanrubia:
### Context

We have JS code that is not currently in use and that is introducing some
complexity in our codebase. In the [RegistrationWizardController][1], we have the option
of serving JS requests using UJS. 

This code was needed when we were trying to OAuth the user when the flow was:

```
Start Page -> question -> question -> GAI service
```

To make it compatible with OAuth, the previous code was sending a remote call
with a response to click the kickoff button. As of now, the sequence of steps
has changed, so we can safely remove the JS dependencies.

### What

This PR removes JS dependency alongside the Stimulus controller that goes with it.

### Notes

I have also tried to remove the dependency `rails-ujs` from `package.json`, but
it is used in the autocomplete select box for the school.

[1]: https://github.com/DFE-Digital/npq-registration/blob/730a170962105a1280e9bb02ca12850bb33f324f/app/controllers/registration_wizard_controller.rb#L1


peteryates (approved):
👍🏽

PR #902, opened 2023-09-25
Rails 7 upgrade

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1341

Following best practices, we are upgrading to Rails 7. I have also removed the
versioning from the `Gemfile` so dependabot can support the team with 
frequent upgrades.


db/migrate/20230925093246_add_service_name_to_active_storage_blobs.active_storage.rb
```diff
@@ -0,0 +1,22 @@
...
+      change_column :active_storage_blobs, :service_name, :string, null: false
+    end
+  end
+
+  def down
+    return unless table_exists?(:active_storage_blobs)
+
+    remove_column :active_storage_blobs, :service_name
+  end
+end
```
peteryates, line 22: Can we drop ActiveStorage if we're not using it?

Gemfile
```diff
@@ -23,16 +23,16 @@ gem "json_api_client", ">= 1.21.1"
...
+gem "omniauth-oauth2"
 gem "omniauth-rails_csrf_protection"
 gem "pagy"
 gem "paper_trail", "~> 15.0"
 gem "pg", ">= 0.18", "< 2.0"
 gem "pg_search"
 gem "puma", "~> 6.4.0"
 gem "rack-attack"
-gem "rails", "~> 6.1.7"
+gem "rails"
```
slawosz, line 35: Wouldn't be better to specify minor version here?

pmanrubia: I am curious about this one. 

If we are using `dependabot` for gems upgrades, and the update is descriptive and small enough of the version that is being upgraded, Isn't it better to receive all the updates for Rails all upgrades once they are available? 

I would like to see an update from version `7.1.2` to `7.2` as soon as it happens because at the very end, we will have to do the update as frequently as possible to decrease risk.

Wouldn't it be better to just say: `gem "rails", "7"` ?

I won't fight much for it :), just curious about your thoughts. I noticed that ECF also fix the minor version

cc/ @peteryates

peteryates: I don't have any strong reservations or thoughts on version. so long as we're keeping on top of it and current or nearly-current 😅

I'd probably be inclined to go with `7` or `7.1` and let dependabot take care of the rest.

pmanrubia: I have finally gone with 

```ruby
gem "rails", "~> 7.0" 
```

This would allow updates from 7.0.x to 7.1.x, 7.2.x, etc., but not to 8.0.0

javier-npq (approved):

PR #908, opened 2023-09-29
Configure AKS plans and memory

peteryates:
### Context

Set up the number of nodes for the web and worker apps and set the PostgreSQL database size and high availability to true.

- Fix indentation
- Set the webapp and worker replicas
- Set postgres server sku and high availability

### DIFF
--- a/terraform/application/application.tf
+++ b/terraform/application/application.tf
@@ -43,6 +43,9 @@ module "web_application" {
 
   docker_image = var.docker_image
   command = var.command
+
+  replicas   = var.webapp_replicas
+  max_memory = var.webapp_memory_max
 }
 
 module "worker_application" {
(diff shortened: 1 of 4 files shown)


terraform/application/application.tf
```diff
@@ -43,6 +43,9 @@ module "web_application" {
 
   docker_image = var.docker_image
   command = var.command
+
+  replicas   = var.webapp_replicas
+  max_memory = var.webapp_memory_max
```
RMcVelia, line 48: the new vars need to be added to variables.tf with appropriate defaults

terraform/application/variables.tf
```diff
@@ -59,6 +59,26 @@ variable "enable_monitoring" {
...
+variable "worker_memory_max" {
+  type    = string
+  default = "1Gi"
+}
+
+variable "worker_replicas" {
+  type = number
+  default = 1
+}
+
```
RMcVelia, line 81: also postgres_enable_high_availability and postgres_flexible_server_sku need to be declared

peteryates: ah of course, thanks 🤦🏽

RMcVelia (approved):

slawosz (approved):

PR #909, opened 2023-10-02
Added github token in actions

javier-npq:
### Context
Investigate permissions issue for dependabot PRs and resolve it.
### Ticket
https://dfedigital.atlassian.net/browse/CPDNPQ-1335
[What has been changed?]

Added new action with github_token

> - name: Push changes
>   uses: ad-m/github-push-action@master
>   with:
>     github_token: ${{ secrets.GITHUB_TOKEN }}

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -96,14 +96,15 @@ jobs:
           git diff-index --quiet HEAD || git commit -m "Update screenshots for pr ${{ github.event.number }}" && git push
 
       - name: Remove screenshots from wiki repo
+        env:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
         run: |
           cd markdown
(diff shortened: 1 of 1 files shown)


.github/workflows/specs.yml
```diff
@@ -102,7 +102,8 @@ jobs:
           git config --local user.name "GitHub Action"
           git rm -r --ignore-unmatch "capybara-screenshots/$GITHUB_REF_NAME"
           git diff-index --quiet HEAD || git commit -m "Remove screenshots for pr ${{ github.event.number }}" && git push
-
+        env:
```
slawosz, line 105: I would move `env` before `run` block.

slawosz (approved):

.github/workflows/specs.yml
javier-npq: Good call

PR #918, opened 2023-10-05
CPDNPQ-206 Improve search capability for Ofsted reference numbers

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-206

Include substring search. Now user can search urn with only numbers as well. 

failures:


slawosz (approved):

app/controllers/private_childcare_providers_controller.rb
```diff
@@ -1,7 +1,7 @@
 class PrivateChildcareProvidersController < ApplicationController
   def index
     @private_childcare_providers = PrivateChildcareProvider
-      .search_by_urn(params[:name])
+      .where("provider_urn ~* ?", params[:name])
```
pmanrubia, line 4: @javier-npq shouldn't the `pg_search_scope` cover what we are looking here? It would be nice to document why this is better, and also add some tests. 

In theory, pg_search scopes allow to search for sub-strings.

PR #943, opened 2023-10-17
Make sandbox and prod use staging's docker image

peteryates:
The 'needs' directive can only access the things directly needed by the current step, so must refer to deploy_staging instead of docker.


pmanrubia (approved):
👍

PR #946, opened 2023-10-19
Allow to disable registration in Flipper

slawosz:
For the Azure migration, we would like to disable registrations and inform user.

To disable the registration, either switch the flag in console or web admin. To switch flag in console, simply use `Services::Feature.disable_registration!` and to enable it back `Services::Feature.enable_registration!`


pmanrubia (approved):

peteryates (approved):
Looks good so long as we're happy with the content cc @EmmaFrith

PR #962, opened 2023-10-24
Use azure konduit app for uploading metrics

slawosz:


peteryates (approved):
Looks fine, but longer term these should move to delayed cron job.

Remember to drop the `Test powershell` commit before merging

PR #965, opened 2023-10-26
Remove flows that depend on CloudFoundry

peteryates:
We've moved away from GOV.UK PaaS and to Azure. These jobs will no longer work as they depend on the cf client.

Instead we should probably add buttons to trigger them in the admin interface, or schedule them via DelayedJob.


pmanrubia:
@peteryates I deleted these files as part of this PR: https://github.com/DFE-Digital/npq-registration/pull/963

We can probably close this PR. 
Thanks for your work on this.

PR #975, opened 2023-11-02
Revert "Bump google-github-actions/setup-gcloud from 0 to 1"

slawosz:
This breaks `Export Dashboard to BigQuery` job.
Reverts DFE-Digital/npq-registration#972


peteryates (approved):
If it works, do it. I still think these should run within the app though

PR #989, opened 2023-11-08
Disable email uniquenes checks for Users

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1470

There are several instances where users can not log in because email validation failed. As now we are relaying on GAI for authentication, we decided to use only GAI `uid` as unique identifier for user.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,8 +4,6 @@ class User < ApplicationRecord
   has_many :applications, dependent: :destroy
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
 
-  validates :email, uniqueness: true
-
   scope :admins, -> { where(admin: true) }
   scope :unsynced, -> { where(ecf_id: nil) }
   scope :synced_to_ecf, -> { where.not(ecf_id: nil) }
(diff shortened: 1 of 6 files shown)


javier-npq (approved):

app/models/user.rb
```diff
@@ -28,7 +28,7 @@ def self.find_or_create_from_provider_data(provider_data, feature_flag_id:)
   end
 
   def self.find_or_create_from_tra_data_on_uid(provider_data, feature_flag_id:)
-    user_from_provider_data = find_or_initialize_by(provider: provider_data.provider, uid: provider_data.uid)
+    user_from_provider_data = find_or_initialize_by(email: provider_data.info.email)
```
pmanrubia, line 31: @slawosz if we don't find the user by the `uid`, if the user changes the email on GAI, then we won't be able to find it.

pmanrubia (changes requested):

app/models/user.rb
pmanrubia: I wasn't probably very clear :)

When a user is authenticated in GAI, on NPQ we should only care about the UID from that user in GAI and update all the other information. This information should then be also passed to ECF (if changes)

pmanrubia (approved):

PR #1001, opened 2023-11-16
Update language about funding

pmanrubia:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1361

### Context

As DfE we want to use the same language for funding (currently in some places we use ‘pay for’ and in others ‘funding’) in order to reduce confusion for users, particularly around different types of eligibility (course and funding).


app/models/registration_wizard.rb
```diff
@@ -206,11 +206,11 @@ def answers
     end
     unless eligible_for_funding?
       if course.ehco?
-        array << OpenStruct.new(key: "How is your EHCO being paid for?",
+        array << OpenStruct.new(key: "How are you funding your course?",
```
pmanrubia, line 209: @dalcJ I have added this file as well, although the copy should change, I think.

dalcJ:
@pmanrubia @EmmaFrith I've tested all the funding outcome pages, they are all good. 
So just the changes to the 'course payment page' mentioned in Slack.

dalcJ:
@pmanrubia on the page /registration/funding-your-npq, which shows when the person isn't funded and for all courses except EHCO. 
The heading and page title should be ‘How are you funding your course?’

Screenshot 2023-12-07 at 09 50 17

dalcJ:
@pmanrubia all looks good now. Thank you :)

slawosz (approved):
Looks good from code point of view.

PR #1002, opened 2023-11-16
Import schools for review apps while replanting data

pmanrubia:
### Context

When are review app is deployed the replant data is executed. This PR limits the number of schools being loaded as the process is costly in time.

I have limited it to 1500 schools. The total process takes around 30 seconds.


slawosz (approved):

daniyaljamshed (approved):

db/seeds.rb
```diff
@@ -128,3 +128,5 @@ def seed_itt_providers!
     )
   end
 end
+
+ImportGiasSchools.new.call(max_schools: 1500) unless Rails.env.test?
```
pmanrubia, line 132:
```suggestion
ImportGiasSchools.new.call(max_schools: 200) unless Rails.env.test?
```

pmanrubia: Reduce the number of schools (speed up review deployment)

PR #1013, opened 2023-11-22
Added feature flag for targeted support funding

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1378

Remove targeted support funding.

Put targeted support funding under a feature flag :targeted_support_funding which is disabled by default.

failures:

### DIFF
--- a/app/services/feature.rb
+++ b/app/services/feature.rb
@@ -14,6 +14,7 @@ class Feature
         Flipper.add(feature_flag_key)
       end
       Flipper.enable(:maths_npq)
+      Flipper.disable(:targeted_support_funding)
     end
 
     # This is always true but is checked so that it is explicit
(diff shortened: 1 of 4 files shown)


app/services/feature.rb
```diff
@@ -14,6 +14,7 @@ def initialize_feature_flags
         Flipper.add(feature_flag_key)
       end
       Flipper.enable(:maths_npq)
+      Flipper.disable(:targeted_support_funding)
```
slawosz, line 17: I think it will disable the feature flag every time the code reloads.

javier-npq: Yes, that is intentional as far as i am concerned. Might be worth it to test this and check with jake.

slawosz (approved):

PR #1036, opened 2023-12-08
Add models to support ECF migration

pmanrubia:
### Context

As part of the separation work between NPQ and ECF, we need to create in NPQ the
domain models that will support all the new processes.

This PR introduces:

- Declaration
- Statement
- StatementItem
- Cohort
- Outcome
- Contract

### Related to
- https://github.com/DFE-Digital/npq-separation/blob/main/data-modelling.md
- Related ticket: https://dfedigital.atlassian.net/browse/CPDLP-2659

### DIFF
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -20,7 +20,7 @@ class Statement < ApplicationRecord
 
   validates :output_fee, presence: true
   validates :month, numericality: { in: 1..12, only_integer: true }
-  validates :year, numericality: { greater_than_or_equal_to: 2020, only_integer: true }
+  validates :year, numericality: { in: 2020..2050, only_integer: true }
 
   validate :validate_max_statement_items_count
 


peteryates:
These models represent the final form @leandroalemao, we'll be using them for some sample (seeded) data which we can work out how to validate.

db/migrate/20231208114021_create_statement_line_items.rb
```diff
@@ -0,0 +1,11 @@
+class CreateStatementLineItems < ActiveRecord::Migration[7.0]
```
peteryates, line 1: `StatementLineItem` should really be `StatementLine` or `StatementItem` here.

app/models/schedule.rb
```diff
@@ -0,0 +1,4 @@
+class Schedule < ApplicationRecord
+  belongs_to :course
```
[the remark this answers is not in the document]
peteryates: No, sorry - I think that might be a mistake in the diagram - should just be `Cohort`. Unless we decided in a purely NPQ context `Course` might help too here.

app/models/statement_item.rb
```diff
@@ -0,0 +1,11 @@
+class StatementItem < ApplicationRecord
+  STATES = %w[eligible payable paid voided ineligible awaiting_clawback clawed_back].freeze
+
+  belongs_to :statement
+  belongs_to :declaration
+
+  validates :statement_id, presence: true
+  validates :declaration_id, presence: true
+
+  validates :state, inclusion: { in: STATES }
```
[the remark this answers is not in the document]
peteryates: Yeah, I prefer [PostgreSQL enums](https://www.postgresql.org/docs/current/datatype-enum.html) and steering clear of too many Rails-generated magic methods. In ECF we have stuff like

```ruby
  enum training_status: {
    active: "active",
    deferred: "deferred",
    withdrawn: "withdrawn",
  }, _prefix: "training_status"
```

which results in us having `training_status_active`, and it's totally ungreppable. I'd rather just validate the value and control the states via a basic state machine.

peteryates:
I don't think we need separate migration models, in my head the migration is a ruby script that ingests the data we dump out of ECF. Unless we take the route of connecting NPQ to two databases and pulling it in with SQL 🤔

app/models/statement_item.rb
```diff
@@ -1,3 +1,9 @@
+# StatementItem
+#
+# * Is the join table of a many-to-many relation between declarations and
+#   statements
+# * A declaration could have two statement line items max, one for the original
+#   payment and a potential second one if it’s clawed back
```
peteryates, line 6: We could perhaps represent the idea of a clawback better via a reference than a status. Having a self-referential foreign key (`clawed_back_from_id`) would give us more information than a raw `state`, because:

* it tells us which item we're clawing back
* it easily allows us to find all clawbacks (or items that have been clawed back, depending on the direction of the key)

app/models/schedule.rb
```diff
@@ -1,3 +1,16 @@
+# Schedule
+#
+# * has many milestones, varying by type of npq course (e.g. a specialist course
+#   has 3 milestones)
```
peteryates, line 4: Clarify this now we've dropped milestones

app/models/statement.rb
```diff
@@ -1,3 +1,17 @@
...
+# * A schedule has many statement
+#   -  A cohort/contract can have many statement
+# * A statement can have many declaration
+# * Not all statements have declarations
+# * A statement is comprised of:
+#   - service fees
+#   - output fee (individual declarations/outputs, uplifts)
+#   - vat?
+# * Generated by us (DfE)
+# * A statement can be ‘frozen’ preventing providers from voiding declarations
```
peteryates, line 12: Let's use `frozen_at` rather than a boolean flag.

app/models/contract.rb
```diff
@@ -0,0 +1,23 @@
+# A contract is a ...
+# recruitment_target:
+# output_payment_percentage:
+class Contract < ApplicationRecord
+  belongs_to :statement
+  belongs_to :course
+  belongs_to :cohort
+  belongs_to :lead_provider
+
+  validates :statement_id, presence: true
```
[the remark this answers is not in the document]
peteryates: Good call, thanks. So long as we're checking the validation's present in the specs

app/models/declaration.rb
```diff
@@ -0,0 +1,37 @@
+# Declaration
+#
+# * Made by lead providers via the API
+# * There can be multiple per application
+# * Must persist even if application state is reverted to pending
```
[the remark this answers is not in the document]
cwrw: worth discussing in person as this one I'm on the fence, as we can allow it for certain rules but declarations can't be deleted basically

app/models/outcome.rb
```diff
@@ -0,0 +1,17 @@
+# Logging the information provided as part of the completed
+# declaration
+#
+# If they pass we pass the info over to the DQT via the API
+#
+# Outcomes can be updates if the situation changes
+class Outcome < ApplicationRecord
+  belongs_to :declaration
+
+  # "completion_date" normally the same date as the final
```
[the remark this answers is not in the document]
cwrw: no needs to always be there, see https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/participant_outcome/npq.rb#L18 for future date validation as well

app/models/application.rb
```diff
@@ -14,6 +14,7 @@ class Application < ApplicationRecord
   belongs_to :private_childcare_provider, optional: true
   belongs_to :itt_provider, optional: true
 
+  has_many :declarations, dependent: :destroy
```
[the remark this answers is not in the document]
cwrw: good shout 👍

app/models/outcome.rb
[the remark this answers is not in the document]
cwrw: yea sounds good :)

app/models/contract.rb
```diff
@@ -0,0 +1,18 @@
...
+  belongs_to :lead_provider
+
+  validates :per_participant, numericality: { greater_than: 0 }
+  validates :recruitment_target, numericality: { only_integer: true, greater_than: 0 }
+  validates :output_payment_percentage, numericality: { only_integer: true, greater_than: 0, less_than: 100 }
+
+  # these three fields should be null after [a certain date]
+  validates :number_of_payment_periods, numericality: { only_integer: true, greater_than: 0 }
+  validates :service_fee_percentage, numericality: { only_integer: true, greater_than: 0 }
+  validates :service_fee_installments, numericality: { only_integer: true, greater_than: 0 }
```
mooktakim, line 17: Maybe some/all of these should allow zero.

app/models/statement.rb
```diff
@@ -0,0 +1,32 @@
...
+#   and logic.
+class Statement < ApplicationRecord
+  belongs_to :cohort
+  belongs_to :lead_provider
+  has_many :statement_items
+
+  validates :cohort_id, presence: true
+  validates :lead_provider_id, presence: true
+
+  validates :month, numericality: { in: 1..12, only_integer: true }
```
mooktakim, line 23: I would prefer to go with strings for `month` instead of integers, ("january" or "jan" etc).

app/models/statement.rb
```diff
@@ -0,0 +1,32 @@
...
+# * A statement can be ‘frozen’ preventing providers from voiding declarations
+#   and initializing clawbacks for any future voids. Refer to mark_as_paid field
+#   and logic.
+class Statement < ApplicationRecord
+  belongs_to :cohort
+  belongs_to :lead_provider
+  has_many :statement_items
+
+  validates :cohort_id, presence: true
+  validates :lead_provider_id, presence: true
```
mooktakim, line 21: `belongs_to` should already require presence, I believe this is redundant.

app/models/statement_item.rb
```diff
@@ -0,0 +1,19 @@
...
+class StatementItem < ApplicationRecord
+  STATES = %w[eligible payable paid voided ineligible awaiting_clawback clawed_back].freeze
```
mooktakim, line 8: Maybe we should use `Declaration::STATES`, so we have one place.

db/migrate/20231208113439_create_statements.rb
```diff
@@ -0,0 +1,15 @@
+class CreateStatements < ActiveRecord::Migration[7.0]
+  def change
+    create_table :statements do |t|
+      t.integer :month
+      t.integer :year
+      t.date :deadline_date
+      t.references :cohort, null: false, foreign_key: true
+      t.references :lead_provider, null: false, foreign_key: true
+      t.datetime :marked_as_paid_at
+      t.decimal :reconcile_amount
```
mooktakim, line 10: We should add precision for money: `t.decimal :reconcile_amount, precision: 8, scale: 2`

app/models/statement.rb
mooktakim: @pmanrubia 
We currently have a `name` field with text "January 2024". We wanted to split this up into `month` and `year` to better validate it.
Migration shouldn't be too hard with this.

app/models/contract.rb
peteryates: Good call, in ECF most are [greater than or equal to](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/npq_contract.rb#L8), but `recruitment_target` and `per_participant` are greater than.

app/models/statement.rb
peteryates: I'd prefer the month number here because it's easier to sort things.

Easy enough to convert to text with `Date::MONTHNAMES[1]` if we need to in the UI.

peteryates: I'm not averse to us using a Postgres enumb for them, that'll let us sort properly _and_ refer to them by name. 

```sql
create type month as enum ('January', 'February', ...);
```

[Apparently Rails supports this properly now too](https://blog.saeloun.com/2022/01/12/rails-7-adds-custom-enum-support-in-postgresql/).

app/models/contract.rb
cwrw: discussed in person: let's mirror what's already in ECF

app/models/statement.rb
cwrw: discussed in person: we'd like to stick to integers due to sorting being easier

app/models/statement_item.rb
mooktakim: I've checked prod, there is max 2 `statement_line_item` for a declaration.

cwrw: sounds like a good idea 👍 but not any other validation for now

config/routes.rb
```diff
@@ -43,7 +43,13 @@
   end
 
   namespace :admin do
-    resources :applications, only: %i[index show]
+    resources :applications, only: %i[index show] do
```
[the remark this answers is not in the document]
cwrw: yea seems to have snuck in :) @peteryates says we need to remerge `npq-separation` onto this branch as it's been rebased with main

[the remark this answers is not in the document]
cwrw: @peteryates found that the branches were out of sync with main, he's rebased `npq-separation` please rebase this branch with that and hopefully all will be well

db/schema.rb
```diff
@@ -64,6 +66,33 @@
...
   end
 
+  create_table "cohorts", force: :cascade do |t|
+    t.integer "start_year", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["start_year"], name: "index_cohorts_on_start_year", unique: true
+  end
+
+  create_table "contracts", force: :cascade do |t|
```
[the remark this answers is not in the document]
cwrw: chatted in person: 
```
  t.boolean "special_course" #  default: false, null: false
    t.integer "recruitment_target" # needs to be there
    t.decimal "per_participant" # Abeer to check if this is every `nil` on prod
    t.integer "output_payment_percentage" # default: 60
    t.integer "number_of_payment_periods" # Abeer to check if this is `nil`
    t.integer "service_fee_percentage" # default: 40
    t.integer "service_fee_installments"# Abeer to check if this is `nil`
    ```

File: db/schema.rb [hunk 18]
```diff
@@ -73,6 +102,17 @@
     t.integer "position", default: 0
     t.boolean "display", default: true
     t.string "identifier"
+    t.index ["identifier"], name: "index_courses_on_identifier", unique: true
+  end
+
+  create_table "declarations", force: :cascade do |t|
+    t.bigint "application_id", null: false
+    t.string "state"
```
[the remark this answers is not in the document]
cwrw: from @pmanrubia : to add a new pr for enum types for states

db/schema.rb
```diff
@@ -166,6 +206,15 @@
     t.index ["ukprn"], name: "index_local_authorities_on_ukprn"
   end
 
+  create_table "outcomes", force: :cascade do |t|
+    t.string "state"
```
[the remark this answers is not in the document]
cwrw: same as above

db/schema.rb
```diff
@@ -166,6 +206,15 @@
...
+    t.date "completion_date"
```
cwrw, line 211: add `null: false`

cwrw: think about naming for another time: should this be consistent with `_on` or `_date`

db/schema.rb
```diff
@@ -250,6 +299,33 @@
...
+    t.integer "month"
+    t.integer "year"
+    t.date "deadline_date"
+    t.bigint "cohort_id", null: false
+    t.bigint "lead_provider_id", null: false
+    t.datetime "marked_as_paid_at"
+    t.decimal "reconcile_amount", precision: 8, scale: 2
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.datetime "frozen_at"
```
cwrw, line 324: @cwrw to confirm the frozen concept

db/schema.rb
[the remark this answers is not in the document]
cwrw: `per_participant`, `service_fee_installments` and `number_of_payment_periods` are never `nil` so we can add validation for those

db/schema.rb
cwrw: so I can confirm now we are using `type` to freeze a statement, so when it becomes payable it is "frozen". We are not using type here, we can either rely on deadline date OR this field, so happy to keep for now

db/schema.rb
```diff
@@ -250,6 +299,33 @@
...
+    t.datetime "updated_at", null: false
+    t.bigint "clawed_back_by_id"
+    t.string "state"
+    t.index ["clawed_back_by_id"], name: "index_statement_items_on_clawed_back_by_id"
+    t.index ["declaration_id"], name: "index_statement_items_on_declaration_id"
+    t.index ["statement_id"], name: "index_statement_items_on_statement_id"
+  end
+
+  create_table "statements", force: :cascade do |t|
+    t.integer "month"
```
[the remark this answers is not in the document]
cwrw: @pmanrubia Checked with Tom, we need output fee for historic statements. We pay providers every month service fees and where we set this flag to true we attach declarations

app/models/declaration.rb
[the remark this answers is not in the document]
cwrw: after discussion I think we landed on maybe we'll have state machine validations at some point so will prob be there, resolving this for now

app/models/statement.rb
```diff
@@ -0,0 +1,30 @@
...
+#   and initializing clawbacks for any future voids. Refer to mark_as_paid field
+#   and logic.
+class Statement < ApplicationRecord
+  belongs_to :cohort
+  belongs_to :lead_provider
+  has_many :statement_items
+  has_many :contracts
+
+  validates :month, numericality: { in: 1..12, only_integer: true }
+  validates :year, numericality: { in: 2020..2030, only_integer: true }
```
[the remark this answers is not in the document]
cwrw: maybe this wasn't pushed as 2030 is still there?

app/models/statement.rb
```diff
@@ -0,0 +1,40 @@
...
+#   and logic.
+class Statement < ApplicationRecord
+  belongs_to :cohort
+  belongs_to :lead_provider
+  has_many :statement_items
+  has_many :contracts
+
+  validates :output_fee, presence: true
+  validates :month, numericality: { in: 1..12, only_integer: true }
+  validates :year, numericality: { greater_than_or_equal_to: 2020, only_integer: true }
```
peteryates, line 23: I think we should have a sensible max on here, just to prevent silly mistakes.

peteryates (approved):

PR #1043, opened 2023-12-15
Clean-up assigning Sentry issues

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1533

### DIFF
--- a/app/forms/questionnaires/qualified_teacher_check.rb
+++ b/app/forms/questionnaires/qualified_teacher_check.rb
@@ -177,7 +177,7 @@ module Questionnaires
     attr_reader :verified_trn
 
     def trn_to_store
-      if trn_verified?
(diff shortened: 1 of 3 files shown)


app/forms/questionnaires/qualified_teacher_check.rb
```diff
@@ -177,7 +177,7 @@ def ni_number_required?
     attr_reader :verified_trn
 
     def trn_to_store
-      if trn_verified?
+      if trn_verified? && verified_trn.present?
```
slawosz, line 180: I would say that if we have `trn_verified?` set to true, it would implied that we have `trn` present. Maybe it would be worth to rework this bit?

slawosz (approved):

PR #1044, opened 2023-12-18
Remove scopes from Course

pmanrubia:
### Description

We have a limited number of courses, each one of them with their unique `identifier`. 
If the identifier is unique having scopes filtering by identifier feels unnecessary.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,14 +1,16 @@
 class Course < ApplicationRecord
-  def self.npqeyl
-    find_by(identifier: "npq-early-years-leadership")
-  end
+  class << self
+    def npqeyl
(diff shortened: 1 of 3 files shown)


app/models/course.rb
```diff
@@ -1,35 +1,50 @@
...
+    end
+
+    def npqltd
+      find_by(identifier: NPQ_LEADING_TEACHING_DEVELOPMENT)
+    end
+
+    def ehco
+      find_by(identifier: NPQ_EARLY_HEADSHIP_COACHING_OFFER)
+    end
+  end
```
peteryates, line 14: Much nicer.

peteryates (approved):

PR #1066, opened 2024-01-05
Add schedule to domain

pmanrubia:
### Description

Add `Schedule` to our domain according to: https://github.com/DFE-Digital/npq-separation/pull/10 and https://github.com/DFE-Digital/npq-separation/pull/9.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -13,7 +13,7 @@ class Application < ApplicationRecord
   belongs_to :school, optional: true
   belongs_to :private_childcare_provider, optional: true
   belongs_to :itt_provider, optional: true
-  belongs_to :cohort, optional: true
+  belongs_to :schedule, optional: true
 
   has_many :declarations
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,10 +1,11 @@
 class Schedule < ApplicationRecord
   belongs_to :course_group
   belongs_to :cohort
+  has_many :applications
 
-  validates \
-    :name,
-    :declaration_start_date,
-    :starts_on,
-    :ends_on, presence: true
+  validates :name, presence: true
+  validates :declaration_starts_on, presence: true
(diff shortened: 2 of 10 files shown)


app/models/course.rb
```diff
@@ -1,4 +1,6 @@
 class Course < ApplicationRecord
+  belongs_to :course_group, optional: true
```
cwrw, line 2:
```suggestion
  belongs_to :course_group
```

from @peteryates : we need every course in a group, as that's the only way it can be linked to a schedule

db/migrate/20240105072457_add_course_group_to_courses.rb
```diff
@@ -0,0 +1,5 @@
+class AddCourseGroupToCourses < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :courses, :course_group, null: true, foreign_key: true
```
cwrw, line 3:
```suggestion
    add_reference :courses, :course_group, foreign_key: true
```

db/migrate/20240105073303_create_schedules.rb
```diff
@@ -0,0 +1,14 @@
+class CreateSchedules < ActiveRecord::Migration[7.0]
+  def change
+    create_table :schedules do |t|
+      t.string :name, null: false
+      t.date :declaration_start_date, null: false
+      t.date :starts_on, null: false
+      t.date :ends_on, null: false
```
cwrw, line 7:
```suggestion
      t.date :applies_from, null: false
      t.date :applies_to, null: false
```

db/migrate/20240108141151_rename_starts_on_column_in_schedules.rb
```diff
@@ -0,0 +1,5 @@
+class RenameStartsOnColumnInSchedules < ActiveRecord::Migration[7.1]
```
cwrw, line 1: maybe we can move this into the initial migration to avoid changes as this is a new table? (understand this might be hard for you locally to revert etc.)

PR #1067, opened 2024-01-05
Reconcile users and applications with ECF

ethax-ross:
### Context

As part of the NPQ separation work we need to be able to reconcile records between NPQ and ECF; this is the first pass at reconciling users and applications. It also contains a service to export orphaned records in Yaml format.

Example usage:

Ensure your `konduit.sh` outputs the database URL on connecting:

```
run_psql() {
  echo "$DB_URL"
  ...
}
```

Then connect to the relevant snapshot DBs (in separate terminals):

```
make ci production konduit-snapshot-ecf
make ci production konduit-snapshot-npq
```

You can then run the reconciliation from the Rails console:

```
ActiveRecord::Base.connection.execute("SET statement_timeout to 0")
Migration::Ecf::BaseRecord.connection.execute("SET statement_timeout to 0")

users_reconciler = Migration::ReconcileUsers.new
users_reconciler.matched.size
```

- Add ECF database connection

We want to be able to pull in ECF models; in order to do this we will need to connect to the ECF database as well as the NPQ database.

Add ECF database to database.yml and update Makefile to allow easier connecting to the NPQ and ECF snapshot databases.

Ensure `ApplicationRecord` connects to the primary (NPQ) database.

- Add models from ECF

In order to query across the two services we need the bare models from ECF.

Add `scenic` gem as the ECF db schema contains materialized views.

Port across the minimum ECF models/methods (convenience methods/aliases are added to make the reconciliation process simpler).

Pull the ECF db schema into NPQ. Hopefully we won't need the migrations.

- Add reconciliation service for User/Application models

Add a service that performs a reconciliation of `User` and `Application` records across NPQ and ECF.

Records are matched on specific attributes. If a match isn't found we treat this as an orphaned record. If more than one NPQ or ECF record is found for each match it is marked as duplicated.

Contains a simple service to export orphan details in Yaml format. As part of this we do a looser/tentative matching process to try and suggest records that _may_ be a match (but will require to be verified manually).


cwrw (approved):
brilliant 🧠

peteryates (approved):
🚀

PR #1077, opened 2024-01-09
Add the initial migration mechanism

ethax-ross:
### Context

We want to automate the ECF -> NPQ migration on a nightly basis within the migration environment and be able to easily see the status/outcome of the migration and iterate/improve on it going forwards.

- DRY up namespace checking logic

Clean up the logic to check if something belongs to ECF or NPQ by pushing it up into the `Reconciler` and abstracting the check into a `NamespaceCheck` service.

- Add skeleton migration service

Add a skeleton migration service that outputs reconciliation metrics to a new `Migration::Result` model.

This will form the basis of the migration script; the migration results will be exposed in a basic UI so that we can easily see the outcome of the migration.

- Flesh out migration service and add UI

Wraps the migration service in a job so that we can run it async and adds a basic UI to trigger and then display the current state/outcome of the migration.

- Add ability to download orphan reports

When we run the migration we have already queried all the data required to compile an orphan report.

Save the orphan reports in the Rails cache (not ideal, but the esiest way to have this work across multiple web instances) and add an endpoint to retrieve the reports for a given migration.

DRY up mocking the in-memory Rails cache.

- Fix indentation on konduit command

It should be indented twice to align with the correct yaml key.

- Minor performance optimizations

Loading the applications for the users (ECF in particular) takes a _long_ time. We use this when tentatively matching orphans on school. By touching the applications in the memoization it means we'll only do this for every user once, but as there are ~270k users it takes quite a while. If it becomes a problem we can look at further optimisations, such as a custom subquery.

- Example of how NPQ data can be updated during migration

Add a contrived example of how the NPQ data can be updated from its ECF counterpart during the migration process.

### Review guidance

As the orphan reports are stored in the Rails cache if you want them to work in development you'll need to enable caching with `rails dev:cache`.

| Start Migration      | Migration Running |  Migration Complete | 
| ----------- | ----------- | ----------- |
|   Screenshot 2024-01-11 at 15 38 08   |   Screenshot 2024-01-11 at 15 38 24    |    Screenshot 2024-01-11 at 15 39 10 |

### DIFF
--- a/app/helpers/migration_helper.rb
+++ b/app/helpers/migration_helper.rb
@@ -1,0 +1,27 @@
+module MigrationHelper
+  def migration_result_attributes(result, key)
+    result.attribute_names.filter { |attribute_name| attribute_name.include?(key) }
+  end
+
+  def migration_result_summary_row(result, key, attribute)
+    value = result.send(attribute)
+    total = result.send("#{key}_count")
+    percentage = number_to_percentage((value.to_f / total) * 100, precision: 0)
+    tag_color = migration_result_percentage_color(key, attribute, value)
+
+    content_tag(:div, class: "govuk-summary-list__row") do
+      content_tag(:dt, attribute.gsub(key.pluralize, "").humanize, class: "govuk-summary-list__key govuk-!-width-one-half") +
+        content_tag(:dd, number_with_delimiter(value), class: "govuk-summary-list__value") +
+        content_tag(:dd, class: "govuk-summary-list__actions") do
+          content_tag(:strong, percentage, class: "govuk-tag govuk-tag--#{tag_color}") if tag_color
+        end
+    end
+  end
+
+  def migration_result_percentage_color(key, attribute, value)
+    return if value.zero?
+    return :green if attribute.include?("matched")
+
+    :red if attribute != "#{key}_count"
+  end
+end
--- a/app/helpers/migration_helper.rb
+++ b/app/helpers/migration_helper.rb
@@ -24,4 +24,9 @@ module MigrationHelper
 
     :red if attribute != "#{key}_count"
   end
+
+  def migration_result_duration_in_words(result)
+    duration_in_seconds = (result.completed_at - result.created_at).to_i
+    ActiveSupport::Duration.build(duration_in_seconds).inspect
+  end
 end
--- a/app/services/migration/indexer.rb
+++ b/app/services/migration/indexer.rb
@@ -17,50 +17,98 @@ module Migration
     # perform a lookup on any matched objects in order to expand the
     # search as far as possible and infer matches.
     def lookup(obj, looked_up_objects = Set.new)
+      # End condition for recursion; we've already looked up this object.
       return Set[obj] if looked_up_objects.include?(obj)
 
+      # Add the object to the set of looked up objects.
       looked_up_objects.add(obj)
 
+      # For each indexed attribute.
       indexes
+        # Fetch the matching objects from the index.
         .map { |attr| fetch(obj, attr) }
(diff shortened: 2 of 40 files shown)


app/views/migrations/_most_recent_migration.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          </ul>
+        </div>
+        <div class="govuk-summary-card__content">
+          <dl class="govuk-summary-list">
+            <% migration_result_attributes(@most_recent_migration, key).each do |attribute| %>
+              <%= migration_result_summary_row(@most_recent_migration, key, attribute) %>
+            <% end %>
+          </dl>
+        </div>
+      <% end %>
```
peteryates, line 28: Doesn't matter too much for this but [these are fully supported in govuk-components](https://govuk-components.netlify.app/components/summary-list/#summary-cards).

peteryates: Something along the lines of:

```erb
<%=
  govuk_summary_card(title: "#{key.pluralize.titleize} Reconciliation") do |card|
    card.with_action do
      govuk_link_to("Orphan report", download_orphan_report_migrations_path(@most_recent_migration.id, key))
    end
    card.with_summary_list(rows: migration_result_summary_rows)
  end
%>
```

app/helpers/migration_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+    value = result.send(attribute)
+    total = result.send("#{key}_count")
+    percentage = number_to_percentage((value.to_f / total) * 100, precision: 0)
+    tag_color = migration_result_percentage_color(key, attribute, value)
+
+    content_tag(:div, class: "govuk-summary-list__row") do
+      content_tag(:dt, attribute.gsub(key.pluralize, "").humanize, class: "govuk-summary-list__key govuk-!-width-one-half") +
+        content_tag(:dd, number_with_delimiter(value), class: "govuk-summary-list__value") +
+        content_tag(:dd, class: "govuk-summary-list__actions") do
+          content_tag(:strong, percentage, class: "govuk-tag govuk-tag--#{tag_color}") if tag_color
```
peteryates, line 16: There's [tag support in govuk-components](https://govuk-components.netlify.app/components/tag/#coloured-tags) too.

```ruby
govuk_tag("some text", colour: tag_colour)
```

peteryates:
I know it's still in draft but I've had a look through and it looks great. Left some super minor comments plugging govuk-components 😅

app/views/migrations/_most_recent_migration.html.erb
ethax-ross: Thanks, I fell back to plain HTML here as I'm misusing the component a bit; the `govuk-components` seems to require the action to be a link (which makes sense and aligns with the component guidelines). It means I can't use the helpers though; unless I'm missing a way of getting a tag into the row actions, but from my understanding it expects a hash with the attributes for the underlying slot, something like:

```
[{
  key: { text: "Row title" },
  value: { text: "Row value" },
  actions: [<these have to be links>],
}]
```

app/services/migration/indexer.rb
```diff
@@ -17,50 +17,98 @@ def initialize(indexes, objects)
     # perform a lookup on any matched objects in order to expand the
     # search as far as possible and infer matches.
     def lookup(obj, looked_up_objects = Set.new)
+      # End condition for recursion; we've already looked up this object.
```
cwrw, line 20: 🙌

cwrw (approved):
🙌 great stuff! also thanks for adding the comments to the reconciler

PR #1087, opened 2024-01-11
Add separate deployment workflow for the migration environment

peteryates:
We now have a migration environment in Azure that currently works in the same way as sandbox - commits/merges to main trigger a deployment which is pushed to sandbox/migration/production at once.

This isn't what we want for our migration purposes. Instead we have a new npq-separation feature branch where all migration work/testing will take place. We want it to be deployed to the migration env whenever commits/merges happen on it.


peteryates:
Note this is based from `main` because we need the change to be present there too, we need to merge `main` into `npq-separation` after to pick up the changes.

PR #1095, opened 2024-01-15
Update state type with PostgreSQL native enum type: Outcomes

pmanrubia:
### Description

We are decided to use PostgreSQL native enum types to store the different 
types (statuses) of a number of entities within NPQ.

In this PR I am updating Outcomes so it validates at the DB level:

```ruby
STATES = %w[passed failed voided].freeze
validates :state, inclusion: { in: STATES }
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -19,6 +19,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_01_19_150601) do
 
   # Custom types defined in this database.
   # Note that some types may not work with other database engines. Be careful if changing database.
+  create_enum "outcome_states", ["passed", "failed", "voided"]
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -328,7 +330,7 @@
   create_table "statement_items", force: :cascade do |t|
     t.bigint "statement_id", null: false
     t.bigint "declaration_id", null: false
-    t.string "state", null: false
+    t.enum "state", null: false, enum_type: "declaration_status_enum"
```
peteryates, line 333: `state` here vs `status` above?

I'd probably lean towards `state` and drop the `_enum`.

cwrw: 👍 for `state` in both, and removing `_enum`, however could we also keep the declaration migrations in the declaration pr if possible?

cwrw (approved):

PR #1096, opened 2024-01-15
Update state type with PostgreSQL native enum type

pmanrubia:
### Description

We are decided to use PostgreSQL native enum types to store the different types (statuses) of a number of entities within NPQ.

In this PR I am updating Outcomes so it validates at the DB level:

```ruby
STATES = %w[submitted eligible ineligible payable voided paid awaiting_clawback clawed_back].freeze
validates :state, inclusion: { in: STATES }
```

### DIFF
--- a/db/migrate/20240115032403_create_declaration_status_enum.rb
+++ b/db/migrate/20240115032403_create_declaration_status_enum.rb
@@ -1,13 +1,9 @@
 class CreateDeclarationStatusEnum < ActiveRecord::Migration[7.1]
   def up
-    execute <<-SQL
-      CREATE TYPE declaration_status_enum AS ENUM ('eligible', 'payable', 'paid', 'voided', 'ineligible', 'awaiting_clawback', 'clawed_back');
-    SQL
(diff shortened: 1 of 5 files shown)


db/migrate/20240115032403_create_declaration_status_enum.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDeclarationStatusEnum < ActiveRecord::Migration[7.1]
+  def up
+    execute <<-SQL
+      CREATE TYPE declaration_status_enum AS ENUM ('eligible', 'payable', 'paid', 'voided', 'ineligible', 'awaiting_clawback', 'clawed_back');
```
peteryates, line 4: Any reason not to use `create_enum(:declaration_status, %w[eligible payable ...])` here?

peteryates: not the biggest fan of including the type of things as part of the name too, it's easy to see within PostgreSQL.

The states etc look fine though

cwrw (approved):

peteryates (approved):

PR #1097, opened 2024-01-15
Add a nightly production backup that runs at 03:30

peteryates:
### Context

The process is copied from ECF's. It uses pg_dump and then uploads the resulting file to a storage account. Unlike ECF's which runs at 03:00 this runs at 03:30.

### DIFF
--- a/.github/actions/backup-and-upload-database/action.yml
+++ b/.github/actions/backup-and-upload-database/action.yml
@@ -1,0 +1,70 @@
+name: Backup and Upload production DB
+description: Backup production DB and upload to azure
+
+inputs:
+  azure-credentials:
+    description: Azure credentials
+    required: true
+
+runs:
+  using: composite
+
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Login
+      uses: azure/login@v1
+      with:
+        creds: ${{ inputs.azure-credentials }}
+
+    - name: Set AKS credentials
+      shell: bash
+      run: az aks get-credentials -g s189p01-tsc-pd-rg -n s189p01-tsc-production-aks
+
+    - name: Install kubectl
+      uses: azure/setup-kubectl@v3
+
+    - name: Install konduit
+      shell: bash
+      run: make install-konduit
+
+    - name: Backup database
+      shell: bash
+      run: |
+        bin/konduit.sh npq-registration-production-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-production.sql.gz
+
+    - name: Set connection string
+      shell: bash
+      run: |
+        STORAGE_CONN_STR=$(az storage account show-connection-string -g s189p01-cpdnpq-pd-rg -n s189p01cpdnpqdbbkppdsa --query 'connectionString')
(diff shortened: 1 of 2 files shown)


.github/actions/backup-and-upload-database/action.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    - name: Backup database
+      shell: bash
+      run: |
+        bin/konduit.sh npq-registration-production-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-production.sql.gz
+
+    - name: Set connection string
+      shell: bash
+      run: |
+        STORAGE_CONN_STR=$(az storage account show-connection-string -g s189p01-cpdnpq-pd-rg -n s189p01cpdnpqdbbkppdsa --query 'connectionString')
```
peteryates, line 40: The storage account `s189p01cpdnpqdbbkppdsa` doesn't exist yet. @cwrw, did you just create this manually using Azure Portal in ECF?

.github/actions/backup-and-upload-database/action.yml
```diff
@@ -0,0 +1,70 @@
...
+        secrets: "SLACK-WEBHOOK"
+
+    - name: Notify Slack channel on job failure
+      if: failure()
+      uses: rtCamp/action-slack-notify@v2
+      env:
+        SLACK_USERNAME: CI Deployment
+        SLACK_TITLE: Database backup failure
+        SLACK_MESSAGE: Production database backup job failed
+        SLACK_WEBHOOK: ${{ steps.key-vault-secrets.outputs.SLACK-WEBHOOK }}
```
peteryates, line 68: We don't have a `SLACK-WEBHOOK` yet, I've put in a request for one and will add it once received.

.github/actions/backup-and-upload-database/action.yml
saliceti: It should be created automatically by https://github.com/DFE-Digital/npq-registration/blob/main/terraform/application/database.tf#L13

saliceti (approved):

.github/actions/backup-and-upload-database/action.yml
cwrw: when we checked however last time in ECF it already existed so this should exist already for NPQ?

cwrw: act nvm I checked and `s189p01cpdnpqdbbkppdsa` already exists so all good

.github/actions/backup-and-upload-database/action.yml
cwrw: maybe can be to the same shared alerts channel for NPQ/ECF?
as this is set to both now, our ECF specific one and the shared one?

.github/actions/backup-and-upload-database/action.yml
peteryates: Ah, thanks. I looked around the code and couldn't see any reference to it being built!

.github/actions/backup-and-upload-database/action.yml
peteryates: Good call, reusing the ECF prod one for now 👌🏽

PR #1121, opened 2024-01-25
Events

peteryates:
This is an experiment to see how we might sensibly log 'events' that happen within NPQ. It's not intended to be final, but a starting point for a conversation on how we save and present this data, which is **really valuable to the support teams when diagnosing problems**. Being able to show a concise timeline of what happened is necessary.

An event is something of note that happens within the app.

Currently in ECF we struggle with this and to show the 'change log' in the admin console we need to [carefully piece together what happened](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/participants/history_builder.rb) in order to build a timeline. This relies on PaperTrail's versions table which is built for record-level auditing purposes not to capture events which affect multiple related entities.

It could be presented using a timeline similar to [MoJ's Timeline component](https://design-patterns.service.justice.gov.uk/components/timeline/).

## How it works

There is an Event model and `events` table. The `events` table has multiple foreign key relationships, because **events don't happen in isolation**. Each one is optional (but it might make sense to have some validation rule requiring at least one is present). Having multiple related records instead of one (like PaperTrail's `object`) means we can look at a single event from multiple points of view.

For example, in the admin console when looking at a _person_ (teacher, applicant, whatever we end up calling them) we might see that they made 2 applications. When looking at the _school_ page in the admin console, using the exact same event data, we can see 3 applications were made by people at their school.

## Finding the events for a thing

```ruby
User.find(123).events                                            # all the events for user 123
Event.where(person: Person.find(123), school: School.find(456))  # events for user 123 at school 456
```

## Metadata

In addition to the foreign keys and Rails' default timestamps there are some fields we can use to add more information about what happened.

* `category` - the kind of thing that happened, 'Application made', 'Statement voided', etc.
* `subject` - a single line of text with an overview of what happened
* `description` - optional, more space to add details of the event
* `importance` - a number (for now) ranging between 1 (not very important) to 10 (critical) - perhaps useful for filtering or not overloading users with boring background stuff

### Application creation

This application was created by a teacher at a school:

```mermaid
graph TD
  Event --> Person
  Event --> School
  Event --> Course
```

While this one was created by someone at a private childcare provider:

```mermaid
graph TD
  Event --> Person
  Event --> PCP[Private Childcare Provider]
  Event --> Course
```

### Declaration voided

When a declaration is voided we could show the voiding event in the timelines of the related objects, like:

```mermaid
graph TD
  Event --> Declaration
  Event --> Application
  Event --> SLI[Statement Line Item]
  Event --> Statement
  Event --> LP[Lead Provider]
```

## Changes so far

- Add some missing relationships specs
- Add events table
- Add event model and relationship specs
- Add event relationships and corresponding specs
- Add event validation

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,0 +1,10 @@
+class Event < ApplicationRecord
+  belongs_to :user, optional: true
+  belongs_to :application, optional: true
+  belongs_to :course, optional: true
+  belongs_to :lead_provider, optional: true
+  belongs_to :school, optional: true
+  belongs_to :statement, optional: true
+  belongs_to :statement_item, optional: true
+  belongs_to :declaration, optional: true
+end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -7,4 +7,15 @@ class Event < ApplicationRecord
   belongs_to :statement, optional: true
   belongs_to :statement_item, optional: true
   belongs_to :declaration, optional: true
+
+  validates :importance,
+            presence: true,
+            numericality: {
+              only_integer: true,
+              in: 1..10,
+            }
+
+  validates :subject,
+            presence: true,
+            length: { maximum: 128 }
 end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -16,6 +16,7 @@ class Application < ApplicationRecord
   belongs_to :schedule, optional: true
 
   has_many :declarations
+  has_many :events
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
 
   scope :unsynced, -> { where(ecf_id: nil) }
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,5 +1,6 @@
 class Course < ApplicationRecord
   belongs_to :course_group, optional: true
+  has_many :events
 
   class << self
     def npqeyl
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -19,6 +19,7 @@ class Declaration < ApplicationRecord
   belongs_to :application
   has_many :outcomes
   has_many :statement_items
+  has_many :events
 
   # validates :declaration_date - this depends on the schedule
   # theoretically the supplier can make multiple declarations at once, that *is* valid
--- a/app/models/lead_provider.rb
+++ b/app/models/lead_provider.rb
@@ -76,6 +76,8 @@ class LeadProvider < ApplicationRecord
     "npq-leading-primary-mathematics" => LPM_PROVIDERS,
   }.freeze
 
+  has_many :applications
+
   scope :alphabetical, -> { order(name: :asc) }
 
   def self.for(course:)
--- a/app/models/lead_provider.rb
+++ b/app/models/lead_provider.rb
@@ -77,6 +77,7 @@ class LeadProvider < ApplicationRecord
   }.freeze
 
   has_many :applications
+  has_many :events
 
   scope :alphabetical, -> { order(name: :asc) }
 
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -3,6 +3,8 @@ class School < ApplicationRecord
   PRIMARY_PHASE = "Primary".freeze
   MIDDLE_DEEMED_PRIMARY_PHASE = "Middle deemed primary".freeze
 
+  has_many :events
+
   pg_search_scope :search_by_name,
                   against: [:name],
                   using: {
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -15,6 +15,7 @@
 class Statement < ApplicationRecord
   belongs_to :cohort
   belongs_to :lead_provider
+  has_many :events
   has_many :statement_items
   has_many :contracts
 
--- a/app/models/statement_item.rb
+++ b/app/models/statement_item.rb
@@ -10,6 +10,7 @@ class StatementItem < ApplicationRecord
   belongs_to :statement
   belongs_to :declaration
   belongs_to :clawed_back_by, class_name: "StatementItem", optional: true
+  has_many :events
 
   validates :state, inclusion: { in: STATES }
   validates :clawed_back_by, presence: true, if: -> { state == "clawed_back" }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,6 +4,7 @@ class User < ApplicationRecord
   has_many :applications, dependent: :destroy
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
   has_many :participant_id_changes
+  has_many :events
 
   scope :admins, -> { where(admin: true) }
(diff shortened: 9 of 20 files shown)


app/models/event.rb
```diff
@@ -0,0 +1,21 @@
+class Event < ApplicationRecord
+  belongs_to :user, optional: true
+  belongs_to :application, optional: true
+  belongs_to :course, optional: true
+  belongs_to :lead_provider, optional: true
+  belongs_to :school, optional: true
+  belongs_to :statement, optional: true
+  belongs_to :statement_item, optional: true
+  belongs_to :declaration, optional: true
```
peteryates, line 9: With all these being optional it would probably make sense to have a validation rule that checks at least one of them is present.

mattonfoot:
I think the main things to consider about an "events" log is what events would there actually be, what we want to get out of it, how granular would we want to go with it and what sort of information might we want to capture within in the record to make it useful.

Examples of potentially complicated events might be;

- Transfer of training to another LP or DP maybe because the previous one is no longer contracted to deliver or has decided not to continue running the training programmes.
- TRN changed because the previous one although valid, but one of three for the same person, was not the one the TRA wanted us to use.
- A "status label" change what it changed to and what it changed from
- A change of employer
- A change of funding eligibility
- Information being corrected on the record as opposed to actually being a record of a real world change in circumstance i.e. the school was changed on the record because there are several with the same name and the wrong one was on the record.

peteryates:
I think combining events with audits doesn't work because we audit (with PaperTrail at least) on a record-by-record basis. and events often relate to a number of things.

Events should be something that can be used by a non-developer admin to get an overview of the history of a _thing_ so they can work out why it's in its current state.

I think trying to encapsulate the state of all the related records at a point in time is more complex than we need for an event log, but we could do that how we do now via PaperTrail.

> Some events may also be recorded from different aspects, for instance a lead provider my indicate that a participant has stopped training or the participant may inform us that they have stopped training. How would we ensure that the events like this are recorded the same no matter where they stem from or what process was performed that resulted in the change.

This is an interesting one and we have a few options. We could:

* have different event categories/types, `LeadProviderStopsTrainingParticipant` vs `ParticipantStopsTraining`.
* have a `ParticipantStopsTraining` event category and say who instigated the stop in the `subject` field.
* model the stoppage as a database table or state and not rely on events for the information.
* add a field to the events table that indicates the instigator

I think overall the direction I want to nudge us is:

* the way we record events should play to the database's strengths and rely on referential integrity (ie. not Rails' polymorphic relationships)
* events should be related to multipe things
* we should explicitly create them at appropriate times in the service layer/business logic, not with callbacks
* they should add context and be useful for debugging but not be central to the operation of the service, that data should be within the main data model

peteryates:
A quick brain dump of how this data might end up looking in the admin console, showing how we can surface the relevant bits depending on the lens we're looking through.

image

frankieroberto:
I like this idea. Makes sense to encapsulate events at a higher level of abstraction than papertrail audits, where you often get tons of records changed for a single event.

A few quick thoughts:

Adding optional foreign keys makes sense (polymorphic associations go wrong all the time) - but are there any event types where you would want a single conceptual 'event' to reference multiple instances of a record type? eg if there are any bulk actions where a user can accept/reject multiple applications in a single click? Or alternatively, if the action is to transfer something from School A to School B?   I think to represent this, you'd have to switch the table reference columns to arrays (but lose referential integrity), or move all the foreign keys to a separate `event_relationships` table which references the `events` table?

Are all events triggered by a user? If so, maybe that should be non-optional? (Presuming that any API requests are also associated with a user).

If 'category' is a defined list (rather than complete free text), then it might be worth moving that to a separate `EventCategory` table to help enforce it? Possibly the `importance` score could move there too, if importance is purely determined by the category type. That'd make it easier to adjust the scores without having to update hundreds of records.

Finally, the one downside of using relationships to other records rather than copies of them, is that this can be confusing in the interface if the record gets renamed afterwards - ie the timeline might mention someone applying to a course named X even though at the time it was called something else.  However this risk is minimised if, generally, things don't get renamed and new records get created instead (schools / courses / etc).  For people's names though, I think it's less confusing if the name gets applied retrospectively.

Good luck!

peteryates:
> Or alternatively, if the action is to transfer something from School A to School B?

This is an interesting case and it comes down to how important we think a transfer is and how we model it internally.

If the transfer itself is a thing we need to attach data to (like a football transfer has a fee and a medical) then we should represent it in the main data model and attach the event to it too.

But, if it's simply the end of one period of employment and the subsequent beginning of another then it could be represented as two events, a 'Person leaves School A' and Person joins School B'.

The advantage of separate events is that when looking at what happened from either school's point of view, we don't need any extra code to work out who should be able to see what.

Of couse, there are times where we might want to group lots of events together. If we run a data tidy up script then maybe we could group them like you suggest:

```ruby
ActiveRecord::Base.transaction do
  EventGroup.create(name: "Lowercase all the emails").tap |event_group|
    people_with_mixed_case_email_addresses.each { |p| p.update!(email_address: p.email_address.lower) }
    Event.create!(category: "EmailAddressFix", event_group:, description: "...")
  end
end
```

This could work for two events too, like in a transfer situation.

> If 'category' is a defined list (rather than complete free text), then it might be worth moving that to a separate EventCategory table to help enforce it?

This is a bit of a tough one and I haven't entirely made my mind up. Adding it as a table would make sense if we wanted it to be editable by non-devs, which we _probably don't_. This might be another case for a [database enum](https://github.com/DFE-Digital/npq-registration/blob/npq-separation/db/schema.rb#L22).

> Finally, the one downside of using relationships to other records rather than copies of them, is that this can be confusing in the interface if the record gets renamed afterwards

Yeah, totally agree with that.

I like the idea of populating the `subject` and `description` text with the relevant data at the time the event's created, so it's representative of what actually happened. The if the school changes name the events still show up for it (which we definitely want) but the event descriptions remain as they were.

tonyheadford:
I like the idea of an events log very much. Presumably this relies on us ensuring we have well defined actions being tied to a service object layer or something like that i.e. not trying to hook event tracking into ActiveRecord

peteryates:
> not trying to hook event tracking into ActiveRecord

Yes, exactly this. Keeping the business logic all together in one place.

frankieroberto:
> This is a bit of a tough one and I haven't entirely made my mind up. Adding it as a table would make sense if we wanted it to be editable by non-devs, which we probably don't. This might be another case for a [database enum](https://github.com/DFE-Digital/npq-registration/blob/npq-separation/db/schema.rb#L22).

There's definitely pros and cons to table vs enum. I've often found that with enums, you often end up wanting to attach extra metadata to them later, and so might then need to move them to a table. (although alternatively can hardcode within code).

If the `importance` score is associated with the category, then _maybe_ it makes sense to have a separate table, and possibly even allow that score to be editable by admins? But the rows could still be only added/deleted via data migrations?  Possibly you might have a human-readable description for each category which could be editable too?

Tricky trade-off!

PR #1123, opened 2024-01-29
CPDNPQ-1579 Show Expired Info on Registrations in Account

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1579

When a registration status is changed to ‘Expired’ the tag next to ‘provider application’ displays as ‘Expired’. And has a grey background.

If the service is closed, this text displays below the tag:

Your registration has expired but you can register again for future courses when registrations open.

If the service is open, this text displays below the tag: 

Submit a new registration if you still want to take this course. 

‘Submit a new registration’ text links to the registration flow (the same as the ‘Register for another NPQ’ button at the bottom of the account page). 

All the expired registrations are grouped together on the account page. 

Above the expired registrations is a h3, large size heading ‘Expired registrations’

Expired registrations have no ‘View details’ link

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -17,6 +17,8 @@ class Application < ApplicationRecord
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
 
   scope :unsynced, -> { where(ecf_id: nil) }
+  scope :expired_applications, -> { where(lead_provider_approval_status: "rejected").where("created_at < ?", cut_off_date_for_expired_applications) }
+  scope :active_applications, -> { where.not(id: expired_applications) }
 
   enum kind_of_nursery: {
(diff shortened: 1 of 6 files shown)


app/models/application.rb
```diff
@@ -17,6 +17,8 @@ class Application < ApplicationRecord
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
 
   scope :unsynced, -> { where(ecf_id: nil) }
+  scope :expired_applications, -> { where(lead_provider_approval_status: "rejected").where("created_at < ?", Time.zone.local(2024, 4, 2)) }
```
slawosz, line 20: Maybe we could move this date to some config value (still hardcoded).

slawosz (approved):

Issue #1130, opened 2024-02-02
Should users vs teachers be separated into different models?


peteryates:
The app currently has a single `users` table that contains two kinds of users:

* admins
* registrants (participants/teachers)

Each of these have abilities the others don't:

* admins can:
  * see the admin screens and make changes to app settings
  * log in with their email address via a magic code
  * only be added by other admins
* registrants can:
  * register for NPQs and have them linked to their account
  * register themselves
  * log in/identify themselves via TRA OpenID Connect (to see the status of their applications?)

We don't have a need for the same people to be both admins and registrants, so it doesn't really make sense for us to keep these two kinds of users together. I think they should be separate, with names that make it clear who they are and what they need.

Moving the non-admins to a separate table like `teachers` (or `applicants`, or `registrants`) seems sensible. Thoughts?

peteryates:
Solved by #1137

PR #1136, opened 2024-02-06
Splitting up users

peteryates:
Refs #1130

## Changes so far

- Add participants model with required fields only
- Add a couple of example scopes and specs
- Add some very basic user to participant sync tasks
- Link applications to participants instead of users


peteryates:
Note, it's probably a better idea to keep the participants in the `users` table _then_ rename it to `participants` as all the foreign keys etc will continue to work, and there are only approx 30 admins to worry about.

Thanks @cwrw for the idea 💡

PR #1142, opened 2024-02-12
Skip ConnectionError class from sentry logs

javier-npq:
### Context

We have been seeing too many JsonApi errors lately, this pr is an effort to silence some of those

[What has been changed?]

failures:

### DIFF
--- a/app/jobs/application_submission_job.rb
+++ b/app/jobs/application_submission_job.rb
@@ -71,6 +71,8 @@ private
       error_messages: ["#{e.class} - #{e.message}"],
       response_body:,
     )
+    return if e.is_a?(JsonApiClient::Errors::ConnectionError)
+
     Sentry.with_scope do |scope|
       scope.set_context("User", { id: user.id })
       Sentry.capture_exception(e)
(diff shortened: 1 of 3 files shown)


app/jobs/application_submission_job.rb
```diff
@@ -71,14 +71,16 @@ def find_ecf_user_by_email(user:)
       error_messages: ["#{e.class} - #{e.message}"],
       response_body:,
     )
-    Sentry.with_scope do |scope|
-      scope.set_context("User", { id: user.id })
-      Sentry.capture_exception(e)
+    unless e.is_a?(JsonApiClient::Errors::ConnectionError)
```
slawosz, line 74: I much more prefer `guard close` instead of wrapping all the things into `unless` block with can be quite misleading:

```
return if e.is_a?(JsonApiClient::Errors::ConnectionError)
```
what do you think?

javier-npq: Yeah that's more easier to read.

slawosz (approved):

slawosz (approved):

PR #1147, opened 2024-02-15
Create super admins on login using env variables

slawosz:


app/controllers/omniauth_controller.rb
```diff
@@ -106,3 +108,15 @@ def try_to_extract_error_type
 rescue StandardError
   "unknown-error-type"
 end
+
+def try_to_mark_super_user
+  return unless Rails.env.review?
+
+  ENV["SUPER_ADMIN_SEED"].split(";").each do |email|
+    User.find_by(email:).update!(admin: true, super_admin: true)
```
peteryates, line 116: Just a heads up, I moved admins to their own table/model today.

peteryates: I also made the seeds generate a super admin. I wonder if that solves this problem too?

PR #1148, opened 2024-02-15
Disable NPQ separation functionality by environment

ethax-ross:
[Jira-2829](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2829)

### Context

We want to be able to develop/merge NPQ separation changes into `main` without exposing them on all environments until we're ready. In order to do this we need a mechanism to exclude routes from certain environments (we are favouring this high-level approach instead of peppering feature checks throughout the codebase).

- Bring in skeleton API endpoints

These were implemented by Pete as part of the NPQ separation spike and provides a nice set of boilerplate controllers/actions we can work from. It also means we can write tests against the routes.

- Add per-env config variables to control NPQ separation access

We only want to allow access to the NPQ separation API endpoints, admin area and migration functionality for non-production environments initially.

Disable access for all environments apart from development, test, review, staging and migration. We may want to allow access in sandbox in once we've briefed providers and then finally production.

- Add stub controllers for NPQ separation admin/migration endpoints

We will be adding routes and controlling access per environment; in order to test we need something to route to.

- Restrict access to NPQ separation routes

While we are migrating across from ECF and developing the functionality in NPQ registration we don't want to allow access to the new routes until we are ready.

Constrain access according to the environment configuration.

### Guidance to review

NPQ separation functionality will initially be enabled everywhere apart from sandbox and production, which are the provider-facing environments.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -153,6 +153,16 @@ Rails.application.routes.draw do
     end
   end
 
+  namespace :npq_separation do
+    namespace :admin do
+      resources :admins, only: %i[index]
+    end
+
+    namespace :migration do
+      resources :migrations, only: %i[index]
+    end
+  end
+
   resource :csp_reports, only: %i[create]
 
   get "/404", to: "errors#not_found", via: :all
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -85,18 +85,15 @@ Rails.application.routes.draw do
         post :accept, path: "accept"
       end
 
-      namespace :participants do
-        resources :npq, only: %i[index show] do
-          put :change_schedule, path: "change-schedule"
-          put :defer, path: "defer"
-          put :resume, path: "resume"
-          put :withdraw, path: "withdraw"
-
-          get :outcomes, path: "outcomes"
-        end
-
-        resources :outcomes, only: %i[index]
+      resources :participants, only: %i[index show], path: "participants/npq" do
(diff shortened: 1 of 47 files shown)


spec/routing/npq_separation_routes_spec.rb
peteryates, line 1: Tidy.

peteryates (approved):
I like this, super neat and a great starting off point.

The `Api` looks a bit weird, should we (as a follow up) switch to `API` with an inflection?

config/routes.rb
```diff
@@ -79,6 +79,89 @@
...
+      end
+    end
+
+    namespace :v2, constraints: ->(_request) { Rails.application.config.npq_separation[:api_enabled] } do
+      resources :applications, path: "npq-applications", only: %i[index show] do
+        post :reject, path: "reject"
+        post :accept, path: "accept"
+      end
+
+      resources :enrolments, path: "npq-enrolments", only: %i[index]
```
cwrw, line 111: not sure we need this?

cwrw (commented):
nicely done 🙌  just the one comment

config/routes.rb
cwrw: urgh this is the csv endpoint, ignore

cwrw (approved):

PR #1161, opened 2024-02-19
Add cohort model

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2827

### Description

Adds Cohort model following the work on the branch: `npq-separation`

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -7,6 +7,15 @@ class Cohort < ApplicationRecord
             },
             uniqueness: true
 
+  validates :registration_start_date, presence: true
+  validate :registration_start_date_matches_start_year
+
+  def registration_start_date_matches_start_year
+    unless registration_start_date && registration_start_date.year == start_year
+      errors.add(:registration_start_date, "year must match the start year")
+    end
+  end
+
   def self.current
     where(registration_start_date: ..Time.zone.today)
       .order(start_year: :desc)
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -1,0 +1,15 @@
+class Cohort < ApplicationRecord
+  validates :start_year,
+            presence: true,
+            numericality: {
+              greater_than_or_equal_to: 2021,
+              less_than: 2030,
+            },
+            uniqueness: true
+
+  def self.current
+    where(registration_start_date: ..Time.zone.today)
+      .order(start_year: :desc)
+      .first!
+  end
+end
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -11,9 +11,9 @@ class Cohort < ApplicationRecord
   validate :registration_start_date_matches_start_year
 
   def registration_start_date_matches_start_year
-    unless registration_start_date && registration_start_date.year == start_year
-      errors.add(:registration_start_date, "year must match the start year")
-    end
+    return if registration_start_date.blank?
+
+    errors.add(:registration_start_date, "year must match the start year") if registration_start_date.year != start_year
   end
 
   def self.current(timestamp = Time.zone.today)
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -16,8 +16,8 @@ class Cohort < ApplicationRecord
     end
   end
 
-  def self.current
-    where(registration_start_date: ..Time.zone.today)
+  def self.current(timestamp = Time.zone.today)
+    where(registration_start_date: ..timestamp)
       .order(start_year: :desc)
       .first!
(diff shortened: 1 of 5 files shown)


app/models/cohort.rb
```diff
@@ -0,0 +1,13 @@
...
+  validates :start_year,
+            presence: true,
+            numericality: {
+              greater_than_or_equal_to: 2021,
+              less_than: 2030,
+            },
+            uniqueness: true
+
+  def self.current
+    find_by!(start_year: Time.zone.today.year)
```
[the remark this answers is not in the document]
cwrw: we need something similar, we currently use `npq_registration_start_date` in ECF app so we should also be porting it here I think as well

db/migrate/20240219082737_add_registration_start_date_to_cohorts.rb
```diff
@@ -0,0 +1,5 @@
+class AddRegistrationStartDateToCohorts < ActiveRecord::Migration[7.1]
+  def change
+    add_column :cohorts, :registration_start_date, :datetime, null: false
```
cwrw, line 3: could we include this in the same migration please as this pr is introducing the table?

peteryates: Yeah I agree, I tend to `db:rollback` and `db:migrate` when working on a branch so there are fewer separate migrations per PR

app/models/cohort.rb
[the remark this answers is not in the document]
cwrw: I think it's worth checking how we drive this logic first, as discussed in person with Pablo this morning around dropping the academic year as that's more ECF related and relying on the registration start date for everything. If that's the case we probably can encapsulate that within the model well using scopes, if not we can talk about service classes, but this feels quite different to for instance other services we will have with business logic so keen to see how it looks first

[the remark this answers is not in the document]
cwrw: I think the main problem is we never know when the end date is, they usually change per year and we know when the next cohort rolls in

db/migrate/20240216095511_create_cohorts.rb
```diff
@@ -0,0 +1,12 @@
+class CreateCohorts < ActiveRecord::Migration[7.1]
+  def change
+    create_table :cohorts do |t|
+      t.integer :start_year, null: false
+      t.datetime :registration_start_date
```
cwrw, line 5: maybe this should also be `null: false`? if we plan to populate this from previous cohorts 🤔 let me know your thoughts

app/models/cohort.rb
```diff
@@ -0,0 +1,15 @@
+class Cohort < ApplicationRecord
+  validates :start_year,
+            presence: true,
+            numericality: {
+              greater_than_or_equal_to: 2021,
```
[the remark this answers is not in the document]
cwrw: also there is something called `NPQ_PLUS_1_YEAR`, might be worth digging what that really means 😓 in ECF

app/models/cohort.rb
```diff
@@ -0,0 +1,15 @@
+class Cohort < ApplicationRecord
+  validates :start_year,
```
peteryates, line 2: Might also be worth adding a validation for the `registration_start_date` here so that the year is the same as the `start_year`?

spec/models/cohort_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        new_cohort = Cohort.new(start_year: existing_cohort.start_year)
+
+        new_cohort.valid?
+        expect(new_cohort.errors[:start_year]).to include("has already been taken")
+      end
+    end
+  end
+
+  describe ".current" do
+    let(:timestamp) { Time.zone.today }
```
cwrw, line 46: ✂️

cwrw (approved):
thanks @pmanrubia

app/models/cohort.rb
```diff
@@ -0,0 +1,24 @@
...
+            uniqueness: true
+
+  validates :registration_start_date, presence: true
+  validate :registration_start_date_matches_start_year
+
+  def registration_start_date_matches_start_year
+    unless registration_start_date && registration_start_date.year == start_year
+      errors.add(:registration_start_date, "year must match the start year")
+    end
+  end
```
peteryates, line 17:
```suggestion
  def registration_start_date_matches_start_year
    return if registration_start_date.blank?
    
    errors.add(:registration_start_date, "year must match the start year") if registration_start_date.year != start_year
  end
```

PR #1162, opened 2024-02-19
Fix inconsistent node and postgresql versions

mattonfoot:
### Context

The versions of Node and PostgreSQL docker containers are misaligned and vary greatly between various test and install files causing issues with clean installs and testing.

- align all node versions to 16.20.2
- align all PostgreSQL versions to 14 to match production terraformed infrastructure


peteryates (approved):
Ah well spotted, I missed this!

PR #1168, opened 2024-02-19
CPDNPQ-1678 Retain TSF Info on Registration Details for Past Registrations

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1678

- Use the existing TSF feature flag 

- On the registration details pages, if the person is marked as eligible for TSF a row should display under the Scholarship funding row. 

- When TSF is shown, an ‘eligible’ green tag should display and below the tag the content: 
 
- Your workplace will receive a targeted support funding payment to support you to do this NPQ.

- If they are not eligible for TSF or the registration is submitted post-April 2024 no info about TSF should display.

### DIFF
--- a/app/views/accounts/_funding_details.html.erb
+++ b/app/views/accounts/_funding_details.html.erb
@@ -28,14 +28,14 @@
 
         <% scholarship_funding_row(application) %>
 
-        <% if Flipper.enabled?(:targeted_support_funding) %>
+        <% if Flipper.enabled?(:targeted_support_funding) || application.created_at < APRIL_2024_CUTOFF_DATE %>
           <div class="govuk-summary-list__row">
             <dt class="govuk-summary-list__key">
               Targeted support funding
             </dt>
             <dd class="govuk-summary-list__value">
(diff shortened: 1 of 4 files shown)


app/views/accounts/_funding_details.html.erb
```diff
@@ -28,14 +28,14 @@
 
         <% scholarship_funding_row(application) %>
 
-        <% if Flipper.enabled?(:targeted_support_funding) %>
+        <% if Flipper.enabled?(:targeted_support_funding) || application.created_at < Time.zone.local(2024, 4, 2) %>
```
slawosz, line 31: It would be lovely to have one source of `magic numbers`. How about file `config/initializers/constants.rb` and there:
```ruby
APRIL_2024_CUTOFF_DATE = Time.zone.local(2024, 4, 2)
CURRENT_CUTOFF_DATE = APRIL_2024_CUTOFF_DATE  # in case we need generic cutoff
```
This way we will have all constants in one page. I don't think we need to worry about namespacing them as the app is rather small.

PR #1169, opened 2024-02-20
Update LeadProvider model for parity with ECF

ethax-ross:
[Jira-2828](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2828)

### Context

As part of the NPQ separation work we want to update the existing `LeadProvider` model to make it consistent with ECF in terms of relationships and validation.

- Add missing applications relationship to LeadProvider

Add a missing relationship between `LeadProvider` and `Application`.

Add test coverage for relationships in general.

- Add missing validation to LeadProvider model

Add missing `name` validation to `LeadProvider` model, as per ECF.

Add missing test coverage for `alphabetical` scope.


db/migrate/20240219160343_add_vat_chargeable_to_lead_providers.rb
```diff
@@ -0,0 +1,5 @@
+class AddVatChargeableToLeadProviders < ActiveRecord::Migration[7.1]
+  def change
+    add_column :lead_providers, :vat_chargeable, :boolean, default: true
```
cwrw, line 3: are we sure we use `vat_chargeable` in NPQ calculations, I couldn't find it in the ECF app, do you mind pointing me to it?

ethax-ross: I had assumed since its in the schema for NPQ lead providers that we would want it, but it doesn't look like its used at the moment (just in the ECF statement calculator). I'll remove it if its not going to be needed.

cwrw (approved):

PR #1174, opened 2024-02-22
Begin work on the new admin console

peteryates:
The current NPQ admin console is serviceable but it's very much geared towards NPQs role in pushing data into ECF.

Once separation from ECF is complete, NPQ will be responsible for many more functions including keeping track of and managing:

* statements
* declarations
* extra lead provider information and API credentials
* cohorts
* participant outcomes
* financial schedules

In addition to separation, the long term goal of the programme is to make NPQ self-sufficient. This means that the course should be able to run without developer intervention. Currently in ECF developers are needed for lots of things, like adding new cohorts and importing schedules. These operations will be possible via this new admin console.

As a result, we will need to build many more screens.

The current admin console has a few and its single nav bar is already overwhelmed to the point it requires horizontal scrolling.

image

In the new version some measures have been taken to address that. Note that the content in these screenshots is just placeholder for now and will be revised over the coming weeks:

### Two levels of navigation

It seems sensible to segment the admin console into _areas_ of related functionality, like applications, user management, finance. These are included in the (horizontal) primary navigation.

Each of these areas will probably have a few screens within, for example under finance we might have 'Statements' and 'Schedules'.

This subnavigation is vertical and to the left of the content.

image

### A wider layout

To accommodate the vertical navigation the template has been made a bit wider. We know from user research that our users predominantly work on computers rather than cramped mobile devices.

image

## Review guidance

Probably easiest to review commit-by-commit.


app/views/layouts/admin/_primary-navigation.html.erb
peteryates, line 1: This is intentionally hardcoded, will be replaced with a component in a follow up PR

app/views/layouts/admin/_sub-navigation.html.erb
peteryates, line 1: This is intentionally hardcoded, will be replaced with a component in a follow up PR.

cwrw (approved):

PR #1176, opened 2024-02-22
[CPDLP-2834] initial models for Statement and StatementItem

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2834

### State machine library options

| Library  | Latest Release | Ruby Toolbox Score |
| ------------- | ------------- | ------------- |
| [State Machines](https://github.com/state-machines/state_machines)  | 2023-06-30 | 4.04 |
| [Acts as state machine (AASM)](https://github.com/aasm/aasm) | 2023-02-05 | 3.77 |
| [Statesman](https://github.com/gocardless/statesman) | 2024-01-09 | 0.80 |

* `Statement` model
* `StatementItem` model
* Proposed state implementation with `state_machines-activerecord` gem
* Added `docs/state_machines.md`, a place for state diagrams


app/models/statement.rb
```diff
@@ -0,0 +1,33 @@
...
+  validates :output_fee, presence: true
+  validates :month, numericality: { in: 1..12, only_integer: true }
+  validates :year, numericality: { in: 2020..2050, only_integer: true }
+
+  validate :validate_max_statement_items_count
+
+  state_machine :state, initial: :open do
+    state :open
+    state :payable
+    state :paid
```
peteryates, line 15: So is it definitely correct to say that every statement ever takes this route? None are cancelled or closed any other way?

db/migrate/20240221113542_create_statements.rb
```diff
@@ -0,0 +1,21 @@
...
+      t.integer :month, null: false
+      t.integer :year, null: false
+      t.date :deadline_date
+      t.date :payment_date
+      t.boolean :output_fee, default: true, null: false
+      t.references :cohort, null: false, foreign_key: true
+      t.references :lead_provider, null: false, foreign_key: true
+      t.datetime :marked_as_paid_at
+      t.decimal :reconcile_amount, precision: 8, scale: 2
+      t.datetime :frozen_at
```
peteryates, line 15: What does frozen mean here? Can `open`, `payable` and `paid` statements all be frozen?

docs/state_machines.md
```diff
@@ -0,0 +1,38 @@
...
+## Statement
+
+```mermaid
+---
+title: State
+---
+stateDiagram-v2
+    [*] --> open
+    open --> payable: mark_payable
+    payable --> paid: mark_paid
```
peteryates, line 12: I like this.

I think we can automate it. Looking at the gem we can get a list of all possible states like this:

```ruby
Statement.state_machines[:state].states.map(&:name)
=> [:open, :payable, :paid]
```

Then we can loop through and do something [similar to the official gem](https://github.com/state-machines/state_machines-graphviz/blob/master/lib/state_machines/graphviz/monkeypatch.rb) but build up Mermaid transitions.

Could even be a rake task where we find and replace in this file.

peteryates:
Looks good to me overall. Just left a couple of small questions on the `Statement` states. To the uninitiated, it looks like `frozen` is a state.

db/migrate/20240221113542_create_statements.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    create_table :statements do |t|
+      t.integer :month, null: false
+      t.integer :year, null: false
+      t.date :deadline_date
+      t.date :payment_date
+      t.boolean :output_fee, default: true, null: false
+      t.references :cohort, null: false, foreign_key: true
+      t.references :lead_provider, null: false, foreign_key: true
+      t.datetime :marked_as_paid_at
```
peteryates, line 13: Another question to the uninitiated, what's the difference between `payment_date` and `marked_as_paid_at` (other than it being a `datetime`)

db/migrate/20240221113542_create_statements.rb
mooktakim: @peteryates 
`frozen_at` is a new field added to the separation branch: https://github.com/DFE-Digital/npq-registration/commit/4380f4ed1973090f03597aa56d2a4e3c8a4eff18

I'm not sure if we need it, the statement should be "frozen" once its paid.

app/models/statement.rb
mooktakim: I don't think there's ever been a cancelled statement. I believe there is always a statement for each month.
Once the statement is `paid` its no longer added to, and shouldn't change.

db/migrate/20240221113542_create_statements.rb
mooktakim: `marked_as_paid_at` when we action statement to be paid, it is used to display some text and for background job:
https://dfedigital.atlassian.net/browse/CPDLP-2295

db/migrate/20240221113542_create_statements.rb
peteryates: Ah, perhaps @pmanrubia or @cwrw can shed some light.

db/migrate/20240221113542_create_statements.rb
peteryates: Ah ok, so:

* `payment_date` is the date on which the payment _was_ made
* `marked_as_paid_at` is the date and time *we recorded it* as being paid

app/models/statement.rb
peteryates: Ok, makes sense - thanks!

db/migrate/20240221113542_create_statements.rb
cwrw: Having a look at the comment here https://github.com/DFE-Digital/npq-registration/pull/1036 basically in ECF currently we use the type "Payable" to allow freezing a statement in our logic. I think we can probably rely on deadline date or the enum state, so probably we can drop this

app/models/statement_item.rb
```diff
@@ -0,0 +1,52 @@
+class StatementItem < ApplicationRecord
```
cwrw, line 1: curious around the naming decision (I guess that's what we landed on with the analysis) vs line item?

cwrw (commented):
very nicely done 🙌  just the comment around frozen at otherwise 👍

app/models/statement_item.rb
peteryates: I'm not massively invested in either line or item but they both mean the same thing here so using both doesn't feel like it makes sense.

mooktakim: In the analysis, we talked about this just being a shorter maybe name.

But I wonder, maybe we should instead call it `StatementDeclaration`, as it closely describes the many-to-many relationship.

db/migrate/20240221113542_create_statements.rb
mooktakim: I've removed `frozen_at` for now, if we need it, we can add.

cwrw (approved):

app/models/statement_item.rb
cwrw: I think that sounds reasonable, the only thing that makes me hesitant is that there is a `state` on the record, so happy to stick to items for now.

PR #1185, opened 2024-02-27
Connect to the ECF database

ethax-ross:
[Jira-2857](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-2857)

### Context

We want to be able to connect to the corresponding ECF database in each environment (apart from review as we don't spin up an ECF database here).

- Connect to the ECF database

Update database config with ECF connections. 
Update `ApplicationRecord` to connect to the NPQ reg database.
Add a `BaseRecord` for the ECF models.
The ECF schema is included so that we can generate a test database (the `scenic` gem is needed for the ECF managed DB views).

- Only migrate primary DB on boot

We don't ever want to migrate the ECF database from NPQ reg.

- Make Ecf::BaseRecord readonly

We don't want to accidentally write to an ECF database.

### Guidance to review

I checked the keyvaults and the `ECF_DATABASE_URL` secret exists for the migration, staging, sandbox and production environments.

Having the `ApplicationRecord` explicitly connecting to the primary DB should be a good enough safety net; we will restrict access to other models connecting to the NPQ reg database at a higher-level (via routes).

I manually deployed this to the NPQ migration environment and it booted up fine/functions. You can do that with:

```
DOCKER_IMAGE=ghcr.io/dfe-digital/npq-registration:baf18e13a2e5b8c9dc024fd58ba8036cc3e7b3d4 make migration terraform-apply
```


app/models/migration/ecf/base_record.rb
```diff
@@ -0,0 +1,7 @@
+module Migration::Ecf
+  class BaseRecord < ApplicationRecord
+    self.abstract_class = true
+
+    connects_to database: { reading: :ecf, writing: :ecf } unless Rails.env.review?
```
cwrw, line 5: maybe we can use the api separation flag instead here to avoid connecting in prod for now?

db/ecf_schema.rb
```diff
@@ -0,0 +1,1474 @@
...
+#
+# This file is the source Rails uses to define your schema when running `bin/rails
+# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
+# be faster and is potentially less error prone than running all of your
+# migrations from scratch. Old migrations may fail to apply correctly if those
+# migrations use external dependencies or application code.
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema[7.0].define(version: 2024_02_06_170145) do
```
cwrw, line 13: Im wondering how will we keep this up to date with ECF going forward?

Dockerfile
```diff
@@ -68,4 +68,4 @@ ENV PORT=8080
 
 EXPOSE ${PORT}
 
-CMD bundle exec rake db:migrate && bundle exec whenever --update-crontab && bundle exec rails s -p ${PORT} --binding=0.0.0.0
+CMD bundle exec rake db:migrate:primary && bundle exec whenever --update-crontab && bundle exec rails s -p ${PORT} --binding=0.0.0.0
```
cwrw, line 71: I guess that's why it wasn't deploying 😅

config/database.yml
```diff
@@ -1,36 +1,70 @@
-default: &default
+
+shared_db_settings: &shared_db_settings
   adapter: postgresql
-  encoding: unicode
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  encoding: unicode
+
+default_primary: &default_primary
```
cwrw, line 7: just wanted to confirm this has been testing on all envs in here to ensure connections aren't lost?

app/models/migration/ecf/base_record.rb
ethax-ross: So its a bit of a difficult one, as if we omit the `connects_to` it will default to the `primary` NPQ DB for the ECF models. We want to do this in the review apps because there is no ECF DB spun up in review, but I figured for prod we can let the connection happen and its usage will be restricted elsewhere (the migration service will have a safety check that looks at the config to ensure its enabled).

We could override `readonly?` to prevent any accidental writes on environments where its disabled though?

https://api.rubyonrails.org/classes/ActiveRecord/Core.html#method-i-readonly-3F

db/ecf_schema.rb
ethax-ross: I think this will just need to be a manual task to begin with; I don't think anything will break if it becomes out of date (unless a migration isn't backwards compatible, but most of ours will be). If we find we get errors in NPQ reg deployments because of it we could revisit automating it? we could pull the latest version in from ECF when we build the docker image

config/database.yml
ethax-ross: I've tested on the review environment and migration environment (which has the same config as sandbox/staging/prod). I think the only way it could fall over is if there is an issue with the production credentials, but we can't really test that short of deploying to prod 😬

app/models/migration/ecf/base_record.rb
ethax-ross: I've made this change [here](https://github.com/DFE-Digital/npq-registration/pull/1185/commits/a94fb076b9bd348a8226d3c040a841c3f13cdba5)

app/models/migration/ecf/base_record.rb
```diff
@@ -0,0 +1,11 @@
+module Migration::Ecf
```
leandroalemao, line 1: Hi @ethax-ross what do you think about adding `inflect.acronym "ECF"` to the inflections file as we have in https://github.com/DFE-Digital/early-careers-framework/blob/main/config/initializers/inflections.rb#L25 so module name can be in capital letter? `Migration::ECF`

leandroalemao (changes requested):

app/models/migration/ecf/base_record.rb
ethax-ross: yeah I think Pete mentioned this too, happy to change it here 👍

ethax-ross: actually, its going to change a lot of existing code (the `Ecf` namespace is already in the app for other things) so lets leave that for a separate PR

cwrw (approved):
🌟

leandroalemao (approved):

PR #1189, opened 2024-02-29
Add skeleton Migrator service with dummy migration run

ethax-ross:
[Jira-2853](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2853)

### Context

We want to introduce the skeleton migration service that we can build on going forward. Initially, we want to simulate a migration so that we can more easily create a UI off the resulting `DataMigration` models.

- Add skeleton Migrator service with dummy migration run

Add the skeleton `Migrator` service, ensuring it can only run on environments where the data migration feature is enabled.

Run a dummy migration for `LeadProvider`, `Cohort` and `Statement`. The dummy migration will simulate processing up to 1,000 records per model and each will take less than a minute to complete. Failures are added at random (with a 1 in 50 change of occurring).

- Add DataMigration model

We will use this model to keep track of how a migration is progressing and the outcome/failure counts.

- Add total_count to DataMigration model

We want to have a clear indication of how many entities we need to migrate.

### DIFF
--- a/app/services/migration/migrator.rb
+++ b/app/services/migration/migrator.rb
@@ -1,0 +1,47 @@
+module Migration
+  class Migrator
+    class UnsupportedEnvironmentError < RuntimeError; end
+
+    MODELS_TO_MIGRATE = %i[lead_provider cohort statement].freeze
+
+    def migrate!
+      check_environment!
+      initialize_dummy_migration
+      run_dummy_migration
+    end
+
+  private
+
+    attr_accessor :data_migrations
+
+    def check_environment!
+      migration_enabled = Rails.application.config.npq_separation[:migration_enabled]
+
+      raise UnsupportedEnvironmentError, "The migration functionality is disabled for this environment" unless migration_enabled
+    end
+
+    def initialize_dummy_migration
+      self.data_migrations = MODELS_TO_MIGRATE.index_with { |model| Migration::DataMigration.create!(model:) }
+    end
+
+    def run_dummy_migration
+      MODELS_TO_MIGRATE.each { |model| simulate_model_migration(model) }
+    end
+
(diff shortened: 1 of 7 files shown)


db/migrate/20240228140211_create_data_migration.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDataMigration < ActiveRecord::Migration[7.1]
+  def change
+    create_table :data_migrations do |t|
+      t.string :model, null: false
+      t.integer :processed_count, null: false, default: 0
```
cwrw, line 5: do we need a total of rows to process, although we can use a total of processed vs failure I guess, do we trust those numbers fully?

app/services/migration/migrator.rb
```diff
@@ -0,0 +1,47 @@
...
+      initialize_dummy_migration
+      run_dummy_migration
+    end
+
+  private
+
+    attr_accessor :data_migrations
+
+    def check_environment!
+      migration_enabled = Rails.application.config.npq_separation[:migration_enabled]
```
cwrw, line 18: 👌

db/migrate/20240228140211_create_data_migration.rb
ethax-ross: I thought about this, but computing it adds complexity (especially for users, where we are only migrating a subset of the dataset) so I left it off for now. I think it may be worth adding if we find we want an more accurate % complete update in the UI, but I'm not sure if that's going to be useful yet so I left it off. We should be able to take `processed_count` as the total at the end of the migration (unless the migration itself fails but that would be evident elsewhere). The `failure_count` is a subset of `processed_count`.

PR #1193, opened 2024-03-01
Admin navigation components

peteryates:
The new admin section is currently hardcoded, this PR converts it into a pair of components. The components are from x-govuk prototype components ([primary](https://x-govuk.github.io/govuk-prototype-components/primary-navigation/), [secondary](https://x-govuk.github.io/govuk-prototype-components/sub-navigation/)).

* [x] Convert the primary navigation to a component
* [x] Convert the secondary navigation to a component
* [x] Add some additional pages (routes, controllers, indexes) so we can ensure the navigation is working

The components don't have a great deal of functionality but they'll provide a good place to add extra logic when we need it, for example showing extra areas to super users.

The one bit of functionality they do have is they compare the path of the current page to a prefix, and if it matches they'll highlight an entry to make it more obvious where you are within the admin console.

| Secondary nav 'section' selected | Secondary nav 'subsection' selected |
| ----------------| -----------------| 
| Screenshot from 2024-03-04 15-35-38 | Screenshot from 2024-03-04 15-35-32 |

## Reviewing

* keep in mind that lots of this is smoke and mirrors (to test nav components we need pages to navigate between, and controllers, and routes, etc)
* review commit-by-commit, it's quite sensibly organised


app/models/statement.rb
```diff
@@ -9,6 +9,9 @@ class Statement < ApplicationRecord
 
   validate :validate_max_statement_items_count
 
+  scope :unpaid, -> { where(state: %w[open payable]) }
+  scope :paid, -> { where(state: "paid") }
```
peteryates, line 13: These are just added to make the pages we're previewing show different data.

spec/factories/statements.rb
```diff
@@ -8,5 +8,13 @@
     marked_as_paid_at { nil }
     reconcile_amount { Faker::Number.decimal(l_digits: 3, r_digits: 2) }
     state { "open" }
+
+    trait(:paid) do
+      state { "paid" }
+      marked_as_paid_at { 1.week.ago }
+    end
```
peteryates, line 15: I assume paid statements should have a `marked_as_paid_at` in the past?

app/components/npq_separation/admin/secondary_navigation_component.rb
```diff
@@ -0,0 +1,107 @@
...
+        if matching_section.nil?
+          Rails.logger.error("No matching admin section for '#{current_path}'")
+
+          return
+        end
+
+        matching_section.name
+      end
+
+      def structure
```
peteryates, line 72: I hardcoded the structure here for now, we could write some extra specs that ensure the names/paths/prefixes are correct if we're happy with the approach.

db/seeds.rb
```diff
@@ -73,66 +73,33 @@ def seed_childcare_providers!
...
-    )
-  end
+  # a user with 4 applications
+  FactoryBot.create_list(
+    :application,
+    4,
+    :with_random_lead_provider_approval_status,
+    :with_random_participant_outcome_state,
+    :with_random_work_setting,
+    user: FactoryBot.create(:user, :with_random_name),
```
ethax-ross, line 96: This is creating 4 applications each with a unique user:

```
[5] pry(main)> FactoryBot.create_list(:application, 4, user: FactoryBot::create(:user)).map(&:user_id)
=> [8, 8, 8, 8]
```

app/controllers/npq_separation/admin/applications/all_controller.rb
```diff
@@ -0,0 +1,5 @@
+class NpqSeparation::Admin::Applications::AllController < NpqSeparation::AdminController
```
ethax-ross, line 1: Why not just `NpqSeparation::Admin::ApplicationsController`?

app/services/applications/find.rb
```diff
@@ -0,0 +1,7 @@
+module Applications
+  class Find
+    def all
```
ethax-ross, line 3: Worth having a spec for this class?

ethax-ross (commented):
Looks good to me, just a couple of questions/suggestions

db/seeds.rb
peteryates: Yeah this was intended to replace [the `multiple_app_user`](https://github.com/DFE-Digital/npq-registration/blob/main/db/seeds.rb#L93C3-L93C20) in the current seeds.

app/controllers/npq_separation/admin/applications/all_controller.rb
peteryates: I imagined there'd be several others alongside it and we might use the `NpqSeparation::Admin::ApplicationsController` for some kind of overview. All smoke and mirrors for now and likely to change.

peteryates: Flattened this to `NpqSeparation::Admin::ApplicationsController` for now, I suspect there'll be a few more flavours of applications controllers soon though.

db/seeds.rb
ethax-ross: Oh ignore me, I'm having a moment 🤦‍♂️

PR #1194, opened 2024-03-01
Add UI for ECF migration

ethax-ross:
[Jira-2854](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2854)

### Context

We want to a simple user interface available to super administrators where we can kick off a new migration, keep track of the progress and view the outcome.

- Add convenience methods to DataMigration

In order to render the state more easily in the UI we want to add some convenience methods to the `DataMigration` model for determining the state of the migration and the outcome.

- Add MigrationJob

Add job for kicking off a migration. Split out the preparation logic on the `Migrator` service so that we can create the `DataMigration` instances in-band and therefore render the state more easily in the UI (if we did this in the job itself it may not happen before the page reloads after starting a migration).

- Add MigrationHelper for rendering migration state in a view

We will want to render the migration state in a view for when a migration is in-progress and completed; this adds the relevant view helper methods.

- Add controller and views for migration

Add the controller actions and views necessary for a user to be able to kick off, view the progress and outcome of a migration. Only super admins will be able to run migrations.

Fixes a typo in the `sign_in.html.erb` error message.

Adds a config option for `rspec` to easily run tests against an in-memory cache and `rack_test` Capybara driver.

- DRY up feature specs using rack_test

- Update nodejs in tool-versions

### Guidance to review

Best reviewed by-commit.

| Start a migration              | Migration in progress | Migration complete |
| ---------------- | ------ | ---- |
|  Screenshot 2024-03-05 at 09 26 16      |   Screenshot 2024-03-05 at 09 26 28   | Screenshot 2024-03-05 at 09 26 41 |

### DIFF
--- a/app/helpers/migration_helper.rb
+++ b/app/helpers/migration_helper.rb
@@ -1,0 +1,50 @@
+module MigrationHelper
+  def migration_started_at(data_migrations)
+    # When a migration is first kicked off all data_migration records are briefly
+    # pending (with a `started_at` of `nil`) until a worker picks up the job.
+    # We use the current time as the start time in this case.
+    data_migrations.map(&:started_at).compact.min || Time.zone.now
+  end
+
+  def migration_completed_at(data_migrations)
+    data_migrations.map(&:completed_at).compact.max
+  end
+
+  def migration_duration_in_words(data_migrations)
+    duration_in_seconds = (migration_completed_at(data_migrations) - migration_started_at(data_migrations)).to_i
+    ActiveSupport::Duration.build(duration_in_seconds).inspect
+  end
+
+  def data_migration_status_tag(data_migration)
+    return govuk_tag(text: "Pending", colour: "grey") if data_migration.pending?
+    return govuk_tag(text: "In progress - #{data_migration.percentage_migrated}%", colour: "blue") if data_migration.in_progress?
+
+    govuk_tag(text: "Completed", colour: "green")
+  end
+
+  def data_migration_failure_count_tag(data_migration)
+    return if data_migration.failure_count.zero?
+
+    govuk_tag(text: number_with_delimiter(data_migration.failure_count), colour: "red")
+  end
+
+  def data_migration_total_count_tag(data_migration)
+    return unless data_migration.total_count&.positive?
+
+    govuk_tag(text: number_with_delimiter(data_migration.total_count), colour: "blue")
+  end
+
+  def data_migration_percentage_migrated_successfully_tag(data_migration)
+    percentage = data_migration.percentage_migrated_successfully
+
+    colour = if percentage < 80
+               "red"
+             elsif percentage < 100
+               "yellow"
+             else
+               "green"
+             end
+
+    govuk_tag(text: "#{percentage}%", colour:)
+  end
+end
--- a/app/views/npq_separation/migration/migrations/_completed_migration.html.erb
+++ b/app/views/npq_separation/migration/migrations/_completed_migration.html.erb
@@ -1,0 +1,29 @@
+
+<h2 class="govuk-heading-l">Completed migration</h2>
+
+<p class="govuk-body">
+  The latest migration was completed <%= tag.strong(time_ago_in_words(migration_completed_at(@data_migrations))) %> ago.<br>
+  The migration took <%= tag.strong(migration_duration_in_words(@data_migrations)) %> to complete.
+</p>
+
+<%= govuk_table do |table|
+  table.with_head do |head|
+    head.with_row do |row|
+      row.with_cell(text: "Model")
+      row.with_cell(text: "Number of records")
+      row.with_cell(text: "Number of failures")
+      row.with_cell(text: "Success rate")
+    end 
+  end 
+  
+  table.with_body do |body|
+    @data_migrations.each do |data_migration|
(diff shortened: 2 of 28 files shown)


spec/controllers/npq_separation/migration/migrations_controller_spec.rb
```diff
@@ -1,9 +1,56 @@
 require "rails_helper"
 
 RSpec.describe NpqSeparation::Migration::MigrationsController, type: :request do
```
peteryates, line 3: I think I caused confusion when adding `request` specs to `spec/controllers`. I've since moved some back to `spec/requests` - should they all be there?

app/views/npq_separation/migration/migrations/_completed_migration.html.erb
```diff
@@ -0,0 +1,29 @@
+
+<h2 class="govuk-heading-l">Completed migration</h2>
+
+<p class="govuk-body">
+  The latest migration was completed <strong><%= time_ago_in_words(migration_completed_at(@data_migrations)) %></strong> ago.<br>
+  The migration took <strong><%= migration_duration_in_words(@data_migrations) %></strong> to complete.
```
peteryates, line 6: Totally doesn't matter here but often less fiddly to write:

```suggestion
  The migration took <%= tag.strong(migration_duration_in_words(@data_migrations)) %> to complete.
```

spec/controllers/npq_separation/migration/migrations_controller_spec.rb
peteryates: For example I created a [sign in as admin](https://github.com/DFE-Digital/npq-registration/blob/main/spec/support/helpers/npq_separation_admin_login.rb#L2) helper which could easily be extended to take a `superadmin` arg (or split into a separate `#sign_in_as_superadmin`)

ethax-ross: > I think I caused confusion when adding `request` specs to `spec/controllers`. I've since moved some back to `spec/requests` - should they all be there?

Yeah I think that would make more sense. Probably worth doing all of them in a separate PR though. I'd also like to move `spec/lib/services` to `spec/services` to reflect the app structure as well.

ethax-ross: > For example I created a [sign in as admin](https://github.com/DFE-Digital/npq-registration/blob/main/spec/support/helpers/npq_separation_admin_login.rb#L2) helper which could easily be extended to take a `superadmin` arg (or split into a separate `#sign_in_as_superadmin`)

That's nicer than mocking `current_admin`, I'll update 👍

peteryates (approved):
Nice, everything looks super tidy 👌🏽

app/helpers/migration_helper.rb
```diff
@@ -0,0 +1,47 @@
+module MigrationHelper
+  def migration_started_at(data_migrations)
+    data_migrations.map(&:started_at).compact.min || Time.zone.now
```
cwrw, line 3: I was wondering why do we need to default to the current time? If the migration hasn't started wouldn't we want to always return nothing 🤔 ?

spec/features/admin_spec.rb
```diff
@@ -1,19 +1,17 @@
 require "rails_helper"
 
-RSpec.feature "admin", type: :feature do
+RSpec.feature "admin", type: :feature, rack_test_driver: true do
```
cwrw, line 3: 👍

config/routes.rb
```diff
@@ -167,7 +167,7 @@
     end
 
     namespace :migration, constraints: ->(_request) { Rails.application.config.npq_separation[:migration_enabled] } do
-      resources :migrations, only: %i[index]
+      resources :migrations, only: %i[index create]
```
cwrw, line 170: I was wondering if we should have a tab appear for super user admins as in `npq-separation/admin/migration/migrations` for this or leave out of admin?

cwrw (approved):
looks great 🙌 nicely done

app/helpers/migration_helper.rb
ethax-ross: Yeah good point, returning `nil` would make more sense here I'll update

config/routes.rb
ethax-ross: I would be inclined to leave it off in to avoid any accidental usage/clicking, but happy to add it in if you feel differently?

cwrw: makes sense, let's leave out 👍

app/helpers/migration_helper.rb
ethax-ross: Ah actually I realised why I did this; when you prepare a migration it creates all the `DataMigration` instances in-band (so we know a migration has started) but there's a short delay before the worker picks up the job where all the `DataMigration` models are pending (so none have a `started_at`). In this scenario I figured the best thing would be to display as if the migration has only just started by returning `Time.zone.now`. I've added a comment to that helper for clarity.

cwrw: ah makes sense, thanks 👍

cwrw (approved):

PR #1196, opened 2024-03-04
Add api serializer statements

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2864

Uses [Blueprinter gem](https://github.com/procore-oss/blueprinter?tab=readme-ov-file) to serialize the statement and statements [according to these specs](https://manage-training-for-early-career-teachers.education.gov.uk/api-reference/reference-v3.html#api-v3-statements-get-responses).

### DIFF
--- a/app/serializers/api/v3/statement_serializer.rb
+++ b/app/serializers/api/v3/statement_serializer.rb
@@ -1,24 +1,24 @@
 module API
   module V3
-    class AttributesSerializer < Blueprinter::Base
-      exclude :id
-
-      field(:month) { |s, _| Date::MONTHNAMES[s.month] }
-      field :year
-      field(:cohort) { |s, _| s.cohort.start_year }
-      field :deadline_date, name: :cut_off_date, datetime_format: "%Y-%m-%d"
-      field :payment_date, datetime_format: "%Y-%m-%d"
-      field(:paid) { |s, _| s.payment_date.present? }
-      field :created_at
-      field :updated_at
-    end
-
     class StatementSerializer < Blueprinter::Base
       identifier :id
       field(:type) { "statement" }
 
       field :attributes do |statement, _options|
         AttributesSerializer.render_as_hash(statement)
+      end
+
+      class AttributesSerializer < Blueprinter::Base
+        exclude :id
+
+        field(:month) { |s, _| Date::MONTHNAMES[s.month] }
+        field(:year) { |s, _| s.year.to_s }
+        field(:cohort) { |s, _| s.cohort.start_year.to_s }
+        field :deadline_date, name: :cut_off_date, datetime_format: "%Y-%m-%d"
+        field :payment_date, datetime_format: "%Y-%m-%d"
+        field(:paid) { |s, _| s.payment_date.present? }
+        field :created_at
+        field :updated_at
       end
     end
   end
--- a/spec/serializers/api/v3/statement_serializer_spec.rb
+++ b/spec/serializers/api/v3/statement_serializer_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe API::V3::StatementSerializer, type: :serializer do
     it "serializes the `cohort start year`" do
       cohort.start_year = 2025
 
-      expect(attributes["cohort"]).to eq(2025)
+      expect(attributes["cohort"]).to eq("2025")
     end
 
     it "serializes the `month`" do
@@ -41,7 +41,7 @@ RSpec.describe API::V3::StatementSerializer, type: :serializer do
     it "serializes the `year`" do
       statement.year = 2023
 
-      expect(attributes["year"]).to eq(2023)
+      expect(attributes["year"]).to eq("2023")
     end
 
     it "serializes the `deadline`" do


app/serializers/api/v3/statements_serializer.rb
```diff
@@ -0,0 +1,7 @@
+module API
+  module V3
+    class StatementsSerializer < Blueprinter::Base
```
cwrw, line 3: Im not sure we need this as we can pass a collection to a serialiser? see collections in https://github.com/procore-oss/blueprinter

Gemfile
```diff
@@ -4,6 +4,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby File.read(".ruby-version").chomp
 
 gem "activerecord-session_store"
+gem "blueprinter"
```
cwrw, line 7: could we set the JSON parster to `oj` as it's faster?

app/serializers/api/v3/statement_serializer.rb
```diff
@@ -0,0 +1,18 @@
...
+      field :month
+      field :year
+      field :cohort_id, name: :cohort
+      field :deadline_date, name: :cut_off_date, datetime_format: "%Y-%m-%d"
+      field :payment_date, datetime_format: "%Y-%m-%d"
+      field :created_at, datetime_format: ->(dt) { dt.iso8601 }
+      field :updated_at, datetime_format: ->(dt) { dt.iso8601 }
+
+      field(:paid) { |s, _| s.payment_date.present? }
+      field(:type) { "npq" }
```
cwrw, line 15: I think we will not be carrying this over to NPQ as it's an ECF specific field, could we remove it please?

Gemfile
cwrw: does the gem support nested objects well?

app/serializers/api/v3/statement_serializer.rb
```diff
@@ -0,0 +1,18 @@
+module API
+  module V3
+    class StatementSerializer < Blueprinter::Base
+      identifier :id
+
+      field :month
+      field :year
+      field :cohort_id, name: :cohort
```
mooktakim, line 8: `cohort` needs to return `cohort.start_year`, not `id`

app/serializers/api/v3/statement_serializer.rb
```diff
@@ -0,0 +1,18 @@
...
+      field :deadline_date, name: :cut_off_date, datetime_format: "%Y-%m-%d"
+      field :payment_date, datetime_format: "%Y-%m-%d"
+      field :created_at, datetime_format: ->(dt) { dt.iso8601 }
+      field :updated_at, datetime_format: ->(dt) { dt.iso8601 }
```
mooktakim, line 12: in ECF we use `created_at.rfc3339` instead of `iso8601`

cwrw (commented):
@pmanrubia not sure why I can't see the addressed commits on this pr 🤔 I get the following message: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

app/serializers/api/v3/statement_serializer.rb
peteryates, line 12: Probably makes sense to just set a [default format for BluePrinter](https://github.com/procore-oss/blueprinter?tab=readme-ov-file#global-config-setting---datetime) here and use `iso8601`, overriding if necessary.

Looking at these examples though it'll give us what we want without needing anything extra to be defined.

```ruby
DateTime.now.iso8601
=> "2024-03-07T10:09:26+00:00"

DateTime.now.strftime("%Y-%m-%d")
=> "2024-03-07"

Date.today.iso8691
=> "2024-03-07"
```

spec/serializers/api/v3/statement_serializer_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  let(:statement) { create(:statement, cohort:) }
+
+  it "serializes the `id`" do
+    response = subject.render_as_hash(statement)
+
+    expect(response[:id]).to eq(statement.id)
+  end
+
+  it "serializes the `cohort start year`" do
+    cohort.start_year = 2025
```
mooktakim, line 17: Its better if you pass these values in `let` when you create the objects

spec/serializers/api/v3/statement_serializer_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+require "rails_helper"
+
+RSpec.describe API::V3::StatementSerializer, type: :serializer do
+  let(:subject) { described_class }
+  let(:cohort) { create(:cohort) }
+  let(:statement) { create(:statement, cohort:) }
+
+  it "serializes the `id`" do
+    response = subject.render_as_hash(statement)
```
mooktakim, line 11: You can make this into a `let`, it'll dry up your code:
```ruby
let(:response) { subject.render_as_hash(statement) }
```

cwrw: interesting thoughts here, I know in ECF we've approached this using the DRY'ed up version, and I've definitely found it hard to maintain and follow. Personally, Im happy with having tests that are self contained and can be easily understood or maintained at the price of duplication. 

The problem with ECF and why we couldn't do that is the large setup required, maybe here we won't be hitting the same issues. The only extra suggestion from me is the risk of not persisting values on objects for test speed which sometimes has unexpected behaviour so something to watch out for. Happy to hear further thoughts as this is a personal preference ofcourse

mooktakim: @pmanrubia 
It'll just reduce the noise, the `response` line is not important for the test, its exactly the same for every test. As long as we keep the DRY within the same file I don't see it being an issue.

spec/serializers/api/v3/statement_serializer_spec.rb
mooktakim: You can add all the other attributes, its fine to initialize it with known values so you can later test if they exist.

app/serializers/api/v3/statement_serializer.rb
```diff
@@ -0,0 +1,17 @@
+module API
+  module V3
+    class StatementSerializer < Blueprinter::Base
```
mooktakim, line 3: The JSON response is not matching the existing [statements endpoint response from ECF](https://manage-training-for-early-career-teachers.education.gov.uk/api-reference/reference-v3.html#api-v3-statements-id-get-responses-examples):

```JSON
{
  "data": {
    "id": "cd3a12347-7308-4879-942a-c4a70ced400a",
    "type": "statement",
    "attributes": {
      "month": "May",
      "year": "2022",
      "type": "ecf",
      "cohort": "2021",
      "cut_off_date": "2022-04-30",
      "payment_date": "2022-05-25",
      "paid": true,
      "created_at": "2021-05-31T02:22:32.000Z",
      "updated_at": "2021-05-31T02:22:32.000Z"
    }
  }
}
```

spec/serializers/api/v3/statement_serializer_spec.rb
peteryates, line 1: Just a thought here, but because BluePrinter supports views might it be worth implementing the versions as views?

Then the bits that are common across the versions can be implemented once and the bits that vary can be nicely isolated:

```ruby
  view :v1 do
    field :some_v1_thing do |statement|
      # some_v1_specific_stuff
    end
  end

  view :v2 do
    field :some_v2_thing do |statement|
      # some_v2_specific_stuff
    end
  end
```

I guess this would be more complicated if the versions are _wildly_ different, but it would make comparing versions much easier.

cwrw: I think for me since statements is only in v3, maybe we can defer this to the next serialisers which implement endpoints across views, and keep it in mind to see how it looks like?

app/serializers/api/v3/statement_serializer.rb
mooktakim: Its close, but we still need to wrap it inside `data`, current response:
```JSON
{
  "type": "statement"
  "id": 1,
  "attributes": {
    "cohort": 2023,
    "created_at": "2024-03-07T18:22:51Z",
    "cut_off_date": "2024-04-05",
    "month": "July",
    "paid": false,
    "payment_date": null,
    "updated_at": "2024-03-07T18:22:51Z",
    "year": 2023
  },
}
```

spec/serializers/api/v3/statement_serializer_spec.rb
peteryates: Ah yeah, that's fair. Definitely worth a bit of investigation.

app/serializers/api/v3/statement_serializer.rb
```diff
@@ -1,17 +1,25 @@
 module API
   module V3
-    class StatementSerializer < Blueprinter::Base
-      identifier :id
+    class AttributesSerializer < Blueprinter::Base
```
peteryates, line 3: Not that it matters now while we have just one serialiser, but if multiple serialisers have an `AttributesSerializer` within would we end up with clashes?

app/serializers/api/v3/statement_serializer.rb
peteryates: Looking at the docs I think you apply the outer document when calling the renderer, so to wrap the above in `data:` you'd need to remember to do something like:

```ruby
API::V3::StatementSerializer.new(statement, root: "data")
```

Otherwise, you need a third serialiser (which is simple and sharable, but kind of ugly).

JBuilder is more flexible here by design.

mooktakim: Also, `cohort` and `year` are expected to be string.

mooktakim (approved):
All my comments have been answered, looks good to me 👍

cwrw (approved):
Thanks @pmanrubia for addressing all comments 🙌 
I'm taking it away to think about the nested attributes class in the statement serialiser to see how this would look like for others endpoints vs it in a new file. But that's very minor, and can leave for now

PR #1197, opened 2024-03-04
CPDNPQ-1697 Add form-level validation for trns to only be numerical

javier-npq:
### Context
Unvalidate trn user should not be bypass
### Ticket
https://dfedigital.atlassian.net.mcas.ms/browse/CPDNPQ-1697

This PR includes essential updates to the `qualification_teacher_check` questionnaire, specifically focusing on the `processed_trn` validations. The modifications encompass both the questionnaire itself and the associated spec file to ensure accurate and reliable validation processes.

**Note**
It's important to note that, due to certain constraints, we are refraining from updating model-level validations in this iteration. The reason behind this decision is that implementing model-level validations at this stage could lead to failures in cases where user information is being updated for individuals with unverified TRNs, such as "/", are present in the previously stored data within the database.

As a strategic measure, the current update enforces the acceptance criteria that users cannot submit with an unverified TRN specifically on the page "/registration/qualified-teacher-check." Once all unverified TRNs have been successfully updated, we can proceed to incorporate model-level validations seamlessly into the system.

This approach ensures a smoother transition and mitigates potential issues arising from the current state of user data, providing a robust foundation for future updates and improvements.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -376,7 +376,7 @@ en:
         course_start_date_hint: "You can also select this option if you have already started"
         maths_understanding_of_approach_hint: "For example, a reference from a professional who has supported your mastery professional development."
         lead_provider_id: These are the training providers who provide <b>%{course_name}</b>. Providers may have different entry requirements.
-        trn: Your TRN is between 6 and 7 characters long. It may begin with 'RP'
+        trn: Your TRN is between 6 and 7 characters long.
         full_name: Your full name exactly as it appears on the Teaching Regulation Agency records
         date_of_birth: For example, 31 3 1980
         national_insurance_number: This will help us match your details. It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’.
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -376,7 +376,7 @@ en:
         course_start_date_hint: "You can also select this option if you have already started"
         maths_understanding_of_approach_hint: "For example, a reference from a professional who has supported your mastery professional development."
         lead_provider_id: These are the training providers who provide <b>%{course_name}</b>. Providers may have different entry requirements.
-        trn: Your TRN is between 6 and 7 characters long. It may begin with 'RP'
+        trn: Your TRN is between 6 and 7 characters long.
```
peteryates, line 379: TRNs [should be 7 digits](https://www.gov.uk/guidance/teacher-reference-number-trn):

> A TRN is a 7-digit number that uniquely identifies you in the education sector in England

spec/lib/questionnaires/qualified_teacher_check_spec.rb
```diff
@@ -119,21 +119,21 @@ def stub_api_request(trn:, date_of_birth:, full_name:, nino:, response_code: 200
     end
 
     describe "#processed_trn" do
-      it "permits legacy style trns" do
+      it "Not permits legacy style trns" do
```
peteryates, line 122: "it doesn't permit legacy style TRNs" ?

config/locales/en.yml
javier-npq: Updated the validations

slawosz (approved):

PR #1199, opened 2024-03-04
[CPDLP-2863] TS1 - Lead provider API token authentication

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2863

* `APIToken` model added
* API routes set to default JSON
* Seed tokens for review apps

### DIFF
--- a/app/controllers/api/v1/get_an_identity/webhook_messages_controller.rb
+++ b/app/controllers/api/v1/get_an_identity/webhook_messages_controller.rb
@@ -1,7 +1,7 @@
 module API
   module V1
     module GetAnIdentity
-      class WebhookMessagesController < BaseController
+      class WebhookMessagesController < ApplicationController
         skip_before_action :verify_authenticity_token
 
         before_action :log_incoming_message
--- a/app/controllers/concerns/api/token_authenticatable.rb
+++ b/app/controllers/concerns/api/token_authenticatable.rb
@@ -1,0 +1,32 @@
+module API
+  module TokenAuthenticatable
+    extend ActiveSupport::Concern
+    include ActionController::HttpAuthentication::Token::ControllerMethods
+
+    included do
+      before_action :authenticate
+    end
+
+  private
+
+    def authenticate
+      authenticate_token || render_unauthorized
+    end
+
+    def authenticate_token
+      authenticate_with_http_token do |unhashed_token|
+        @current_api_token = APIToken.find_by_unhashed_token(unhashed_token).tap do |api_token|
+          api_token.update!(last_used_at: Time.zone.now) if api_token
+        end
+      end
+    end
+
+    def render_unauthorized
+      render json: { error: I18n.t(:unauthorized) }.to_json, status: :unauthorized
+    end
+
+    def current_lead_provider
+      @current_api_token&.owner
+    end
+  end
+end
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -1,0 +1,21 @@
+class APIToken < ApplicationRecord
+  belongs_to :lead_provider
+
+  validates :hashed_token, presence: true
+
+  def self.create_with_random_token!(**options)
+    unhashed_token, hashed_token = Devise.token_generator.generate(APIToken, :hashed_token)
+    create!(hashed_token:, **options)
+    unhashed_token
+  end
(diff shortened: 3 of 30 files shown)


Gemfile
```diff
@@ -59,6 +59,7 @@ group :development, :test do
   gem "parallel_tests"
   gem "pry-byebug"
   gem "pry-rails"
+  gem "rspec-default_http_header"
```
cwrw, line 62: I was wondering if we really need this dependency, and how complex it would be to set up the authorisation using normal requests or a helper?

app/controllers/api/v1/get_an_identity/webhook_messages_controller.rb
```diff
@@ -1,7 +1,7 @@
 module API
   module V1
     module GetAnIdentity
-      class WebhookMessagesController < BaseController
+      class WebhookMessagesController < ApplicationController
```
cwrw, line 4: not sure how this happened 😅

mooktakim: I changed this to `ApplicationController` just incase it messed up the existing behaviour.
Every controller inheriting `BaseController` will be authenticated only.

Gemfile
mooktakim: Without this gem we'll have to pass in `headers` to every `get`, `post` and `put`.
eg:
```ruby
get("/api/v3/statements", headers: { "Authorization" => "Bearer #{token}" })
```

db/seeds.rb
```diff
@@ -136,3 +136,18 @@ def seed_childcare_providers!
 end
 
 Setting.create!(course_start_date: 3.months.from_now)
+
+# APIToken for testing
+if Rails.env.in?(%w[development review])
+  {
+    "Ambition Institute" => "ambition-token",
```
cwrw, line 143: maybe we can include tokens for all NPQ lead providers? and if Capita isn't there we can remove?

spec/requests/api/v1/applications_controller_spec.rb
```diff
@@ -1,27 +0,0 @@
-require "rails_helper"
```
cwrw, line 1: having a think about those, I think I'd prefer to keep them but let's discuss in person

app/controllers/api/v1/get_an_identity/webhook_messages_controller.rb
mooktakim: Move `defaults: { format: "json" }` to only migration API endpoints

app/models/api_token.rb
```diff
@@ -0,0 +1,23 @@
+class APIToken < ApplicationRecord
+  has_paper_trail ignore: %i[updated_at last_used_at]
```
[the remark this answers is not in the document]
mooktakim: lets remove this for now, we add paper_trail at the end

spec/requests/api/v1/applications_controller_spec.rb
mooktakim: bring them back with authentication added

app/controllers/concerns/api/token_authenticatable.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+    end
+
+    def render_unauthorized
+      render json: { error: I18n.t(:unauthorized) }.to_json, status: :unauthorized
+    end
+
+    def current_lead_provider
+      @current_api_token&.owner
+    end
```
mooktakim, line 31: Instead of `current_user` we use `current_leader_provider` as we actually only return `LeadProvider`

app/controllers/concerns/api/token_authenticatable.rb
```diff
@@ -0,0 +1,33 @@
+module API
+  module TokenAuthenticatable
+    extend ActiveSupport::Concern
+    include ActionController::HttpAuthentication::Token::ControllerMethods
+
+    included do
+      before_action :authenticate
+      before_action :set_paper_trail_whodunnit
```
cwrw, line 8: ✂️

cwrw (approved):
Thanks Mook 🙌 just one tiny comment otherwise 👍

PR #1200, opened 2024-03-04
Performance comparison Blueprint & JBuilder

pmanrubia:

Tests performance of Blueprinter VS JBuilder with the following context:

- Rendering 3000 statements
- Caching disabled
- Adds nested Cohort

```bash
# Using Blueprint Gem
curl -H "Accept: application/json" -H  -  
0.01s user 0.01s system 2% cpu 0.559 total

# Using JBuilder
curl -H "Accept: application/json" -H  
-  0.00s user 0.01s system 2% cpu 0.514 total
```

Performance-wise both take the same amount of time; slightly better JBuilder after many repetitions but nothing meaningful. Advise is to use the one that fits better according to coding practices, testing and code maintenance. There is no big performance gap between both when rendering 3K elements.


cwrw (commented):
thank you for this @pmanrubia! Just for visibility we have chosen blueprinter as the serialiser for the API for now. We can also close this pr

PR #1202, opened 2024-03-04
[CPDLP-2844] Add dummy/test/seed data statements, cohorts & lead_providers

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2844

We'd like to test the new NPQ reg app so we need test data conforming with production data. 

Add dummy/test/seed data statements, cohorts & lead_providers.


spec/models/statement_spec.rb
```diff
@@ -10,6 +10,8 @@
   describe "validations" do
     it { is_expected.to validate_numericality_of(:month).is_in(1..12).only_integer }
     it { is_expected.to validate_numericality_of(:year).only_integer.is_in(2020..2050) }
+    it { is_expected.to allow_value(%w[true false]).for(:output_fee) }
```
ethax-ross, line 13: ```it { is_expected.to validate_inclusion_of(:output_fee).in_array([true, false]).allow_nil```

db/data/statements/statements.csv
```diff
@@ -0,0 +1,113 @@
+month,year,deadline_date,payment_date,output_fee,cohort
+
+1,2022,2021-12-25,2022-1-25,true,2021
```
ethax-ross, line 3: I'm not really familiar with these seeds on ECF, but why do we have CSVs here instead of just using the factories? I don't really get the grouping going on here either, I expect there's a pattern but its not very clear just from looking at it 🤔 I wonder if it would be easier to follow if it was just setup in code, that way we could at least add comments?

cwrw (commented):
thanks @leandroalemao for adding the seeds. I was wondering if we should think about avoiding using the services/csvs in the new world as we'd like to drive those changes instead by the admin interface. Maybe we could keep it very simple for now and load some base files with just a few statements for each cohort? Happy to chat in person also with @ethax-ross

db/data/statements/statements.csv
peteryates: Yeah I'd be eager to avoid CSVs if possible too. Preferably dynamic so they won't need updating every year or two too.

leandroalemao:
Hi @cwrw @ethax-ross @peteryates all comments have been addressed in https://github.com/DFE-Digital/npq-registration/pull/1202/commits/dae3d17cd98e26f9cffc1e5448522e76c69d9e62 tks for youre review

cwrw (commented):
@leandroalemao maybe we can keep the csvs in the seeds directory, unless they're used elsewhere, what do you think?

spec/factories/seeds/cohort_factory.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
cwrw, line 3: instead of adding new factories like ECF (which has a mix of old and new world factories), could we reuse the existing factories everywhere? eg https://github.com/DFE-Digital/npq-registration/blob/main/spec/factories/cohorts.rb#L1

Same for the rest, what do you think?

leandroalemao:
Hi @cwrw tks for your review.. comments addressed in [f83c5d8](https://github.com/DFE-Digital/npq-registration/pull/1202/commits/f83c5d83123e3ef51e04c4930b24b6e88c6956b3) tks

db/seeds/base/add_cohorts.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.create(:cohort, start_year: 2021)
+
+# Ensure Cohort.next is always created
```
cwrw, line 3: I think this is ECF specific :)

db/seeds.rb
```diff
@@ -4,121 +4,40 @@
...
+    files.push(
+      "add_users.rb",
+      "add_applications.rb",
+      "add_api_tokens.rb",
+    )
   end
 end
 
-# IDs have been hard coded to be the same across all envs
-seed_childcare_providers!
```
cwrw, line 49: could we maintain the order of the seeds created here above? apart from the new required seeds?

db/seeds/base/add_users.rb
```diff
@@ -0,0 +1,53 @@
...
+# Create admin user
+Admin.create!(
+  email: "superadmin@example.com",
+  full_name: "example super admin",
+  otp_hash: otp_testing_code,
+  otp_expires_at: "3000-01-01 00:00:00.000000000 +0000",
+  super_admin: true,
+)
+
+@single_app_user = User.create!(
```
cwrw, line 21: I wonder if we can avoid creating the users here and instead create them with the applications to encapsulate the logic in that file instead? I think this way we avoid linking across files also we avoid the risk of having 1 user attached to too many applications?

db/seeds/base/add_applications.rb
```diff
@@ -0,0 +1,26 @@
+school_settings = %w[a_school an_academy_trust a_16_to_19_educational_setting]
+
+Application.create!(
```
cwrw, line 3: I wonder if we could retain the original application logic here (see comment about users above) to ensure we create multiple single application user and multiple application users?

db/seeds.rb
```diff
@@ -4,121 +4,40 @@
 # Parallel Tests is seeding the database, so I am skipping this in the test environment
 return if Rails.env.test?
```
cwrw, line 5: I wonder if we can instead here only run this in `Rails.env.in?(%w[review development])` to avoid duplication across the other files with this logic, just worth double checking with NPQ team if they use this somewhere else we might have missed?

leandroalemao:
Hi @cwrw tks for your comments. All addressed in https://github.com/DFE-Digital/npq-registration/pull/1202/commits/6e0c26898ccf111fb276900d9fdeb05efd2f1c96 👍

cwrw (approved):
thanks for addressing all comments, look great 👍

PR #1203, opened 2024-03-05
[CPDLP-2866] TS1 - Statements endpoint - API base and error handling

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2866

* Added error handling for `BaseController`
* Created `API::Errors::Response` service class to render errors as JSON

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -1,5 +1,25 @@
 module API
-  class BaseController < ApplicationController
+  class BaseController < ActionController::API
     include API::TokenAuthenticatable
+
+    before_action :remove_charset
+    rescue_from ActionController::UnpermittedParameters, with: :unpermitted_parameter_response
+    rescue_from ActionController::BadRequest, with: :bad_request_response
+    rescue_from ArgumentError, with: :bad_request_response
+
+  private
+
+    def remove_charset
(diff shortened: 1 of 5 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -1,5 +1,54 @@
...
+
+    before_action :remove_charset
+    rescue_from ActionController::ParameterMissing, with: :missing_parameter_response
+    rescue_from ActionController::UnpermittedParameters, with: :unpermitted_parameter_response
+    rescue_from ActionController::BadRequest, with: :bad_request_response
+    rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
+    rescue_from ArgumentError, with: :bad_request_response
+    rescue_from ActiveRecord::RecordNotUnique, with: :bad_request_response
+    rescue_from ActiveRecord::RecordInvalid, with: :invalid_transition
+    rescue_from API::Errors::InvalidTransitionError, with: :invalid_transition
```
cwrw, line 13: I wonder if we should only keep the generic error messages and remove the more explicit ones till we get to the implementation, like outcomes etc?

mooktakim: I've removed all the custom error rescues 👍

PR #1207, opened 2024-03-06
Add statements query service

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2865

Adds a Service to retrieve statements filtered by: `cohorts`, `lead_provider` and `updated_since`

### DIFF
--- a/app/services/statements/statements_query.rb
+++ b/app/services/statements/statements_query.rb
@@ -15,7 +15,7 @@ module Statements
     end
 
     def statement(id)
-      Statement.find_by!(id: id, lead_provider: lead_provider)
+      Statement.find_by!(id:, lead_provider:)
     end
 
   private
--- a/spec/services/statements/statements_query_spec.rb
+++ b/spec/services/statements/statements_query_spec.rb
@@ -5,10 +5,18 @@ RSpec.describe Statements::StatementsQuery do
 
   describe "#statement" do
     it "returns a statement by id" do
-      statement = create(:statement, lead_provider: lead_provider)
+      statement = create(:statement, lead_provider:)
       query = Statements::StatementsQuery.new(lead_provider:)
 
       expect(query.statement(statement.id)).to eq(statement)
+    end
+
+    it "raises an error if the statement is not found" do
+      query = Statements::StatementsQuery.new(lead_provider:)
+
+      expect {
+        query.statement(999)
+      }.to raise_error(ActiveRecord::RecordNotFound)
     end
   end
 
@@ -29,8 +37,8 @@ RSpec.describe Statements::StatementsQuery do
         let(:updated_since) { 1.day.ago }
 
         it "filters by updated since" do
-          statement1 = create(:statement, lead_provider:, updated_at: 2.days.ago) 
-          statement2 = create(:statement, lead_provider:, updated_at: Time.zone.now) 
+          create(:statement, lead_provider:, updated_at: 2.days.ago)
+          statement2 = create(:statement, lead_provider:, updated_at: Time.zone.now)
 
           query = Statements::StatementsQuery.new(
             lead_provider:,


app/services/statements/statements_query.rb
```diff
@@ -0,0 +1,28 @@
...
+      statements = Statement.includes(:cohort).where(lead_provider:)
+      statements = statements.where(cohorts: { start_year: cohorts }) if cohorts.any?
+      statements = statements.where("updated_at >= ?", @updated_since) if @updated_since.present?
+
+      statements
+    end
+
+    def statement(id)
+      Statement.find_by!(id:, lead_provider:)
+    end
```
mooktakim, line 19: in ECF we have pattern of passing in the `id` with `params`:
```ruby
    def statement
      Statement.find_by!(id: params[:id], lead_provider:)
    end
```

app/services/statements/statements_query.rb
```diff
@@ -0,0 +1,28 @@
...
+    def initialize(lead_provider:, params: {})
+      @lead_provider = lead_provider
+      @cohorts = String(params[:cohort]).split(",")
+      @updated_since = params[:updated_since]
+    end
+
+    def statements
+      statements = Statement.includes(:cohort).where(lead_provider:)
+      statements = statements.where(cohorts: { start_year: cohorts }) if cohorts.any?
+      statements = statements.where("updated_at >= ?", @updated_since) if @updated_since.present?
```
mooktakim, line 12: `updated_since` is a `attr_reader` and can we use range instead:
```ruby
      statements = statements.where(updated_at:, updated_since..) if updated_since.present?
```

Also, we do some date parsing check with `updated_since` on ecf: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/concerns/filter_updated_since.rb, lets do that here.

app/services/statements/statements_query.rb
peteryates: You wouldn't have access to `params` in a service class through? 🤔

spec/services/statements/statements_query_spec.rb
```diff
@@ -0,0 +1,91 @@
+require "rails_helper"
+
+RSpec.describe Statements::StatementsQuery do
+  let(:lead_provider) { create(:lead_provider) }
+
+  describe "#statement" do
+    it "returns a statement by id" do
+      statement = create(:statement, lead_provider:)
+      query = Statements::StatementsQuery.new(lead_provider:)
```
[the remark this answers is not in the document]
peteryates: Agree on `subject` but I'm not sure `described_class` has any worthwhile advantages over being explicit with the class name. I reserve it only for shared examples where it needs to be dynamic.

peteryates (commented):
Looks like a good start.

FYI I added [`Statements::Find`](https://github.com/DFE-Digital/npq-registration/blob/main/app/services/statements/find.rb) in #1193  I think this query stuff and that should be integrated (or at least organised nicely)

app/services/statements/statements_query.rb
```diff
@@ -0,0 +1,28 @@
+module Statements
```
cwrw, line 1: do we need the statement directory if we're going with the name `statements_query`?

app/services/statements/statements_query.rb
cwrw: I was wondering if we could keep filtering to the filtering ticket? as we can do the whole slice then?

app/services/statements/statements_query.rb
cwrw: Maybe we can use explicit naming when initialising this service, rather than passing in params?

app/services/statements/statements_query.rb
```diff
@@ -0,0 +1,28 @@
...
+  class StatementsQuery
+    def initialize(lead_provider:, params: {})
```
peteryates, line 3: Should definitely be a list of keyword arguments here, it's self-documenting. To know which params are allowed would mean reading the code.

peteryates: Also, it might be worth thinking about how we build/structure these queries, i.e., do we:

* allow loads of params/keyword arguments at initialization and just have a single method called `#statements` that returns all i.e., `Statements::StatementsQuery.new(lead_provider:, cohorts:, since: some_date)`
* make it chainable so it's something more like `Statements::StatementsQuery.new(lead_provider:).in_cohorts(cohorts).since(some_date).all`

Ultimately you can build the same queries with both, but I suspect the second implementation will be easier o the eye.

PR #1209, opened 2024-03-07
[CPDLP-2855] Add logic to migrate cohorts and statements

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2855

We want to migrate cohorts and statements from ECF into NPQ reg. These models are grouped together as they don’t currently exist in NPQ reg (so its all new data coming across/nothing to clash into).

Add logic to migrate cohorts and statements.

### **Note: Cohort 2020 not migrated due new validation in place (start_year >= 2021)**

image

### DIFF
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -3,7 +3,7 @@ class Statement < ApplicationRecord
   belongs_to :lead_provider
   has_many :statement_items
 
-  validates :output_fee, presence: true
+  validates :output_fee, inclusion: [true, false]
   validates :month, numericality: { in: 1..12, only_integer: true }
   validates :year, numericality: { in: 2020..2050, only_integer: true }
 
--- a/app/services/migration/migrator.rb
+++ b/app/services/migration/migrator.rb
@@ -6,12 +6,12 @@ module Migration
     class MigrationAlreadyPreparedError < RuntimeError; end
 
     class << self
+      MODELS_TO_BE_MIGRATED = %i[lead_provider cohort statement].freeze
+
       def prepare_for_migration
         raise MigrationAlreadyPreparedError, "The migration has already been prepared" if DataMigration.exists?
 
-        DataMigration.create!(model: :lead_provider)
-        DataMigration.create!(model: :cohort)
-        DataMigration.create!(model: :statement)
+        MODELS_TO_BE_MIGRATED.each { |model| DataMigration.create!(model:) }
       end
     end
 
@@ -45,19 +45,23 @@ module Migration
       migrate_statements
     end
 
-    def migrate_lead_providers
-      data_migration = DataMigration.find_by(model: :lead_provider)
+    def migrate(items, model)
+      data_migration = DataMigration.find_by(model:)
+      data_migration.update!(started_at: Time.zone.now, total_count: items.count)
 
-      data_migration.update!(started_at: Time.zone.now, total_count: ecf_npq_lead_providers.count)
-
-      ecf_npq_lead_providers.find_each do |ecf_npq_lead_provider|
+      items.each do |item|
         data_migration.increment!(:processed_count)
-        npq_lead_provider = LeadProvider.find_by(ecf_id: ecf_npq_lead_provider.id)
-
-        data_migration.increment!(:failure_count) unless npq_lead_provider
+        result = yield(item)
+        data_migration.increment!(:failure_count) unless result
       end
 
       data_migration.update!(completed_at: Time.zone.now)
+    end
+
+    def migrate_lead_providers
+      migrate(ecf_npq_lead_providers, :lead_provider) do |ecf_npq_lead_provider|
+        LeadProvider.find_by(ecf_id: ecf_npq_lead_provider.id)
+      end
     end
 
     def ecf_npq_lead_providers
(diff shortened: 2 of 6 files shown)


app/services/migration/migrator.rb
```diff
@@ -10,6 +10,8 @@ def prepare_for_migration
         raise MigrationAlreadyPreparedError, "The migration has already been prepared" if DataMigration.exists?
 
         DataMigration.create!(model: :lead_provider)
```
ethax-ross, line 12: We could do:

```
%i[lead_provider cohort statement].each { |model| DataMigration.create!(model) }
```

app/services/migration/migrator.rb
```diff
@@ -59,5 +63,61 @@ def migrate_lead_providers
...
     end
+
+    def ecf_cohorts
+      @ecf_cohorts ||= Ecf::Cohort.all
+    end
+
+    def migrate_cohorts
+      data_migration = DataMigration.find_by(model: :cohort)
+
+      data_migration.update!(started_at: Time.zone.now, total_count: ecf_cohorts.count)
```
ethax-ross, line 74: I feel like we could extract a lot of this to another method that accepts a block now its being reused, along the lines of:

```
def migrate(items, model)
  data_migration = DataMigration.find_by(model: :cohort)
  data_migration.update!(started_at: Time.zone.now, total_count: items.count)

  items.each do |item|  
    yield(item, data_migration)
    data_migration.increment!(:processed_count)
  end

  data_migration.update!(completed_at: Time.zone.now)
end
```

Then this method becomes something like:

```
def migrate_cohorts
  migrate(items, model) do
      cohort = Cohort.find_or_initialize_by(start_year: ecf_cohort.start_year)
      result = cohort.update(registration_start_date: ecf_cohort.registration_start_date)
      data_migration.increment!(:failure_count) unless result
  end
end
```

app/services/migration/migrator.rb
```diff
@@ -40,21 +39,82 @@ def check_environment!
...
+    end
+
+    def ecf_npq_lead_providers
+      @ecf_npq_lead_providers ||= Ecf::NpqLeadProvider.all
     end
 
-    def simulate_model_migration(data_migration)
-      data_migration.update!(started_at: Time.zone.now, total_count: rand(1...1000))
+    def ecf_cohorts
+      @ecf_cohorts ||= Ecf::Cohort.all
```
cwrw, line 68: chatted in person: might be worth ignoring 2020 here

app/models/statement.rb
```diff
@@ -3,7 +3,7 @@ class Statement < ApplicationRecord
   belongs_to :lead_provider
   has_many :statement_items
 
-  validates :output_fee, presence: true
+  validates :output_fee, inclusion: [true, false]
```
cwrw, line 6: should we add a test for this?

app/services/migration/migrator.rb
```diff
@@ -39,25 +41,69 @@ def check_environment!
...
     end
+
+    def ecf_cohorts
+      @ecf_cohorts ||= Ecf::Cohort.where(start_year: 2021..)
+    end
+
+    def migrate_cohorts
+      migrate(ecf_cohorts, :cohort) do |ecf_cohort|
+        cohort = Cohort.find_or_initialize_by(start_year: ecf_cohort.start_year)
+        result = cohort.update(registration_start_date: ecf_cohort.registration_start_date)
```
cwrw, line 78:
```suggestion
        result = cohort.update(registration_start_date: ecf_cohort.npq_registration_start_date)
```

cwrw: some of those are `nil`, do we want to check the logic in ECF around how those defaulted before 2023 and maybe then populating with `registration_start_date`?

app/services/migration/migrator.rb
```diff
@@ -39,25 +41,69 @@ def check_environment!
...
+    def migrate_statements
+      migrate(ecf_statements, :statement) do |ecf_statement|
+        statement = Statement.find_or_initialize_by(month: Date::MONTHNAMES.find_index(ecf_statement.name.split[0]), year: ecf_statement.name.split[1])
+
+        result = statement.update(
+          deadline_date: ecf_statement.deadline_date,
+          payment_date: ecf_statement.payment_date,
+          output_fee: ecf_statement.output_fee,
+          cohort: Cohort.find_by(start_year: ecf_cohorts.find_by(id: ecf_statement.cohort_id).start_year),
+          lead_provider: LeadProvider.find_by(ecf_id: ecf_npq_lead_providers.find_by(cpd_lead_provider_id: ecf_statement.cpd_lead_provider_id).id),
```
cwrw, line 96:
```suggestion
          lead_provider: LeadProvider.find_by(ecf_id: ecf_statement.cpd_lead_provider.npq_lead_provider.id),
```
since we import both tables maybe this is simpler?

app/services/migration/migrator.rb
```diff
@@ -39,25 +41,69 @@ def check_environment!
...
+
+    def ecf_cohorts
+      @ecf_cohorts ||= Ecf::Cohort.where(start_year: 2021..)
+    end
+
+    def migrate_cohorts
+      migrate(ecf_cohorts, :cohort) do |ecf_cohort|
+        cohort = Cohort.find_or_initialize_by(start_year: ecf_cohort.start_year)
+        result = cohort.update(registration_start_date: ecf_cohort.registration_start_date)
+        result
```
ethax-ross, line 79: `result` is redundant here (and on the statements method) we could just return the `cohort.update` result

app/models/statement.rb
leandroalemao: Hi @cwrw Tests are in the other PR but I've added the same in this one as well.. tkss

app/services/migration/migrator.rb
leandroalemao: HI @cwrw I can only see `npq_registration_start_date` being used to fetch the current `active_npq_registration_cohort` when building a npq application in ECF app.. not being used in the "past" so maybe that's why we have nil `npq_registration_start_date` value for past cohorts..

app/services/migration/migrator.rb
leandroalemao: Hi @ethax-ross Yeah I know but had to add `result` to get around this [rubocop warning ](https://www.rubydoc.info/gems/rubocop/0.44.1/RuboCop/Cop/Rails/SaveBang)

cwrw (approved):
thanks @leandroalemao !

ethax-ross (approved):
Looks good! 👍 I think when we get on to the next table to migrate it would be worth thinking about how we could start to pull out some of the model-specific logic in here to their own classes

PR #1213, opened 2024-03-11
Statements admin pages

peteryates:
This PR starts adding some admin pages to the finance section around statements.

It is light on tests because these pages are still being actively designed and everything's likely to change.

| Statements overview | One Lead provider's statements| Individual statement view |
| ----------- | -------------- | ------------|
| Screenshot from 2024-03-11 16-19-12 | Screenshot from 2024-03-11 16-19-22 | Screenshot from 2024-03-11 16-19-29 |

## Thoughts

* Originally I was a fan of the side nav listing the lead providers but now it seems a bit busy and there's a chance of it overlapping with the actual 'Lead providers' bit of the primary nav
* We need to have a think about 'the C word' (cohorts) and what they mean to us, how/where we use them, etc. Here I opted to show them "2023/4" because that's what we do elsewhere, but that's encapsulated in a single helper so it's easy to change if we need to
* the tables look a bit 'link heavy' and all the blue text is distracting
* we'll need breadcrumbs for delving into the statements but this is complicated by the fact there are two 'routes' in, via the main statements list and via the lead provider


peteryates:
I've changed this a bit, I think having a second 'Lead providers' section in Finance is going to cause confusion so I promoted it to the primary one for now. This makes the nav cleaner and removes confusion from breadcrumbs (which will be added soon).

| Description | Screenshot |
| ---------- | --------------|
| The [text coloured](https://govuk-components.netlify.app/helpers/link/#other-link-styles) link style has been used to make the table rows a bit less overwhelming |Screenshot from 2024-03-11 17-47-20 |
| Lead providers index moved up to primary nav | Screenshot from 2024-03-11 17-51-52 |

spec/factories/statements.rb
```diff
@@ -16,5 +16,9 @@
 
     trait(:open) { state { "open" } }
     trait(:payable) { state { "payable" } }
+
+    trait(:with_existing_lead_provider) do
```
cwrw, line 20: 👍

peteryates:
I have no idea why the suite fails with parallel specs but works fine when run normally. Removing them in #1216 and will rebase.

cwrw (approved):
looks great! I know you mentioned it's intentionally light on testing, but only suggestion would be to add specs for the find services even if very light, otherwise all good 👍 and starting to take excellent shape

PR #1215, opened 2024-03-12
[CPDLP-2859] TS1 - Statements endpoint - happy path (show/index)

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2859

* `API::V3::StatementsController` updated with `index` and `show` actions
* Tests added

### DIFF
--- a/app/controllers/api/v3/statements_controller.rb
+++ b/app/controllers/api/v3/statements_controller.rb
@@ -2,11 +2,11 @@ module API
   module V3
     class StatementsController < BaseController
       def index
-        render json: StatementSerializer.render(statements_query.statements, root: "data")
+        render json: to_json(statements_query.statements)
       end
 
       def show
-        render json: StatementSerializer.render(statements_query.statement(id: statement_params[:id]), root: "data")
+        render json: to_json(statements_query.statement(id: statement_params[:id]))
       end
 
     private
@@ -19,6 +19,10 @@ module API
 
       def statement_params
         params.permit(:id)
+      end
+
+      def to_json(obj)
+        StatementSerializer.render(obj, root: "data")
       end
     end
   end
--- a/app/controllers/api/v3/statements_controller.rb
+++ b/app/controllers/api/v3/statements_controller.rb
@@ -1,8 +1,25 @@
 module API
   module V3
     class StatementsController < BaseController
-      def show = head(:method_not_allowed)
-      def index = head(:ok)
+      def index
+        render json: StatementSerializer.render(statements_query.statements, root: "data")
+      end
+
+      def show
+        render json: StatementSerializer.render(statements_query.statement(id: statement_params[:id]), root: "data")
+      end
+
+    private
+
+      def statements_query
+        ::Statements::Query.new(
+          lead_provider: current_lead_provider,
+        )
+      end
+
+      def statement_params
+        params.permit(:id)
+      end
     end
   end
 end
--- a/spec/requests/api/v3/statements_spec.rb
+++ b/spec/requests/api/v3/statements_spec.rb
@@ -1,13 +1,36 @@
 require "rails_helper"
 
 RSpec.describe "Statements endpoint", type: "request" do
+  let(:current_lead_provider) { create(:lead_provider) }
+
   describe "GET /api/v3/statements" do
     context "when authorized" do
-      it "returns 200 - ok" do
-        api_get("/api/v3/statements")
+      context "when 2 statements exist for current_lead_provider" do
+        let!(:statement1) { create(:statement, lead_provider: current_lead_provider) }
+        let!(:statement2) { create(:statement, lead_provider: current_lead_provider) }
 
-        expect(response.status).to eq 200
-        expect(response.headers["Content-Type"]).to include("application/json")
+        before do
+          create(:statement, lead_provider: create(:lead_provider, name: "Another lead provider"))
+        end
+
+        it "returns 2 statements" do
+          api_get("/api/v3/statements")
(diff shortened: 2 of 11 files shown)


app/controllers/api/v3/statements_controller.rb
```diff
@@ -1,8 +1,21 @@
...
   module V3
     class StatementsController < BaseController
-      def show = head(:method_not_allowed)
-      def index = head(:ok)
+      def index
+        render json: StatementSerializer.render(statements_query.statements, root: "data")
+      end
+
+      def show
+        render json: StatementSerializer.render(statements_query.statement(id: params[:id]), root: "data")
```
peteryates, line 9: Having to add `root: "data"` everywhere isn't the prettiest 😞 

[Doesn't look like there's any way around it either](https://github.com/procore-oss/blueprinter/issues/253).

spec/requests/api/v3/statements_spec.rb
```diff
@@ -1,13 +1,36 @@
...
-        expect(response.status).to eq 200
-        expect(response.headers["Content-Type"]).to include("application/json")
+        it "returns 2 statements" do
+          api_get("/api/v3/statements")
+
+          expect(response.status).to eq 200
+          expect(response.content_type).to eql("application/json")
+          expect(parsed_response["data"].size).to eq(2)
+          expect(parsed_response["data"][0]["id"]).to eq(statement1.id)
+          expect(parsed_response["data"][1]["id"]).to eq(statement2.id)
```
peteryates, line 23: I'm not sure we should be testing the actual response here - shouldn't we trust our query object and serialiser? Otherwise we end up with request specs overlapping with both serialiser and query object ones.

The controller's responsibility here is just to respond to a request appropriately.

app/controllers/api/v3/statements_controller.rb
cwrw: Maybe we can add a wrapper for this method in the base controller, and pass in the serialiser/params to avoid calling this everywhere?

Agree though its not great, and as @mooktakim mentioned in another conversation keep our options open to using other serialisers if we feel it might be a problem

cwrw: could we also add strong params, something similar to:
```
def statement_params
  params.permit(:id)
end
```

spec/requests/api/v3/statements_spec.rb
mooktakim: I think its important to make sure the result is as we expect. In this case, we want to make sure the correct statements are returned, so only checked for the `id`'s. I shouldn't check all values as that's tested in the serializer unit tests.

peteryates: We end up in a situation where changes to the serializers or query objects mean we need to update API specs though

app/controllers/api/v3/statements_controller.rb
peteryates: Also if we stick with BluePrinter and extend the query objects so they're modular and chainable, we could put th serialization there make this controller a little less involved, something more like:

```ruby
render(json: Statements::Query.new.for_lead_provider(current_lead_provider).serialize(:v3))
```

mooktakim: Update code with strong params 👍

mooktakim: `root: "data"` definitely not nice. I'm actually leaning towards going back to `jsonapi-serializer` 😂

spec/requests/api/v3/statements_spec.rb
mooktakim: We have to check whether we return the correct objects at least, otherwise its not really end to end.

cwrw: I was having a think about this, we need to exercise some form of reliable end to end testing, without going to deep into retesting the unit tests covered in serialisers/queries. I think checking size/ids is a good balance. If we stick to the following structure:
- Unit tests for serialisers
- Unit tests for queries and all edge cases
- Tests for ensuring schema is met for swagger
- Request specs which check statuses are ok, size and ids are ok and are very light happy path scenarios as in this PR it would be enough without delving into the fields returned and should cover us, otherwise I feel like we'll miss something if it does break

app/controllers/api/v3/statements_controller.rb
mooktakim: I tried to DRY up the code a bit with `to_json` method

app/controllers/api/v3/statements_controller.rb
```diff
@@ -1,8 +1,29 @@
...
+
+    private
+
+      def statements_query
+        ::Statements::Query.new(
+          lead_provider: current_lead_provider,
+        )
+      end
+
+      def statement_params
```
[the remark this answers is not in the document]
cwrw: Although no mass assignments is happening, I think it's good practice to use and we'll be adding more params soon like filtering etc. so its safe so start early

cwrw (commented):
looks great @mooktakim! only last comment would be around the order of params appearing in the API responses, testing locally they are ordered alphabetically not sure if you're also seeing this? any way we can preserve the expected order instead?

mooktakim:
> looks great @mooktakim! only last comment would be around the order of params appearing in the API responses, testing locally they are ordered alphabetically not sure if you're also seeing this? any way we can preserve the expected order instead?

I was able to add attributes ordering based on definition 👍

cwrw (approved):
thanks mook 👍

PR #1115, opened 2024-01-22
Move old features to new combined format

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1650

Fixes #1093 

We have journey features in two places:

the old ones in features/happy_journeys/features/sad_journeys - these are then divided again into with/without JavaScript

the new ones in fetures/journeys - these can run in JavaScript mode and no-JavaScript mode

PR #1216, opened 2024-03-12
Remove `parallel_tests` gem and all config

peteryates:
The test suite is now fast enough (since #707, #1115) to run sequentially. This simplifies our setup, makes the test output clearer and removes a dependency.

The entire run is 1 minute slower, from 6 mins to 7 mins.

We could make the whole thing *a lot faster* by testing the JavaScript pages on their own rather than as part of the end-to-end runs.

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -68,7 +68,7 @@ jobs:
         run: mkdir -p tmp/capybara
 
       - name: Set up test database
-        run: bin/rails parallel:setup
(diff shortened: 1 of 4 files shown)


cwrw (approved):

cwrw:
ah sorry was looking at the original implementation, there's a bit more cleanup to do see https://github.com/DFE-Digital/npq-registration/pull/767/files around `TEST_ENV_NUMBER`

PR #1221, opened 2024-03-14
Add filtering to Statements endpoint

pmanrubia:
Closes: https://github.com/DFE-Digital/npq-registration/pull/1207

### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2861

Adds filtering by `cohort` and `updated_at`. 

It modifies the Query that currently retrieve statements, by making it `chainnable`.

Related to: https://github.com/DFE-Digital/npq-registration/pull/1207

### DIFF
--- a/app/controllers/api/v3/statements_controller.rb
+++ b/app/controllers/api/v3/statements_controller.rb
@@ -15,7 +15,7 @@ module API
         Statements::Query.new
                          .belonging_to(lead_provider: current_lead_provider)
                          .by_cohorts(cohorts)
-                         .since(updated_since)
+                         .updated_since(updated_since)
       end
 
       def statement_params
--- a/app/services/statements/query.rb
+++ b/app/services/statements/query.rb
@@ -20,7 +20,7 @@ module Statements
       self
     end
 
-    def since(date)
+    def updated_since(date)
       if date.present?
         @scope = @scope.where(updated_at: date..)
       end
--- a/spec/services/statements/query_spec.rb
+++ b/spec/services/statements/query_spec.rb
@@ -67,7 +67,7 @@ RSpec.describe Statements::Query do
 
           query = Statements::Query.new
                                    .belonging_to(lead_provider:)
-                                   .since(updated_since)
+                                   .updated_since(updated_since)
 
           expect(query.statements).to eq([statement2])
         end


app/services/statements/query.rb
```diff
@@ -1,21 +1,39 @@
...
+      if start_years.any?
+        @scope = @scope.where(cohort: { start_year: start_years })
+      end
+
+      self
+    end
+
+    def since(date)
+      if date.present?
+        @scope = @scope.where("updated_at >= ?", date)
```
peteryates, line 25:
```suggestion
        @scope = @scope.where(updated_at: date..)
```

app/controllers/api/v3/statements_controller.rb
```diff
@@ -1,8 +1,38 @@
...
+      end
+
+      def show
+        render json: to_json(statements_query.statement(id: statement_params[:id]))
+      end
+
+    private
+
+      def statements_query
+        ::Statements::Query.new
```
peteryates, line 15: Not sure we need the `::` prefix here 🤔

spec/requests/api/v3/statements_spec.rb
```diff
@@ -1,13 +1,66 @@
...
-        expect(response.headers["Content-Type"]).to include("application/json")
+          expect(response.status).to eq 200
+          expect(response.content_type).to eql("application/json")
+          expect(parsed_response["data"].size).to eq(2)
+          expect(parsed_response["data"][0]["id"]).to eq(statement1.id)
+          expect(parsed_response["data"][1]["id"]).to eq(statement2.id)
+        end
+      end
+
+      context "when not statements exist" do
```
peteryates, line 27:
```suggestion
      context "when no statements exist" do
```

spec/requests/api/v3/statements_spec.rb
```diff
@@ -1,13 +1,66 @@
...
+          it "returns statements updated since the specified date" do
+            create(:statement, lead_provider: current_lead_provider, updated_at: 6.hours.ago)
+            create(:statement, lead_provider: current_lead_provider, updated_at: 3.hours.ago)
+            create(:statement, lead_provider: current_lead_provider, updated_at: 1.hour.ago)
+
+            api_get("/api/v3/statements", params: { filter: { updated_since: 4.hours.ago } })
+
+            expect(parsed_response["data"].size).to eq(2)
+          end
+        end
```
peteryates, line 63: Like with #1215, I think this is going to cause overlap down the line.

The responsibility of the controller here is to receive a request with some params, make sure the params are sound and pass them into the query service.

In my opinion that's all we should be testing here.

app/services/statements/query.rb
```diff
@@ -1,19 +1,39 @@
 module Statements
   class Query
-    def initialize(lead_provider:)
-      @lead_provider = lead_provider
+    def initialize
+      @scope = Statement.all
+    end
+
+    def by_lead_provider(lead_provider)
```
peteryates, line 7: `#by_lead_provider` sounds like they'll be grouped by lead provider.

Maybe `#belonging_to_lead_provider`, or even `#belonging_to(lead_provider:)`?

app/services/statements/query.rb
```diff
@@ -1,19 +1,39 @@
...
+      @scope = @scope
+                 .includes(:cohort)
+                 .where(lead_provider:)
+
+      self
+    end
+
+    def by_cohorts(*start_years)
```
peteryates, line 15: `#with_cohort_start_years`? 🤔

peteryates: Yeah - the `by_` makes it sound like it's 'grouped by' them rather than limited to them. Naming is always the hardest bit!

app/services/statements/query.rb
```diff
@@ -1,21 +1,39 @@
...
+        @scope = @scope.where(cohort: { start_year: start_years })
+      end
+
+      self
+    end
+
+    def since(date)
+      if date.present?
+        @scope = @scope.where(updated_at: date..)
+      end
```
peteryates, line 26: I guess this is guarding against `nil` dates, but I'm not sure it's necessary, it just leaves the scope open.

```ruby
Statement.where(created_at: nil..).to_sql

#=> "SELECT \"statements\".* FROM \"statements\" WHERE 1=1"
```

Also `since` isn't the clearest name - i.e., is it created since or updated since?

spec/requests/api/v3/statements_spec.rb
peteryates: > Also, if we don't test some filtering, how can we be sure that we are even plugged to the right service with the right parameters?

cwrw: I agree with @pmanrubia request specs exercise the full stack and can drive full integration testing. Keeping them simple and testing the happy path seems fair to me, as long as we don't go too far with them.

app/controllers/api/v3/statements_controller.rb
```diff
@@ -12,13 +12,22 @@ def show
...
+                         .since(updated_since)
       end
 
       def statement_params
-        params.permit(:id)
+        params.permit(:id, :cohort, filter: %i[cohort updated_since])
+      end
+
+      def cohorts
+        statement_params.dig(:filter, :cohort)&.split(",")
```
cwrw, line 26: maybe we can pass the filter as is and attempt the `&.split(",")` in the service to avoid having to test this at the request spec level but we can at the query service level?

app/services/statements/query.rb
cwrw: preference is to have lead provider in the method name, as to avoid someone later adding more to this method to keep it explicit

app/services/statements/query.rb
cwrw: `with_cohorts` maybe?

app/services/statements/query.rb
cwrw: +1 for method name being `updated_since`

cwrw: in ECF, we have some concerns around rescuing bad data being added to filters, see this [concern](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/concerns/filter_updated_since.rb). Could we add something similar for protecting against bad data and returning relevant errors in the [controller](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/controllers/api/api_controller.rb#L4)?

app/services/statements/query.rb
```diff
@@ -1,21 +1,39 @@
 module Statements
   class Query
-    def initialize(lead_provider:)
-      @lead_provider = lead_provider
+    def initialize
+      @scope = Statement.all
+    end
+
+    def belonging_to(lead_provider:)
+      @scope = @scope
```
cwrw, line 8: ah something I missed in the previous pr, could we add the following to the scope `where(output_fee: true)` as that's required for the statements returned in the API? maybe as a different method?

cwrw: also could we maintain the order also missed in the previous pr `scope.order(payment_date: :asc)`?

spec/requests/api/v3/statements_spec.rb
```diff
@@ -24,14 +24,42 @@
...
+          let(:cohort_2025) { create(:cohort, start_year: 2025) }
+
+          it "returns statements for the specified cohort" do
+            create(:statement, lead_provider: current_lead_provider, cohort: cohort_2023)
+            create(:statement, lead_provider: current_lead_provider, cohort: cohort_2024)
+            create(:statement, lead_provider: current_lead_provider, cohort: cohort_2025)
+
+            api_get("/api/v3/statements", params: { filter: { cohort: "2023,2024" } })
+
+            expect(parsed_response["data"].size).to eq(2)
```
mooktakim, line 49: We should still check here if we've returned the correct two statements, as there's a possibility of returning the wrong statements.

PR #1222, opened 2024-03-14
Make admin navigation components reusable

peteryates:
### Context

This change makes the new navigation components reusable outside of the admin context.

Within the component class there are classes for each variant of navigation we need, and it's easy to add new ones. Each is self-contained and holds a set of nodes or sections which are injected into the component via the `structure:` keyword argument.

### Review suggestions

I'd recommend going through commit-by-commit, they're well organised.

It's worth disabling whitespace for the latest commit too, it's 99% indentation.

### Changes

- **Remove lambdas from subnav**
- **Disable some more spurious Rubocop rules**
- **Move config out of SubNavigationComponent**
- **Move config out of PrimaryNavigationComponent**
- **Update admin template to use new nav formats**
- **Move nav components up from admin directory**


app/components/npq_separation/admin/sub_navigation_component.rb
```diff
@@ -27,79 +29,70 @@ def sub_navigation_link(section)
...
-      end
-
-      def current_section
-        all_sections = [
-          Node.new(name: "Applications", prefix: "/npq-separation/admin/applications"),
-          Node.new(name: "Finance", prefix: "/npq-separation/admin/finance"),
-          Node.new(name: "Schools", prefix: "/npq-separation/admin/schools"),
-          Node.new(name: "Lead providers", prefix: "/npq-separation/admin/lead-providers"),
-          Node.new(name: "Settings", prefix: "/npq-separation/admin/settings"),
+      class SubNavigationStructure
```
cwrw: could we avoid having everything within the same file, and break every class into it's own file/spec as found slightly confusing to follow?

app/components/npq_separation/primary_navigation_component.rb
```diff
@@ -0,0 +1,85 @@
...
+      sections.find(&:current) || sections.first
+    end
+
+  private
+
+    def mark_current(structure)
+      structure.sections.each { |node| node.current = current_path.start_with?(node.prefix) }
+    end
+
+    class PrimaryNavigationStructure
```
cwrw, line 34: `PrimaryNavigationStructure` and `SubNavigationStructure` seem similar, if we extract those to let's say a separate `NavigationStructure` maybe we can use that in all the structures and they can live outside this file?

app/components/npq_separation/admin/sub_navigation_component.rb
peteryates: Yeah, that's fair, updated.

cwrw (approved):
thanks @peteryates !

PR #1223, opened 2024-03-14
[CPDLP-2860] TS1 - Statements endpoint -  pagination

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2860

* Added `API::Pagination` concern
* Updated `API::V3::StatementsController` with pagination
* Added pagination tests for statements.

### DIFF
--- a/app/controllers/api/v3/statements_controller.rb
+++ b/app/controllers/api/v3/statements_controller.rb
@@ -1,12 +1,13 @@
 module API
   module V3
     class StatementsController < BaseController
+      include Pagination
       include ::API::Concerns::FilterByUpdatedSince
 
       before_action :validate_updated_since, only: %i[index]
 
       def index
-        render json: to_json(statements_query.statements)
+        render json: to_json(paginate(statements_query.statements))
       end
 
       def show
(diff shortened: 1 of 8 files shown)


app/controllers/api/v3/statements_controller.rb
```diff
@@ -1,8 +1,10 @@
 module API
   module V3
     class StatementsController < BaseController
+      include Pagination
```
cwrw, line 4: maybe we can incorporate handing wrong page parameters as well, I think now in ECF if you pass in a string it throws an error?

mooktakim: @cwrw added error rescue for wrong page parameter.

config/locales/en.yml
```diff
@@ -4,6 +4,7 @@ en:
   unauthorized: "HTTP Token: Access denied"
   bad_request: "Bad request"
   not_found: "Not found"
+  invalid_page_parameters: "The '#/page[page]' and '#/page[per_page]' parameter values must be a valid number (equal and more than 1)"
```
peteryates, line 7: "...must be a positive number" maybe?

"equal and more than 1" doesn't sound quite right.

cwrw (approved):

cwrw:
@mooktakim comparing this to the old ECF side, I think we should return no results rather than error when the pagination goes over as that's a change in behaviour, let me know your thoughts

cwrw (approved):

PR #1224, opened 2024-03-14
Add rswag and swagger-ui for API documentation

ethax-ross:
[Jira-2862](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-2862)

### Context

We want to document the API endpoints using the OpenAPI specification and surface the documentation using Swagger UI so that users can understand the API behaviour and functionality.

- Add swagger-ui to render OpenAPI yaml

Initially I tried adding via the `swagger-ui` package, which would have been ideal as we would only pull in/package the parts we use, but getting this to properly transpile with our outdated version of `webpacker` proved more hassle than its worth and would have polluted our `package.json` with a whole bunch of extra modules.

Instead, I've pulled in via `swagger-ui-dist` which is a prepackaged version we can simply include and use.

- Add rswag to document statements endpoints

Add `rswag` gem, configured to output OpenAPI spec in `yaml` format.

Set up initial swagger docs for the statements endpoints.

Regenerate the API docs with `rake rswag:specs:swaggerize`.

- Check OpenAPI schema for changes during CI

We want to make sure the various swagger.yaml files are regenerated when a change is made. If they aren't it can get out of sync with the code and leads to issues. We also want to prevent anyone manually modifying the schema files.

Run a checksum on the schema files, then regenerate them from the specs, recalculate the checksums again and fail the build if they are different.

- Precompile assets before running tests

If we don't precompile the assets they are compiled as part of the first feature spec, which can take a long time and cause it to timeout (swagger-ui-dist is the culprit).

- Add documentation for swagger

### Guidance to review

Best reviewed by-commit.

I've opted to keep the `rswag` specs separate to our normal request specs; this adds a bit of duplication but I think its better than polluting them with `rswag` DSL.

It would make this PR too large/complex, but I think we should aim to augment `blueprinter` to accept swagger details so that we could auto-generate the models. For example:

```
field :created_at, description: "The date the statement was created.", type: :string, format: :"date-time", example: "2021-05-31T02:22:32.000Z"
```

[Link to Swagger UI](https://npq-registration-review-1224-web.test.teacherservices.cloud/api/docs/v3)


spec/swagger_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+        version:,
+      },
+      externalDocs: {
+        description: "Find out more about Swagger",
+        url: "https://swagger.io/",
+      },
+      paths: {},
+      servers: [
+        {
+          url: "https://register-national-professional-qualifications.sandbox.education.gov.uk/",
```
ethax-ross, line 36: I need to look into if we can get this dynamically for the environment; I'm not sure if AKS sets an ENV var somewhere or if I'll have to stick it manually into the keyvault.

cwrw: I think we might be able to set the ENV variable here and build the url for the specific environment: https://github.com/DFE-Digital/npq-registration/blob/main/terraform/application/application.tf#L21 so we can read it in the app

spec/swagger_schemas/attributes/id.rb
```diff
@@ -0,0 +1,5 @@
+ID_ATTRIBUTE = {
```
cwrw, line 1: wondering if all those schemas should be under a versioned folder?

.github/workflows/openapi.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Run swaggerize
+        run: bundle exec rake rswag:specs:swaggerize
+
+      - name: Generate API doc checksums (after swaggerize)
+        run: find public/api/docs/ -type f | sort | xargs shasum -a 256 > api-doc-checksums-after-swaggerize.txt
+
+      - name: Compare Checksums
+        run: |
+          if ! diff -q api-doc-checksums-original.txt api-doc-checksums-after-swaggerize.txt; then
+            echo "OpenAPI schema has changed! run rake rswag:specs:swaggerize"
```
cwrw, line 82: 🌟

.github/workflows/deploy.yml
```diff
@@ -23,6 +23,10 @@ jobs:
     name: RSpec
     uses: ./.github/workflows/rspec.yml
 
+  openapi:
+    name: OpenAPI
+    uses: ./.github/workflows/openapi.yml
```
cwrw, line 28: should we particular to the lead provider API here in the naming?

spec/requests/api/docs/v3/statements_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+      parameter name: :page,
+                in: :query,
+                required: false,
+                schema: {
+                  "$ref": "#/components/schemas/PaginationFilter",
+                }
+
+      response "200", "A list of statements as part of which the DfE will make output payments for ecf or npq participants" do
```
cwrw, line 29:
```suggestion
      response "200", "A list of statements as part of which the DfE will make output payments for participants" do
```

public/api/docs/v3/swagger.yaml
```diff
@@ -0,0 +1,211 @@
...
+          "$ref": "#/components/schemas/ListStatementsFilter"
+      - name: page
+        in: query
+        required: false
+        schema:
+          "$ref": "#/components/schemas/PaginationFilter"
+      responses:
+        '200':
+          description: A list of statements as part of which the DfE will make output
+            payments for ecf or npq participants
```
cwrw, line 31:
```suggestion
            payments for participants
```

spec/swagger_schemas/models/statement.rb
```diff
@@ -0,0 +1,71 @@
+STATEMENT = {
+  v3: {
+    description: "A financial statement.",
+    type: :object,
+    required: %i[id type attributes],
+    properties: {
+      id: {
+        description: "The unique identifier of the financial statement.",
+        type: :integer,
+        example: 123,
```
cwrw, line 10: I think since Leandro's pr has been merged, this can change to a UUID value

cwrw (commented):
beautifully and cleanly done 👏 just a few minor comments

spec/swagger_schemas/attributes/id.rb
ethax-ross: So I originally intended to put the schemas in a folder relative to their version, but when I came to construct the schema definition in the `swagger_helper` I found that made it more difficult to do it dynamically; I think having all versions in the same file/constant is going to be more straight forward when we come to add other version definitions, but happy to review it if we run into issues when we come to do that if that would make sense?

cwrw: sure we can keep for now and review as we go 👍

cwrw (approved):
💯 !!

PR #1225, opened 2024-03-14
Add frozen_string_literal ruby magic comment to all files

leandroalemao:
### Context

This `frozen_string_literal` magic comment ensures that once a string literal is defined, its content cannot be modified, making it safer to work with and preventing accidental modifications that may lead to bugs or unexpected behaviour in your Ruby code and also reduce overhead of the garbage collector by eliminating unnecessary memory allocations.

Improve application performance reducing the overhead of garbage collector by adding ruby's magic comment `frozen_string_literal: true` to all files.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -1,8 +1,9 @@
 module API
   class BaseController < ActionController::API
+    before_action :remove_charset
+
     include API::TokenAuthenticatable
 
-    before_action :remove_charset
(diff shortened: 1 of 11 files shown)


peteryates:
Personally I'm not a fan of having to have magic comments in every file but there is that tiny extra level of safety. This is far reaching, we shouldn't merge until the NPQ devs have discussed/approved. cc @javier-npq @slawosz @pmanrubia

PR #1226, opened 2024-03-14
[CPDLP-2879] Add missing `ecf_id` to statements

leandroalemao:
### Context

Ticket: [CPDLP-2879](https://dfedigital.atlassian.net/browse/CPDLP-2879)

We need to maintain all existing ecf ids in the API. This also applies to statements.

- Add `ecf_id` to the statements table
- Migrate the `ecf_id` when copying over statement data
- Surface `ecf_id` and the id of the statement in the API
- Add to seeds as well as a generated UUID

[CPDLP-2879]: https://dfedigital.atlassian.net/browse/CPDLP-2879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/api/v3/statement_serializer.rb
+++ b/app/serializers/api/v3/statement_serializer.rb
@@ -17,6 +17,7 @@ module API
         field :deadline_date, name: :cut_off_date, datetime_format: "%Y-%m-%d"
         field :payment_date, datetime_format: "%Y-%m-%d"
         field(:paid) { |s, _| s.payment_date.present? }
+        field :ecf_id
         field :created_at
         field :updated_at
       end
--- a/app/serializers/api/v3/statement_serializer.rb
+++ b/app/serializers/api/v3/statement_serializer.rb
@@ -1,7 +1,7 @@
 module API
   module V3
     class StatementSerializer < Blueprinter::Base
-      identifier :id
+      identifier :ecf_id, name: :id
       field(:type) { "statement" }
 
       field :attributes do |statement, _options|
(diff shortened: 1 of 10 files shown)


db/migrate/20240312212008_add_ecf_id_to_statements.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddEcfIdToStatements < ActiveRecord::Migration[7.1]
+  def change
+    add_column :statements, :ecf_id, :text, null: true
+    add_index :statements, :ecf_id, unique: true
```
cwrw, line 6: could we add the uniqueness validation to the model as well?

app/serializers/api/v3/statement_serializer.rb
```diff
@@ -17,6 +17,7 @@ class AttributesSerializer < Blueprinter::Base
         field :deadline_date, name: :cut_off_date, datetime_format: "%Y-%m-%d"
         field :payment_date, datetime_format: "%Y-%m-%d"
         field(:paid) { |s, _| s.payment_date.present? }
+        field :ecf_id
```
cwrw, line 20: could we use this instead of the actual `id` and not a new field?

db/ecf_schema.rb
```diff
@@ -207,6 +207,7 @@
     t.datetime "updated_at", null: false
     t.integer "disable_from_year"
     t.boolean "listed", default: false, null: false
+    t.integer "listed_for_school_type_codes", default: [], array: true
```
mooktakim, line 210: I'm not sure if these changes are intentional; seems unrelated to the PR.

mooktakim (approved):
Minor comment about `ecf_schema.rb`, otherwise looks good

db/ecf_schema.rb
cwrw: I think its ok to keep this this time as we have to keep the ecf schema in sync with ECF

cwrw (approved):

PR #1230, opened 2024-03-18
Add support for filtering statements

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2861

Closes: #1221

### Description

Filter statements by `cohorts` and `updated_at`. It has been decided by the team
to not chain the different filters, and pass the parameters to the query service
explicitly.

### DIFF
--- a/app/controllers/api/v3/statements_controller.rb
+++ b/app/controllers/api/v3/statements_controller.rb
@@ -13,27 +13,27 @@ module API
         render json: to_json(statements_query.statement(id: statement_params[:id]))
       end
 
-      private
+    private
 
-        def statements_query
-          Statements::Query.new(
-            lead_provider: current_lead_provider,
-            cohort_start_years:,
-            updated_since:,
-          )
-        end
+      def statements_query
+        Statements::Query.new(
+          lead_provider: current_lead_provider,
+          cohort_start_years:,
+          updated_since:,
+        )
+      end
 
-        def statement_params
-          params.permit(:id, filter: %i[cohort updated_since])
-        end
+      def statement_params
+        params.permit(:id, filter: %i[cohort updated_since])
+      end
 
-        def cohort_start_years
-          statement_params.dig(:filter, :cohort)
-        end
+      def cohort_start_years
+        statement_params.dig(:filter, :cohort)
+      end
 
-        def to_json(obj)
-          StatementSerializer.render(obj, root: "data")
-        end
+      def to_json(obj)
+        StatementSerializer.render(obj, root: "data")
+      end
     end
   end
 end
--- a/spec/requests/api/v3/statements_spec.rb
+++ b/spec/requests/api/v3/statements_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe "Statements endpoint", type: "request" do
           expect(response.status).to eq 200
           expect(response.content_type).to eql("application/json")
           expect(parsed_response["data"].size).to eq(2)
-          expect(response_ids).to match_array([statement1.id, statement2.id])
+          expect(response_ids).to match_array([statement1.ecf_id, statement2.ecf_id])
         end
       end
 


app/controllers/api/v3/statements_controller.rb
```diff
@@ -1,6 +1,10 @@
 module API
   module V3
     class StatementsController < BaseController
+      include ::API::Concerns::FilterByUpdatedSince
```
cwrw, line 4: maybe this can be a service concern rather that in the controller?

config/locales/en.yml
```diff
@@ -4,6 +4,7 @@ en:
   unauthorized: "HTTP Token: Access denied"
   bad_request: "Bad request"
   not_found: "Not found"
+  invalid_updated_since_filter: "The filter '#/updated_since' must be a valid ISO 8601 date"
```
cwrw, line 7:
```suggestion
  invalid_updated_since_filter: "The filter '#/updated_since' must be a valid RCF3339 date"
```

spec/support/helpers/api_helpers.rb
```diff
@@ -30,6 +30,10 @@ def parsed_response
       Oj.load(response.body)
     end
 
+    def response_ids
+      parsed_response["data"].map { |statement| statement["id"] }
```
cwrw, line 34: this is changing since Leandro's pr went in with `ecf_id`

cwrw (commented):
looking good, just some minor comments :)

app/services/statements/query.rb
```diff
@@ -1,13 +1,21 @@
...
+      @cohort_start_years = cohort_start_years&.split(",")
+      @updated_since = updated_since
     end
 
     def statements
-      Statement
-        .includes(:cohort)
-        .where(lead_provider:)
+      scope = Statement
+                .includes(:cohort)
```
peteryates, line 11: Any reason we're using `#includes` here over `#eager_load` - is there any reason we wouldn't want  PostgreSQL to do a left join?

app/services/statements/query.rb
```diff
@@ -1,13 +1,21 @@
...
+                .where(lead_provider:)
```
peteryates, line 12: This ensures all queries need to specify a lead provider and won't really work for admin stuff where we might want to list all the statements.

app/services/statements/query.rb
cwrw: I think leaving the decision to `includes` depending on what's best is probably my preference here for performance rather than being explicit with `eager_load`

app/services/statements/query.rb
cwrw: I think when it comes to reconciliation of services we can revisit this and amend as required, happy to keep as is for now

app/controllers/api/v3/statements_controller.rb
mooktakim: The concern should be located `app/controllers/concerns/api/filter_by_updated_since.rb`
That way the include would be:
```ruby
include FilterByUpdatedSince
```

spec/support/helpers/api_helpers.rb
mooktakim: @cwrw `id` here is correct. Leandros change to the serializer should return `ecf_id` as `id` in the JSON response.

@pmanrubia could you rename the variable `statement` to something general like `data`, as this helper is not specific to statement.

cwrw: ah sorry your right!

app/controllers/api/v3/statements_controller.rb
```diff
@@ -12,13 +16,19 @@ def show
...
+        Statements::Query.new(
           lead_provider: current_lead_provider,
+          cohort_start_years:,
+          updated_since:,
         )
       end
 
       def statement_params
-        params.permit(:id)
+        params.permit(:id, :cohort, filter: %i[cohort updated_since])
```
mooktakim, line 27: fix, `cohort` added twice:
```ruby
params.permit(:id, filter: %i[cohort updated_since])
```

cwrw (approved):
thanks @pmanrubia!

PR #1233, opened 2024-03-19
Groundwork for API Guidance

pmanrubia:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2881

1. Introduces Markdown support for rendering Guidance pages. Uses [Govuk Markdown](https://github.com/DFE-Digital/govuk-markdown) renderer.
2. Supports rendering JSON documents with formatting [via Rouge](https://github.com/rouge-ruby/rouge).
3. Adds the foundation for guidance pages (index / show)

### Review notes

1. A layout has been added as a placeholder, but will be reworked in incoming PRs to match the specs
2. A couple of pages (`test.md` and `nested/nested/test.md`) have been added for testing purposes (feature tests), but they need to be replaced with proper pages once Guidance is completed with final pages.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,7 +62,7 @@ group :development, :test do
   gem "dotenv-rails"
   gem "pry-byebug"
   gem "pry-rails"
-  gem "rspec-rails", "~> 6.1.1"
+  gem "rspec-rails", "~> 6.1.2"
   gem "rubocop-govuk"
   gem "scss_lint-govuk"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -343,12 +343,12 @@ GEM
     net-smtp (0.4.0.1)
       net-protocol
     nio4r (2.7.0)
-    nokogiri (1.16.2)
+    nokogiri (1.16.3)
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
-    nokogiri (1.16.2-x86_64-darwin)
+    nokogiri (1.16.3-x86_64-darwin)
       racc (~> 1.4)
-    nokogiri (1.16.2-x86_64-linux)
+    nokogiri (1.16.3-x86_64-linux)
       racc (~> 1.4)
     notifications-ruby-client (5.4.0)
       jwt (>= 1.5, < 3)
@@ -474,15 +474,15 @@ GEM
     rspec-mocks (3.13.0)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.13.0)
-    rspec-rails (6.1.1)
+    rspec-rails (6.1.2)
       actionpack (>= 6.1)
       activesupport (>= 6.1)
       railties (>= 6.1)
-      rspec-core (~> 3.12)
-      rspec-expectations (~> 3.12)
-      rspec-mocks (~> 3.12)
-      rspec-support (~> 3.12)
-    rspec-support (3.13.0)
+      rspec-core (~> 3.13)
+      rspec-expectations (~> 3.13)
+      rspec-mocks (~> 3.13)
+      rspec-support (~> 3.13)
+    rspec-support (3.13.1)
     rubocop (1.60.2)
       json (~> 2.3)
       language_server-protocol (>= 3.17.0)
@@ -696,7 +696,7 @@ DEPENDENCIES
   rack-attack
   rails (~> 7)
   rouge
-  rspec-rails (~> 6.1.1)
+  rspec-rails (~> 6.1.2)
   rubocop-govuk
   rubyzip
   scenic
--- a/docs/environments.md
+++ b/docs/environments.md
@@ -4,28 +4,36 @@
 
 There are three permanent environments for NPQ and a fourth set of transient environments, the three permanent are sandbox, dev (Staging in Azure) , and production. The fourth transient set are review apps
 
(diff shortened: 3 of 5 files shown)


app/models/guidance_page.rb
```diff
@@ -0,0 +1,22 @@
+class GuidancePage
+  def initialize(path, content: nil)
+    @path = path
+    @content = content
+  end
+
+  def sub_headings
+    page_contents.scan(/^##\s(.*)$/).map(&:first)
```
ethax-ross, line 8: I wonder if we should have this as `def sub_heading_links` and return a hash here to make rendering the links easier? something like:

```
{
  "#sub-heading-1" => "Sub heading 1",
  "#sub-heading-2" => "Sub heading 2",
}
```

config/initializers/markdown.rb
```diff
@@ -0,0 +1,25 @@
+require "rouge/plugins/redcarpet"
+
+module TemplateHandlers
```
ethax-ross, line 3: Can we add a test for this to check its converting generic Markdown into HTML, doing the appropriate code highlighting correctly and rendering specific markdown from the gov uk renderer? we should be able to do it as a view spec, something like:

```
describe TemplateHandlers::Markdown, type: :view do
  subject { rendered }

  context "when rendering generic markdown" do
    let(:markdown) do
      <<~MARKDOWN
        # Heading

        Lorem ipsum

        With a [link](https://www.gov.uk)

        With a https://www.gov.uk/autolink
      MARKDOWN
    end

    before do
      stub_template "test.md" => markdown
      render template: "test"
    end

    it { is_expected.to have_css("p", text: "Lorem") }
    it { is_expected.to have_css("h1", text: "Heading") }
    it { is_expected.to have_css("p", count: 3) }
    it { is_expected.to have_css('a[href="https://www.gov.uk"]', text: "link") }
    it { is_expected.to have_css('a[href="https://www.gov.uk/autolink"]', text: "https://www.gov.uk/autolink") }
  end

  context "when rendering GOV UK specific markdown" do
    ...
  end

  context "when rendering code snippets" do
    ...
  end
end
```

app/views/layouts/api_guidance.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<html lang="en" class="govuk-template admin--wide">
+<%= render partial: "layouts/shared/head" %>
+
+<body class="govuk-template__body ">
+
+<a href="#main-content" class="govuk-skip-link">Skip to main content</a>
+
+<%= render partial: "layouts/shared/header" %>
+
+<div class="govuk-width-container">
```
ethax-ross, line 11: If you branch off #1222 you could pull in the new navigation components here

app/models/guidance_page.rb
peteryates: Would that lead to the slight difficulty of making sure the `href` matched the `id` generated by the markdown library?

ethax-ross: Potentially, but I'm not sure if there's a way around that as we render the side navigation outside of the markdown template rendering anyway? one way or another we need to render some side links that link into the separately rendered markdown headings, so the link construction has to happen somewhere?

ethax-ross: ah I'm forgetting about being able to do `anchor: 'some heading'` in the path :) never mind

config/routes.rb
```diff
@@ -77,6 +77,9 @@
   get "/admin", to: "admin#show"
 
   namespace :api do
+    constraints -> { Rails.application.config.npq_separation[:api_enabled] } do
```
cwrw, line 80: maybe we can add this under the npq-separation branch to avoid this being in prod?

ethax-ross: The `api_enabled` flag for npq separation is already disabled in prod ^ we should probably add a few request specs to cover this constraint though

cwrw: ah ok so not under `npq-separation/` but it's protected anyway, makes sense

app/views/layouts/api_guidance.html.erb
ethax-ross: Yeah that's fair, I don't think I called it out explicitly in any of the tickets but I've added it to the index page ticket 👍

app/models/guidance_page.rb
```diff
@@ -0,0 +1,24 @@
+class GuidancePage
+  def initialize(path, content: nil)
+    @path = path
+    @content = content
+  end
+
+  def sub_headings
+    headings = page_contents.scan(/^##\s(.*)$/).map(&:first)
+
+    headings.index_by { |heading| heading.underscore.parameterize.gsub("_", "-") }
```
ethax-ross, line 10: `parameterize(separator: "-")`

ethax-ross: ah sorry I wasn't very clear, I just meant you could drop the `gsub` in favour of specifying `separator` to the `parameterize` call, so this should be equivalent but without the need for the extra `gsub`:

```
headings.index_by { |heading| heading.underscore.parameterize(separator: "-") }
```

ethax-ross (approved):
Looks good 👍

app/models/guidance_page.rb
ethax-ross: oh I follow now - ignore me, I must need a coffee 😄

PR #1235, opened 2024-03-19
Add user validation rules from ECF

ethax-ross:
### Context

We want the validation rules for users to be consistent with ECF, as we're bringing across logic and data from that service into NPQ reg.

- Add user validation rules from ECF

Brings user validation rules from ECF into NPQ reg for consistency.

- Make email/uid validation conditional

As we have invalid data in the NPQ reg database at the moment that we can't easily fix, we only want to apply the email validation when validating logic that is part of the NPQ separation work. Likewise, the `uid` of an existing user can change via the `UserUpdater` at the moment, so we only want to scope that validation to particular scenarios.

Change `email` and `uid` validation to apply only with a custom context, for example: `user.valid?(:npq_separation)`.

Supress Rubocop warning; until we can fix the data and apply the unique constraint on the database.

### Guidance to review

On testing this against users in NPQ reg production we have the following data issues:

```
{
  "Full name can't be blank" => 2211,
  "Email has already been taken" => 339,
  "Email can't be blank" => 73,
  "Email failed notify_email validation" => 75
}
```

We have cleaned up the blank emails/full names and invalid emails in the NPQ reg production database. The duplicate emails are going to be harder to fix, hence why we only apply the validation with a custom context.

It also appears a user uid/get an identity id can change through the `UserUpdater` service; we need to decide if we want to continue to allow this and therefore relax the validation rule this PR introduces that prevents it:

https://github.com/DFE-Digital/npq-registration/blob/main/app/services/get_an_identity_service/user_updater.rb#L25

Fields in ECF that were not in NPQ reg:

```
last_sign_in_at - not used
current_sign_in_at - check if induction tutor has signed in, not needed
current_sign_in_ip - not used
last_sign_in_ip - not used
sign_in_count - not used
discarded_at - soft delete, not needed

archived_email - not used as we're not migrating archived users
archived_at - not used as we're not migrating archived users

get_an_identity_id - already present as uid in NPQ users
```

Running on prod after fixing up data:

```
invalid_users = []
npq_separation_invalid_users = []
total = User.count
i = 0

User.find_each do |user|
  puts "Progress: #{(i.to_f / total) * 100}"
  
  invalid_users << user if user.invalid?
  npq_separation_invalid_users << user if user.invalid?(:npq_separation)
  i += 1
end

[2] pry(main)> invalid_users.size
=> 0
[3] pry(main)> npq_separation_invalid_users.size
=> 316
```


app/models/user.rb
```diff
@@ -4,6 +4,11 @@ class User < ApplicationRecord
   has_many :applications, dependent: :destroy
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
 
+  validates :full_name, presence: true
+  validates :email, presence: true, uniqueness: true, notify_email: true, on: :npq_separation # rubocop:disable Rails/UniqueValidationWithoutIndex
+  validates :uid, uniqueness: { allow_blank: true }
```
cwrw, line 9: maybe we can scope this to separation as well just incase? Im worried that this might still happen while the email validation is not set?

cwrw (commented):
looks great, just the one comment otherwise happy for this to be merged, might be worth double checking that everything is still valid today :)

app/models/user.rb
ethax-ross: Yeah can do; I checked the database and all the `uid` values were currently unique but that may not always be the case I guess 👍

ethax-ross: Actually I think we're OK to keep this one as there is a unique constraint in the DB:

https://github.com/DFE-Digital/npq-registration/blob/main/db/schema.rb#L362

cwrw (approved):

PR #1243, opened 2024-03-21
Fix GitHub Codespaces

ethax-ross:
The main issues fixed in this PR are:

- Expose the Postgres port from the container to the host
- Set Postgres credentials on Dockerfile
- Create `vscode` role/user in Postgres on boot
- Start `rails server` automatically
- Disable CSRF checks on Codespaces (as its failing but I'm not sure why)


config/environments/development.rb
```diff
@@ -78,4 +78,10 @@
     api_enabled: true,
     migration_enabled: true,
   }
+
+  # Disable origin check for Cross-Site Request Forgery (CSRF) protection for codespaces.
```
peteryates, line 82: Does this happen when the CodeSpace is running in a browser or locally via VS Code? Or both? 🤔

ethax-ross: I tested it running in a browser as I figured that's how the content editors are going to use it... its an odd one but there's a few similar reports kicking around in various GitHub issues about it

peteryates (approved):
Nice, this looks great!

config/environments/development.rb
peteryates: Yeah, I've seen it in the past but from memory the process is smoother when running from VS Code locally. But the fix  here looks like it'll do the trick nicely 👌🏽

PR #1246, opened 2024-03-21
[CPDLP-2897] Add failures reporting to data migration

leandroalemao:
### Context

Ticket: [CPDLP-2897](https://dfedigital.atlassian.net/browse/CPDLP-2897)

In our current data migrators between ECF and NPQ, we've only tackled tables that had no failures.
But failures may happen with other tables, so in order to capture those failures we'd like to build in the correct tooling to store those values and allow this to be downloaded/analysed separately.

Add reporting to the data migrators in NPQ. Also allow this report to be downloaded from the FE.

Screenshot 2024-03-21 at 15 53 14

[CPDLP-2897]: https://dfedigital.atlassian.net/browse/CPDLP-2897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/controllers/npq_separation/migration/migrations_controller_spec.rb
```diff
@@ -55,4 +55,26 @@
...
+      it "asks the user to sign in as an admin" do
+        follow_redirect!
+        expect(response.body).to include("Sign in with your administrator account")
+      end
+    end
+
+    context "when signed in as a super admin" do
+      let(:super_admin) { true }
+
+      it { expect(response).to be_successful }
```
ethax-ross, line 77: Worth asserting on the contents of the response `yaml` here as well?

app/services/migration/migrator.rb
```diff
@@ -49,12 +49,16 @@ def migrate(items, model)
...
       data_migration.update!(started_at: Time.zone.now, total_count: items.count)
 
-      items.each do |item|
+      failures_to_record = items.each_with_object([]) do |item, failures|
         data_migration.increment!(:processed_count)
         result = yield(item)
-        data_migration.increment!(:failure_count) unless result
+        unless result
+          data_migration.increment!(:failure_count)
+          failures << item
```
ethax-ross, line 57: Would this be simpler/easier to follow if we recorded on an item by item basis instead of collecting them? so:

```
data_migration.increment!(:failure_count)
failure_recorder.record_failure(data_migration:, item:)
```

app/services/migration/failures_recorder.rb
```diff
@@ -0,0 +1,46 @@
+module Migration
+  class FailuresRecorder
```
ethax-ross, line 2: It doesn't feel quite right to have failure logic and cache references in the `DataMigration` class as well as the `FailureRecorder`. If we come to change how this is done in the future (writing them to a DB for example instead of the cache) I would only expect to have to change the logic in one place.

Would this class be better suited as a more generic module that both reads and writes, something like:

```
class FailureManager
  def initialize(data_manager)
  end
  
  def add_failure(failure)
    # add failure to cache
  end
  
  def failures
    # return all failures from cache
  end
end
```

ethax-ross (commented):
Looking good, just some general thoughts on a couple of bits

spec/support/helpers/admin_login.rb
```diff
@@ -9,6 +9,8 @@ def sign_in_as_super_admin
     end
 
     def sign_in_as(admin_account)
+      Rack::Attack.enabled = false
```
cwrw, line 12: ah was rack attack catching those in the tests?

app/services/migration/migrator.rb
cwrw: also to your question earlier Leandro, maybe we can also record the error message as well and pass it in with the item 🤔 otherwise will be quite hard to debug for now? let me know your thoughts

spec/lib/services/migration/failures_manager_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe Migration::FailuresManager, in_memory_rails_cache: true do
+  let(:instance) { described_class.new(data_migration:, items:) }
+
+  describe "#record" do
```
cwrw, line 6: maybe we can test the reading of the failures as well?

spec/support/helpers/admin_login.rb
leandroalemao: yeah rack attack was blocking the sign in page in tests ([/session/sign-in-code](https://github.com/DFE-Digital/npq-registration/blob/main/config/initializers/rack_attack.rb#L15)) max 5 calls per minute.

leandroalemao:
Hi @cwrw @ethax-ross .. comments have been addressed tks :)

app/models/migration/data_migration.rb
```diff
@@ -40,5 +40,9 @@ def in_progress?
     def complete?
       completed_at.present?
     end
+
+    def migration_failures_key
```
ethax-ross, line 44: Can we encapsulate this in the `FailureManager` service? I don't think the model should know about failures

app/services/migration/failures_manager.rb
```diff
@@ -0,0 +1,60 @@
+module Migration
+  class FailuresManager
+    class << self
+      def record_failures(data_migration:, items:)
+        new(data_migration:, items:).record_failures
+      end
+
+      def failures_for(data_migration:)
+        new(data_migration:).failures
```
ethax-ross, line 9: I think having the class and instance methods makes this interface more confusing than it needs to be; can we settle on one or the other? personally I think having instance methods makes it all neater, then I think we only really need two methods, one to record a failure and one to get all recorded failures:

```
failure_manager = FailureManager.new(data_migration:)
failure_manager.record_failure(item, failure_message)
failure_manager.all_failures
```

app/services/migration/failure_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      @item = item
+      @failure_message = failure_message
+
+      write_failure
+      increment_counter
+    end
+
+    def all_failures
+      return unless data_migration
```
ethax-ross, line 14: We could just check `data_migration` is present in the initializer and raise an `ArgumentError` if its `nil` instead of doing it on every instance method

app/services/migration/failure_manager.rb
```diff
@@ -0,0 +1,65 @@
+module Migration
+  class FailureManager
+    def record_failure(item, failure_message)
+      return unless data_migration && item.present?
+
+      @item = item
+      @failure_message = failure_message
```
ethax-ross, line 7: Do these need to be instance variables? why not `write_failure(message, item)`

app/services/migration/failure_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+    def initialize(data_migration:)
+      @data_migration = data_migration
+      @counter = 1
+    end
+
+    def increment_counter
+      @counter += 1
+    end
+
+    def attributes
```
ethax-ross, line 32: Is this used?

app/services/migration/failure_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+        failure_message:,
+      }
+    end
+
+    def migration_failure_key
+      @migration_failure_key ||= "migration_failure_#{data_migration.model}_#{data_migration.id}"
+    end
+
+    def write_failure
+      Rails.cache.write("#{migration_failure_key}_#{counter}", failure, expires_in: 1.month)
```
ethax-ross, line 52: I think storing the failures under a single key per data migration would simplify a lot of this (I don't think there Redis value size limit will be an issue). So having:

```
data_migration.id => {
  failure_message_1 => [item1, item2],
  failure_message_2 => [item1, item2]
}
```

Then we don't need to maintain a `counter` which could cause issues if we ended up with two instances of a `FailureManager` with the same `data_migration` recording failures. Having the failures grouped by failure message is probably how we would look at the data as well I imagine 🤔

app/controllers/npq_separation/migration/migrations_controller.rb
```diff
@@ -14,6 +14,20 @@ def create
     redirect_to npq_separation_migration_migrations_path
   end
 
+  def download_report
+    data_migration = Migration::DataMigration.find(params[:id])
+    failures = Migration::FailureManager.new(data_migration:).all_failures
+
+    failures_csv = CSV.generate(headers: true) do |csv|
```
ethax-ross, line 21: If we're changing this to a `CSV` it would be better in `FailureManager#all_failures_csv` so we can test it explicitly

spec/lib/services/migration/failure_manager_spec.rb
```diff
@@ -0,0 +1,70 @@
+require "rails_helper"
+
+RSpec.describe Migration::FailureManager, in_memory_rails_cache: true do
+  let(:instance) { described_class.new(data_migration:) }
+
+  describe "#record_failures" do
```
ethax-ross, line 6: `#record_failure`

spec/lib/services/migration/failure_manager_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    context "with correct params" do
+      let(:data_migration) { create(:data_migration, model: :statement, failure_count: 2) }
+      let(:item) { create(:ecf_migration_statement) }
+      let(:failure_message) { "Test failure" }
+
+      it "records a failure" do
+        subject
+
+        failure = Rails.cache.read("migration_failure_#{data_migration.model}_#{data_migration.id}_1")
```
ethax-ross, line 17: Could call `all_failures` here to get it back so we're not relying on an implementation detail

spec/lib/services/migration/failure_manager_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+
+    context "with incorrect params" do
+      let(:data_migration) { create(:data_migration) }
+      let(:item) { "" }
+      let(:failure_message) { "Test failure" }
+
+      it "returns nil" do
+        expect(subject).to be_nil
```
ethax-ross, line 31: This is testing the return value of the method, not that it hasn't recorded a failure?

```
expect(instance.all_failures).to be_empty
```

spec/lib/services/migration/failure_manager_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+  end
+
+  describe "#failures" do
```
ethax-ross, line 36: `#all_failures`

ethax-ross (commented):
Looks good, just one last nit pick 😄

ethax-ross (approved):
Looks good 👏

app/services/migration/failure_manager.rb
```diff
@@ -0,0 +1,45 @@
+module Migration
+  class FailureManager
+    def record_failure(item, failure_message)
```
cwrw, line 3: 👍

cwrw (approved):

PR #1252, opened 2024-03-26
[CPDLP-2898] Add logic to migrate ECF user data into NPQ

leandroalemao:
### Context

Ticket: [CPDLP-2898](https://dfedigital.atlassian.net/browse/CPDLP-2898)

We would like to migrate existing user data into NPQ from ECF. There is an existing table in ECF which means there will be conflicts

 Add logic to migrate ECF user data into NPQ.

Screenshot 2024-03-26 at 18 40 18

[CPDLP-2898]: https://dfedigital.atlassian.net/browse/CPDLP-2898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/migration/ecf/finance/schedule.rb
+++ b/app/models/migration/ecf/finance/schedule.rb
@@ -1,0 +1,10 @@
+module Migration::Ecf::Finance
+  class Schedule < Migration::Ecf::BaseRecord
+    self.inheritance_column = nil
+
+    belongs_to :cohort
+    has_many :participant_profiles
+
+    default_scope { where("schedules.type ilike ?", "Finance::Schedule::NPQ%") }
+  end
+end
--- a/app/models/migration/ecf/npq_application.rb
+++ b/app/models/migration/ecf/npq_application.rb
@@ -1,0 +1,10 @@
+module Migration::Ecf
+  class NpqApplication < BaseRecord
+    belongs_to :participant_identity
+    belongs_to :npq_lead_provider
+    belongs_to :npq_course
+    belongs_to :cohort, optional: true
+    has_one :profile, class_name: "ParticipantProfile", foreign_key: :id
+    has_one :user, through: :participant_identity
+  end
+end
--- a/app/models/migration/ecf/participant_profile.rb
+++ b/app/models/migration/ecf/participant_profile.rb
@@ -1,0 +1,12 @@
+module Migration::Ecf
+  class ParticipantProfile < BaseRecord
+    self.inheritance_column = nil
+
+    belongs_to :teacher_profile
+    belongs_to :schedule, class_name: "Finance::Schedule"
+    has_one :user, through: :teacher_profile
+    has_one :npq_application, foreign_key: :id
+
+    default_scope { where(type: "ParticipantProfile::NPQ") }
+  end
+end
--- a/app/services/migration/migrator.rb
+++ b/app/services/migration/migrator.rb
@@ -5,13 +5,18 @@ module Migration
     class MigrationNotPreparedError < RuntimeError; end
     class MigrationAlreadyPreparedError < RuntimeError; end
 
-    class << self
-      MODELS_TO_BE_MIGRATED = %i[lead_provider cohort statement].freeze
+    MIGRATORS = [
+      Migration::Migrators::LeadProvider,
+      Migration::Migrators::Cohort,
+      Migration::Migrators::Statement,
+      Migration::Migrators::User,
+    ].freeze
 
+    class << self
       def prepare_for_migration
         raise MigrationAlreadyPreparedError, "The migration has already been prepared" if DataMigration.exists?
 
-        MODELS_TO_BE_MIGRATED.each { |model| DataMigration.create!(model:) }
+        MIGRATORS.each(&:prepare!)
       end
     end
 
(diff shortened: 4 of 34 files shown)


app/models/migration/ecf/participant_profile.rb
```diff
@@ -0,0 +1,12 @@
+module Migration::Ecf
+  class ParticipantProfile < BaseRecord
+    self.inheritance_column = nil
+
+    belongs_to :teacher_profile
+    belongs_to :schedule, class_name: "Finance::Schedule"
+    has_one :user, through: :teacher_profile
+    has_one :npq_application, foreign_key: :id
+
+    default_scope { where("participant_profiles.type = ?", "ParticipantProfile::NPQ") }
```
peteryates, line 10: Out of interest, how come SQL over just using AR?

```suggestion
    default_scope { where(type: "ParticipantProfile::NPQ") }
```

app/models/migration/ecf/participant_profile.rb
```diff
@@ -0,0 +1,12 @@
+module Migration::Ecf
+  class ParticipantProfile < BaseRecord
+    self.inheritance_column = nil
+
+    belongs_to :teacher_profile
+    belongs_to :schedule, class_name: "Finance::Schedule"
```
cwrw, line 6: Im wondering if we need to pull in schedules 🤔

app/models/migration/ecf/npq_application.rb
```diff
@@ -0,0 +1,10 @@
+module Migration::Ecf
+  class NpqApplication < BaseRecord
```
cwrw, line 2: I wonder if we should pull NPQ application yet, even though we're looking for TRNs those all come from NPQ reeg anyway, maybe we hold off on this for now?

app/services/migration/migrator.rb
```diff
@@ -5,8 +5,15 @@ class MigrationAlreadyRanError < RuntimeError; end
...
+    MIGRATORS = [
+      Migration::Migrators::LeadProvider,
+      Migration::Migrators::Cohort,
+      Migration::Migrators::Statement,
+      Migration::Migrators::User,
+    ].freeze
+
     class << self
-      MODELS_TO_BE_MIGRATED = %i[lead_provider cohort statement].freeze
+      MODELS_TO_BE_MIGRATED = %i[lead_provider cohort statement user].freeze
```
ethax-ross, line 16: I think it would make sense to encapsulate this/the prep logic in the individual migrators, that way we only need to maintain the array of `MIGRATORS` here and we can do something like:

```
def prepare_for_migration
  raise MigrationAlreadyPreparedError, "The migration has already been prepared" if DataMigration.exists?
  MIGRATORS.each(:prepare!)
end
```

And in the migrator:

```
def prepare!
  DataMigration.create!(model: :foo)
end
```

That feels more extensible as well in case we need to do more prep for a particular migrator?

app/services/migration/migrators/user.rb
```diff
@@ -0,0 +1,34 @@
+module Migration::Migrators
+  class User < Base
+    def call
+      migrate(ecf_users, :user) do |ecf_user|
```
ethax-ross, line 4: Are we going to match on other attributes as part of this PR, like `trn` and `get_an_identity_id`? and how long does it take to run this migration?

cwrw: I think matching on just `ecf_id` for now might be the first step, but interested in the numbers and failures returned, before any data cleanup, also the length as you mentioned

app/models/migration/ecf/finance/schedule.rb
```diff
@@ -0,0 +1,10 @@
+module Migration::Ecf::Finance
+  class Schedule < Migration::Ecf::BaseRecord
+    self.inheritance_column = nil
+
+    belongs_to :cohort
+    has_many :participant_profiles
+
+    default_scope { where("schedules.type ilike ?", "Finance::Schedule::NPQ%") }
```
cwrw, line 8:
```suggestion
    default_scope { where(type: "Finance::Schedule::NPQ") }
```

leandroalemao: Hi @cwrw Think we need to keep the `ilike ?", "Finance::Schedule::NPQ%"` as the schedules types in ECF are named as for example: `"Finance::Schedule::NPQEhco",
 "Finance::Schedule::NPQSpecialist",
 "Finance::Schedule::NPQLeadership",
 "Finance::Schedule::NPQSupport", etc`

cwrw: ah right gotcha! 👍

app/services/migration/migrators/user.rb
```diff
@@ -0,0 +1,34 @@
...
+        user = ::User.find_or_initialize_by(ecf_id: ecf_user.id)
+
+        user.update!(
+          trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : ecf_user.npq_applications.map(&:teacher_reference_number).compact.uniq.last,
```
cwrw, line 8: should we throw an error if there are multiple TRNs from npq applications 🤔 as there should only be one

app/services/migration/migrators/user.rb
```diff
@@ -0,0 +1,34 @@
...
+          full_name: ecf_user.full_name || user.full_name,
+          email: ecf_user.email || user.email,
+          uid: ecf_user.get_an_identity_id || user.uid,
```
cwrw, line 11: I think we should match on user id rather than get an identity id?

cwrw (approved):
Thanks @leandroalemao 🙌 looking great!

I've deployed and tested it and it all goes through on the migration env ⭐ 
A few next steps for other tickets:
- Look into performance on next model we migrate, although it did go through cpu was very high so we can tackle this separately
- Add Redis so we can view the reports, and iterate over matching

PR #1271, opened 2024-04-04
Add workflow params to deploy github actions

leandroalemao:
### Context

We need a simpler way of deploying a specific branch/PR to a specific environment.

Add workflow params to `deploy` Github action like we have in ECF app.

image


cwrw:
@leandroalemao thanks for this, will test/deploy this today and hopefully approve after :)

cwrw (approved):
let's try merging as deploying didn't work and monitor

PR #1276, opened 2024-04-05
Add aks upload and download commands to Makefile

peteryates:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1766

This PR adds some commands to the Makefile which will let us copy data to and from Kubernetes pods. It will be used to sync the school metadata we hold on NPQ Applications.

I've manually tested this for both `staging` and a review app with success.


leandroalemao (approved):
👍 tks

PR #1279, opened 2024-04-09
[CPDLP-2940] Add courses migrator

leandroalemao:
### Context

Ticket: [CPDLP-2962](https://dfedigital.atlassian.net/browse/CPDLP-2962)

We would like to migrate existing courses data into NPQ from ECF.

 Add logic to migrate ECF npq courses data into NPQ reg app.

image

[CPDLP-2962]: https://dfedigital.atlassian.net/browse/CPDLP-2962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/migration/migrator.rb
+++ b/app/services/migration/migrator.rb
@@ -11,6 +11,7 @@ module Migration
       Migration::Migrators::Statement,
       Migration::Migrators::User,
       Migration::Migrators::School,
+      Migration::Migrators::Course,
     ].freeze
 
     class << self
--- a/app/services/migration/migrators/course.rb
+++ b/app/services/migration/migrators/course.rb
@@ -1,0 +1,19 @@
+module Migration::Migrators
+  class Course < Base
+    def call
+      migrate(ecf_npq_courses, :course) do |ecf_npq_course|
+        course = ::Course.find_or_initialize_by(ecf_id: ecf_npq_course.id)
+        course.update!(
+          name: ecf_npq_course.name,
+          identifier: ecf_npq_course.identifier,
+        )
+      end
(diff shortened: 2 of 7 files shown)


config/environments/migration.rb
```diff
@@ -4,6 +4,13 @@
   config.log_level = :debug
   config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/healthcheck") } } }
 
+  config.action_controller.perform_caching = true
```
ethax-ross, line 7: Does the migration environment not have Redis already setup for caching? I think we should keep it as close to prod as possible

cwrw: we don't have Redis in NPQ reg, adding a ticket to add it

app/models/course.rb
```diff
@@ -1,4 +1,6 @@
 class Course < ApplicationRecord
+  validates :name, presence: true
```
cwrw, line 2: ✂️ as this has been added in another pr :)

spec/models/course_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 RSpec.describe Course do
+  describe "validations" do
```
cwrw, line 4: ✂️

cwrw (commented):
looks good, however keeps failing when I try to test on migration, will try running tomorrow

cwrw (approved):
tested on migration, thanks @leandroalemao

PR #1283, opened 2024-04-11
Support database refreshing in monitoring checks

ethax-ross:
### Context

We have a `migration` environment where we can trigger a database refresh from the `production` database via a GitHub action at any time. As the monitoring checks to see if the database is populated this causes the instance to become unhealthy whilst refreshing the database.

We want to allow the database to be empty for the `migration` environment to allow for this.

- Support database refreshing in monitoring checks

General monitoring house keeping and test coverage improvements.


cwrw (approved):
👍 for moving to request specs

app/controllers/monitoring_controller.rb
```diff
@@ -1,35 +1,43 @@
-class MonitoringController < ActionController::Base
+class MonitoringController < ApplicationController
+  skip_before_action :set_feature_flag_users
```
cwrw, line 2: we should also skip the set sentry user as it hits the database as well (tested on migration)

cwrw: and also the session as it's not needed for the healthcheck

cwrw (approved):
⭐

PR #1284, opened 2024-04-11
Add Redis cache to migration environment

ethax-ross:
### Context

We are using the cache in the migration environment to store failure reports for the migration runs. At the moment these persist to an in-memory cache, so we can't access it from the application worker (as the cache is written from a job worker).

Add Redis for a shared cache that we can reliably query after writing from any instance.

- Add Redis cache to migration environment


terraform/application/application.tf
```diff
@@ -25,6 +25,7 @@ module "application_configuration" {
   }
   secret_variables = {
     DATABASE_URL = module.postgres.url
+    REDIS_CACHE_URL = var.deploy_redis_cache ? module.redis-cache[0].url : ""
```
cwrw, line 28: should this be ` module.redis-cache.url` 🤔 rather than `module.redis-cache[0].url`?

ethax-ross: I thought so too, but I think because I'm using `count` to determine if it should spin up a Redis instance this ends up being a collection. I'll make sure to double check the env variable when I deploy it before I merge this down though.

terraform/application/variables.tf
```diff
@@ -39,6 +39,10 @@ variable "deploy_snapshot_database" {
   type    = string
   default = false
 }
+variable "deploy_redis_cache" {
+  type    = string
```
leandroalemao, line 43: `type = bool` maybe?

VenkataChalla554:
[Create terraform environment for NPQ separation](https://trello.com/c/u3polftu/1710-create-terraform-environment-for-npq-separation)

leandroalemao (approved):
Happy to test it in migration env

cwrw (approved):

PR #1289, opened 2024-04-12
Add missing fields to Application from ECF

ethax-ross:
> ⚠️ We need to see if a unique index can be safely added to the `ecf_id` column on the `users` table.

### Context

Adds missing fields to the `Application` model from ECF.

- Add missing fields to Application from ECF

### Guidance to review

We need to make the `ecf_id` on the `users` table unique in order to have it as a foreign key for the `elgiible_for_funding_udpated_by` relationship. It looks like it [was unique originally](https://github.com/DFE-Digital/npq-registration/commit/face8c072dce65ee163d004adb727c9f66f5761a) but [later changed](https://github.com/DFE-Digital/npq-registration/commit/62d30b4786b536afb004a7be86a7fbc370bf0522).

It also looks like the `ecf_id` on the existing NPQ reg models is a `text` field whereas it has a `uuid` type in ECF. We should probably update all these in a separate PR to be consistent.


cwrw:
chatted in person but for posterity: we will be changing the ECF id unique index to allow nils until we fix the data

cwrw (approved):
thanks @ethax-ross

PR #1295, opened 2024-04-15
[CPDLP-2992] Amend application fields to enums

leandroalemao:
### Context

After the investigation in https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2932, we've found that a number of fields are enums in ECF but not in NPQ.

We need to change those fields to enums in NPQ.

Change the following fields to enums in NPQ applications table:

- headteacher_status
- funding_choice
- lead_provider_approval_status

Note: I'm using migrations `up/down` instead of `change` method so migrations can be reversible.

### DIFF
--- a/db/migrate/20240412131057_create_headteacher_statuses_enum.rb
+++ b/db/migrate/20240412131057_create_headteacher_statuses_enum.rb
@@ -1,0 +1,5 @@
+class CreateHeadteacherStatusesEnum < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :headteacher_statuses, %w[no yes_when_course_starts yes_in_first_two_years yes_over_two_years yes_in_first_five_years yes_over_five_years]
+  end
+end
--- a/db/migrate/20240415103921_create_funding_choices_enum.rb
+++ b/db/migrate/20240415103921_create_funding_choices_enum.rb
@@ -1,0 +1,5 @@
+class CreateFundingChoicesEnum < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :funding_choices, %w[school trust self another employer]
+  end
+end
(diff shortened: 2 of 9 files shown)


db/schema.rb
```diff
@@ -19,6 +19,9 @@
 
   # Custom types defined in this database.
   # Note that some types may not work with other database engines. Be careful if changing database.
+  create_enum "funding_choices", ["school", "trust", "self", "another", "employer"]
+  create_enum "headteacher_statuses", ["no", "yes_when_course_starts", "yes_in_first_two_years", "yes_over_two_years", "yes_in_first_five_years", "yes_over_five_years"]
```
peteryates, line 23: I've double checked these values in prod and they look fine ✅

db/migrate/20240412131057_create_headteacher_statuses_enum.rb
```diff
@@ -0,0 +1,5 @@
+class CreateHeadteacherStatusesEnum < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :headteacher_statuses, %w[no yes_when_course_starts yes_in_first_two_years yes_over_two_years yes_in_first_five_years yes_over_five_years]
```
peteryates, line 3: I've double checked these values in prod and they look fine ✅

db/migrate/20240415104435_create_lead_provider_approval_statuses_enum.rb
```diff
@@ -0,0 +1,5 @@
+class CreateLeadProviderApprovalStatusesEnum < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :lead_provider_approval_statuses, %w[pending accepted rejected]
```
peteryates, line 3: I've checked these in prod too and they're fine ✅

db/migrate/20240415103921_create_funding_choices_enum.rb
```diff
@@ -0,0 +1,5 @@
+class CreateFundingChoicesEnum < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :funding_choices, %w[school trust self another employer]
```
peteryates, line 3: And these ✅

peteryates (approved):
This looks great. I think we should do a manual deploy to `migration` to make sure the existing values are cast without problems before merging.

cwrw (approved):
Thanks @leandroalemao tested on migration env and all works as expected

PR #1298, opened 2024-04-16
Migrate schools data from ECF to NPQ reg

ethax-ross:
### Context

Add migrator to migrate school data from ECF to NPQ reg. Initially we are only performing an integrity check to ensure a corresponding school with the same URN/name exists. We are also only looking at schools in ECF that have NPQ applications associated with them.

- Migrate schools data from ECF to NPQ reg

### Guidance for review

I ran this manually on the migration environment:

```
 id: 16,
 model: "school",
 processed_count: 21026,
 failure_count: 3,
 started_at: Tue, 16 Apr 2024 09:39:15.366959000 UTC +00:00,
 completed_at: Tue, 16 Apr 2024 10:07:33.029796000 UTC +00:00,
 created_at: Tue, 16 Apr 2024 09:34:15.669314000 UTC +00:00,
 updated_at: Tue, 16 Apr 2024 10:07:33.031010000 UTC +00:00,
 total_count: 21026
```

### DIFF
--- a/app/models/migration/ecf/npq_application.rb
+++ b/app/models/migration/ecf/npq_application.rb
@@ -6,5 +6,6 @@ module Migration::Ecf
     belongs_to :cohort, optional: true
     has_one :profile, class_name: "ParticipantProfile", foreign_key: :id
     has_one :user, through: :participant_identity
+    has_one :school, class_name: "School", foreign_key: :urn, primary_key: :school_urn
   end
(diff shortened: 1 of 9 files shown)


cwrw (approved):
Looks good, ran on the migration env and there are 3 schools that don't match name and urn which can be easily fixed.

PR #1300, opened 2024-04-17
[CPDLP-2961] Add applications migrator

leandroalemao:
### Context

Ticket: [CPDLP-2961](https://dfedigital.atlassian.net/browse/CPDLP-2961)

We would like to migrate existing applications data into NPQ from ECF.

- Add logic to migrate ECF npq applications data into NPQ reg app.
- Raise error when an application from ECF is not found in NPQ Reg or if found but there's any discrepancies in attributes values when comparing the 2 records.

image

[CPDLP-2961]: https://dfedigital.atlassian.net/browse/CPDLP-2961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/migration/migrators/application.rb
```diff
@@ -0,0 +1,48 @@
...
+        application = ::Application.find_by!(ecf_id: ecf_npq_application.id)
+
+        compare_attributes_values!(ecf_npq_application, application)
+      end
+    end
+
+  private
+
+    def ecf_npq_applications
+      @ecf_npq_applications ||= Migration::Ecf::NpqApplication.joins(:participant_identity).all.limit(500)
```
ethax-ross, line 14: Why are we limiting the results here?

app/services/migration/migrators/application.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def attributes_to_compare
+      %i[
```
ethax-ross, line 18: As we're doing this a lot of times it would be more performant to move these to a frozen constant

spec/services/migration/migrators/application_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      it "calls FailureManager with correct params" do
+        expect_any_instance_of(Migration::FailureManager).to receive(:record_failure).with(ecf_npq_application1, "Validation failed: There are some discrepancies in one or more attributes values").and_call_original
+        expect_any_instance_of(Migration::FailureManager).to receive(:record_failure).with(ecf_npq_application2, "Validation failed: There are some discrepancies in one or more attributes values").and_call_original
+
+        subject
+      end
+    end
+
+    context "when a application is not correctly created" do
```
ethax-ross, line 49: `when an application exists in ECF but cannot be matched to an application in NPQ registration`?

spec/services/migration/migrators/application_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    before do
+      create(:application, ecf_id: ecf_npq_application1.id)
+      create(:application, ecf_id: ecf_npq_application2.id)
+
+      create(:data_migration, model: :application)
+    end
+
+    context "when all attributes values of interest matches" do
+      before do
+        allow_any_instance_of(Migration::Ecf::NpqApplication).to receive(:attributes).and_return({})
```
ethax-ross, line 21: This feels a bit heavy handed on the mocking as it will skip all the comparison logic in the migrator; would it be better to set the application attributes explicitly for the state where they should match?

ethax-ross (commented):
Looks good just a few minor comments/questions 👍

spec/services/migration/migrators/application_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  subject { instance.call }
+
+  describe "#call" do
+    let(:ecf_npq_application1) { create(:ecf_migration_npq_application) }
+    let(:ecf_npq_application2) { create(:ecf_migration_npq_application) }
+
+    before do
+      create(:application,
+             ecf_id: ecf_npq_application1.id,
```
ethax-ross, line 14: Might be able to do something like: 

```
create(:application, ecf_id: ecf_application1.id, *ecf_npq_application1.attributes.slice(described_class:: ATTRIBUTES_TO_COMPARE))
```

ethax-ross (commented):
Just one minor suggestion, but otherwise I think this is good to go once its been ran on the migration env 👏

ethax-ross (approved):
Looks good! happy for it to be merged down if the run on migration was a success

cwrw (commented):
thanks @leandroalemao for this, looks great, I think something we are missing is Applications in NPQ that don't exist in ECF erroring, but we can do this at a later iteration/ticket after we analyse the current reports

app/services/migration/migrators/application_not_in_ecf.rb
```diff
@@ -0,0 +1,15 @@
+module Migration::Migrators
+  class ApplicationNotInEcf < Base
+    def call
+      migrate(applications, :application_not_in_ecf) do |application|
```
cwrw, line 4: I was wondering if we could do those in the same migrator if possible, or do they have to be separate?

leandroalemao: Hi @cwrw I've tried to do all in the same migrator but couldn't find an easy way of doing it as we use `in_batches` in the migrator which accepts a `ActiveRecord::Relation` query but technically `Application` and `Migration::Ecf::NpqApplication` are different models, so decided to split in 2 migrators. Easier to test as well and see results.

cwrw (approved):
thanks @leandroalemao

PR #1305, opened 2024-04-18
Add application endpoint

ethax-ross:
### Context

We want to add the GET `/application/:id` endpoint to NPQ reg. This will involve populating the existing controllers and adding serializer/query services.

- Add rebranded_alternative_courses to Course

This is used when working out if an Application has been previously funded.

Improve test coverage in `Course` model.

-  Add funding status logic to Application

When we serialize an application we need to specify a `ineligible_for_funding_reason`, which we determine based on if the application has been previously funded and is currently eligible for funding.

Add `previously_funded?` and `ineligible_for_funding_reason` to `Application` (lifted from ECF).

- Add ApplicationSerializer

Ports the `ApplicationSerializer` and respective versions from ECF. The only field we can't currently return correctly is `schedule_identifier` as we haven't migrated schedules yet.

Instead of inheriting between serializers as we do in ECF we're leveraging Blueprinter views to be able to return version-specific fields in the serializers.

Move existing `StatementSerializer` out of the version-specific module.

- Add Applications::Query service

Add a query service to find applications, ordered by `created_at: :asc`, which mirrors the ECF query service.

Only the filtering used by the application endpoint has been implemented; the rest can be added as and when we add the endpoints that use them.

Remove `Find` service and replace usage with `Query`.

- Fix declaration spec file paths

These were incorrectly hitting the application endpoints.

- Default application factory to have ecf_id

We have this as a trait currently but it makes more sense for applications to get it by default as the vast majority will be in this state.

- Add application show endpoint

Add endpoint to retrieve an application. Rename specs to remove `_controller` suffix to match statements. Extract specs into a shared context to DRY up between versions/endpoints.

### Guidance to review

Best reviewed by-commit.

The GET `/applications` endpoint is in a follow PR as this already contains a lot of changes: #1306

The `previously_funded?` call in the serializer is not optimised and will result in a DB call for each application; we'll do this in a follow up PR.


spec/requests/api/v2/declarations_controller_spec.rb
```diff
@@ -2,25 +2,19 @@
 
 RSpec.describe API::V2::DeclarationsController, type: "request" do
   describe("index") do
-    before { api_get(api_v2_applications_path) }
+    before { api_get(api_v1_declarations_path) }
```
cwrw, line 5: should those be v2 declarations?

spec/requests/api/v3/declarations_controller_spec.rb
```diff
@@ -2,13 +2,19 @@
 
 RSpec.describe API::V3::DeclarationsController, type: "request" do
   describe("index") do
-    before { api_get(api_v3_declarations_path) }
+    before { api_get(api_v1_declarations_path) }
```
cwrw, line 5: should those be v3 declarations?

app/serializers/api/application_serializer.rb
```diff
@@ -0,0 +1,57 @@
+module API
+  class ApplicationSerializer < Blueprinter::Base
+    identifier :ecf_id, name: :id
+
+    class AttributesSerializer < Blueprinter::Base
```
cwrw, line 5: could we add the type to the serialiser `"npq_application"`?

cwrw: could we also order the fields similar to before except for the timestamps, happy for them to go last

cwrw (commented):
fantastic work, especially with cleaning up the serialisers, controllers and tests! Just a few very minor comments otherwise ✅

app/serializers/api/application_serializer.rb
```diff
@@ -0,0 +1,57 @@
...
+      field(:funding_choice)
+      field(:headteacher_status)
+      field(:school_urn) { |a| a.school&.urn }
+      field(:school_ukprn) { |a| a.school&.ukprn }
+      field(:works_in_school)
+      field(:email_validated) { true }
+      field(:lead_provider_approval_status, name: :status)
+      field(:targeted_delivery_funding_eligibility)
+      field(:eligible_for_funding)
+      field(:teacher_catchment) { |a| a.teacher_catchment.present? }
```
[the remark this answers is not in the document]
ethax-ross: Good catch 👍 it looks like there is an existing `inside_catchment?` method but its just for England; I'll add a separate method for UK catchment but not sure if they should be one and the same 🤔

app/serializers/api/application_serializer.rb
ethax-ross: I've added the `type` and tried to copy the ordering in ECF but a few attributes are defined twice in ECF so the ordering isn't quite clear, let me know if I haven't got it right.

app/services/applications/query.rb
```diff
@@ -0,0 +1,31 @@
...
+          :private_childcare_provider,
+          :itt_provider,
+        )
+
+      scope = scope.where(lead_provider:) if lead_provider.present?
+
+      scope.order(created_at: :asc)
+    end
+
+    def application(id:)
```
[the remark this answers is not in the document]
ethax-ross: How do you feel about supporting them interchangeably from a callee perspective? would this lead to more confusion 🤔 

https://github.com/DFE-Digital/npq-registration/pull/1305/commits/86fbb9975b721f3104c279a0bb325b6faf18d560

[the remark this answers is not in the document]
peteryates: I think internally we'll use `id` so maybe if they looked more [like this](https://github.com/DFE-Digital/npq-registration/pull/1311/files#diff-6444f0bb86900e702fad0a69e08e91a93df5dca42676b515bb8be929bf1fd29fR25)?

peteryates (approved):

PR #1309, opened 2024-04-19
Add human readable messages for validations

peteryates:
### Context

This change adds validation messages that will make sense in the user interface (when we build it). Setting the trend early should encourage developers to continue adding them as we implement more validation rules. This prevents Rails' default error messages from unintentionally appearing in the UI, and they don't work with the GOV.UK error summary because they don't include the field name and aren't very helpful to users.

- **Add validation messages for Statement month**
- **Add validation messages for Statement year**
- **Add validation messages for Statement ecf_id**
- **Add validation messages for Statement output_fee**
- **Make statement items validation message friendlier**
- **Add validation message for User full_name**
- **Add validation message for User email**
- **Add validation message for Admin full_name**
- **Add validation message for Admin email**

### DIFF
--- a/app/models/admin.rb
+++ b/app/models/admin.rb
@@ -1,5 +1,8 @@
 class Admin < ApplicationRecord
-  validates :full_name, presence: true, length: { maximum: 64 }
+  validates :full_name,
+            presence: { message: "Enter a full name" },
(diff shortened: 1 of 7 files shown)


leoapost (approved):
LGTM

PR #1311, opened 2024-04-19
Combine `Statements::Find` and `Statements::Query`

peteryates:
### Context

This PR combines the `Statements::Query` and `Statements::Find` classes as specified in [CPDLP-3006](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3006&text=find).

The `Statements::Find` class has been removed and references to it switched to point at `Statements::Query`.

The ability to query statements by state has been added and it works imilarly to cohort start years in that it can take a comma-separated list.

The `#statement` method has been updated to work with either an `ecf_id` (used by the API) or `id` (used internally).

- **Remove Statements::Find class**
- **Make statement method take ecf_id or id**
- **Allow Statements::Query to query by state**
- **Replace references to Statements Find with Query**

### Next steps

* [x] make the object take comma separated lists as well as arrays (which we're more likely to use internally). [Done](cf332e5c3525).
* ~~work out how we intend on querying `where column is null` as we're already using `nil` to signify the field isn't being searched on~~ - **I'll do this in a separate PR, it's more complicated**
* [x] make `.with_output_fee` scope optional - looks like it's needed for the API but probably not everywhere internally. [Done](2afa3d3e0e612231904).

[CPDLP-3006]: https://dfedigital.atlassian.net/browse/CPDLP-3006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/statements/query.rb
```diff
@@ -1,29 +1,45 @@
...
+    def extract_conditions(list)
+      case list
+      when String
+        list.split(",")
+      when Array
+        list.compact
+      else
+        list
+      end
+    end
```
peteryates, line 43: This could be moved somewhere communal for query objects to share once we have some more.

PR #1315, opened 2024-04-22
[CPDNPQ-1752] Styling fixes after frontend update

TBenedyk:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1752

- Tags should be sentence case.
- Questions should be h2.
- Questions should have medium text if there is an XL title above them.

Question text configuration was already set but was being overridden in `_radio_button_group.html.erb` because of the ordering of arguments. They also needed changing from h1 to h2.

**Before:**
Screenshot 2024-04-22 at 11 36 00

**After:**
Screenshot 2024-04-22 at 11 36 13

**Before:**
Screenshot 2024-04-22 at 11 36 50

**After:**
Screenshot 2024-04-22 at 11 36 38

**Before:**
Screenshot 2024-04-22 at 11 37 23

**After:**
Screenshot 2024-04-22 at 11 37 34

**Before:**
Screenshot 2024-04-22 at 11 45 27

**After:**
Screenshot 2024-04-22 at 11 45 06

**Before:**
Screenshot 2024-04-22 at 12 35 01

**After:**
Screenshot 2024-04-22 at 12 05 08

### DIFF
--- a/app/views/accounts/_course_details.html.erb
+++ b/app/views/accounts/_course_details.html.erb
@@ -50,12 +50,12 @@
         </dt>
         <dd class="govuk-summary-list__value">
           <% if application.lead_provider_approval_status.blank? || pending?(application) %>
-            <strong class="govuk-tag govuk-tag--yellow">Apply with provider</strong>
+            <%= govuk_tag(text: "Apply with provider", colour: "yellow") %>
             <p class="govuk-!-margin-top-2 govuk-!-margin-bottom-0"><%= I18n.t("provider_details.pending_status")  %></p>
           <% elsif accepted?(application) %>
-            <strong class="govuk-tag govuk-tag--green">Successful</strong>
+            <%= govuk_tag(text: "Successful", colour: "green") %>
           <% elsif rejected?(application) %>
-            <strong class="govuk-tag govuk-tag--red">Unsuccessful</strong>
+            <%= govuk_tag(text: "Unsuccessful", colour: "red") %>
           <% end %>
         </dd>
         <!-- This code is only written for review apps in order to update the external statuses -->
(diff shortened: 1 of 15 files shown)


peteryates:
Coincidentally I just raised [another PR](https://github.com/DFE-Digital/npq-registration/pull/1316#pullrequestreview-2014509265) that did some tag case fixing 😅

app/views/accounts/_course_details.html.erb
```diff
@@ -89,10 +89,10 @@
           </dt>
           <dd class="govuk-summary-list__value">
               <% if application.participant_outcome_state.capitalize == "Passed" %>
-                <strong class="govuk-tag govuk-tag--green">PASSED</strong>
+                <strong class="govuk-tag govuk-tag--green">Passed</strong>
```
peteryates, line 92: We can use govuk-components here, it's already installed and set up:

```suggestion
                <%= govuk_tag(text: "Passed", colour: "green") %>
```

TBenedyk: OK thanks I'll update them

peteryates (commented):
This looks great. I left a comment about using govuk_components.

The other thing is on the final screenshot I think having both the heading and subheading (legend) is unnecessary

image

Maybe one for @Dom-Collins to cast his eye over (can be addressed as a follow up)

app/views/accounts/_course_details.html.erb
peteryates: Cheers, [here's a link to the gem docs](https://govuk-components.netlify.app/components/tag/).

slawosz (approved):

peteryates (approved):
Looks great. @Dom-Collins - mind taking a look at the page I referenced above and seeing if it's worth tweaking in a follow up PR?

dalcJ:
@TBenedyk All looks good, thank you :)

PR #1316, opened 2024-04-22
Refactor tag helpers

peteryates:
### Context

We have a couple of spots where we're manually doing things which are already covered by govuk-components.

- **Capitalise 'Yes' and 'No' and use govuk_tag helper**
- **Replace other manual tag building with component**

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,10 +18,10 @@ module ApplicationHelper
   end
 
   def boolean_red_green_tag(bool, text = nil)
-    text ||= bool ? "YES" : "NO"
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -18,10 +18,10 @@ def npq_registration_link
   end
 
   def boolean_red_green_tag(bool, text = nil)
-    text ||= bool ? "YES" : "NO"
+    text ||= bool ? "Yes" : "No"
```
peteryates, line 21: Note, also fixing capitalisation here, see comment in #1310.

TBenedyk (approved):

PR #1325, opened 2024-04-25
Improve query object searching

peteryates:
### Context

The current implementation of the statements query object has the slight flaw that the default values for the keyword query arguments were `nil`, and `nil` is meaningful in ActiveRecord, which converts it to `is null`.

Instead if we use the `:ignore` symbol to mark arguments as 'omitted', it frees up `nil` for `null` queries.

This change also shifts the building of the scope to the initializer and allows us to just keep the scope attr instead of one for each argument.

Almost inadvertantly, this format combined with the `:ignore` pattern brings back chaining, which might be useful as we progress.

### Changes

- **Rename QueryStore to RegistrationQueryStore**
- **Add base Query class and put reusable logic there**
- **Introduce sentinel default values** (thanks @pmanrubia for the name!)

## Reviewer notes

* the first two commits are just preparatory, pay most attention to ff23bd649988cf060f8f5efe1b6714f439e2ab46
* if we're happy with the approach I'll rework the `Statements::Query` tests so the new methods are covered without loads of repetition


app/services/query.rb
```diff
@@ -0,0 +1,12 @@
+class Query
```
peteryates, line 1: Could possibly be `BaseQuery` or something I guess. Or even a module that's mixed in where needed.

app/services/statements/query.rb
```diff
@@ -1,32 +1,66 @@
 module Statements
   class Query < Query
-    attr_reader :lead_provider, :cohort_start_years, :updated_since, :state, :output_fee
+    attr_reader :scope
 
-    def initialize(lead_provider: nil, cohort_start_years: nil, updated_since: nil, state: nil, output_fee: true)
```
cwrw, line 5: maybe we can have a test to show that with this implementation we could pass in `nil` values for `null` queries?

app/services/statements/query.rb
```diff
@@ -1,32 +1,66 @@
...
+    def where_state_is(state)
+      return if state == :ignore
+
+      scope.merge!(Statement.with_state(state))
+
+      self
+    end
+
+    def where_output_fee_is(output_fee)
+      return if output_fee == :ignore
```
cwrw, line 59: since output fee will default to `true` do we need this check?

app/services/registration_query_store.rb
```diff
@@ -1,4 +1,4 @@
-class QueryStore
+class RegistrationQueryStore
```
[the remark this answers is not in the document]
peteryates: Missed that, done.

app/services/statements/query.rb
peteryates: Yeah I'm rewriting the tests to take account of the new `where_` methods. Then we just need to ensure `initialize` calls them, which should reduce doubling up.

They're kind of also testing Postgres at the moment which seems a bit odd.

app/services/query.rb
peteryates: Sounds good, module for now.

app/services/statements/query.rb
peteryates: If we want to query for statements where output fee is either `true` or `false` we'd need to override the default with `:ignore`.

Normally I'd expect `:ignore` to be the default but it looks like it's assumed to be `true` throughout the API so I guess it makes sense to make an exception.

app/services/statements/query.rb
peteryates: I just made them private until we might want to use them individually.

cwrw (approved):

PR #1326, opened 2024-04-25
[CPDNPQ-1786] explain funding eligibility more clearly

TBenedyk:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1786

### Better messaging around funding eligbility

- Updates funding content on registration summarypage.
- Remove negative contraction ‘doesn’t' and replace with ‘does not’.
- Updates content about funding during wizard

**Before**

Screenshot 2024-04-25 at 15 41 22

**After**
Screenshot 2024-04-25 at 15 41 30

**After:**
Screenshot 2024-04-25 at 15 14 24
Screenshot 2024-04-25 at 15 14 08

### DIFF
--- a/app/views/accounts/_eligible_for_funding_details.html.erb
+++ b/app/views/accounts/_eligible_for_funding_details.html.erb
@@ -5,7 +5,7 @@
   Contact <%= application.lead_provider.name %> to check if they can offer you a scholarship funded space, if you have not done so already.
 </p>
 
-<% def details_text %>
+<%= govuk_details(summary_text: "What if no funded spaces are available with my chosen provider?") do %>
   <p class="govuk-!-margin-bottom-2">You can:</p>
 
   <ul class="govuk-list govuk-list--bullet">
@@ -22,8 +22,3 @@
     </li>
   </ul>
 <% end %>
-
-<%= govuk_details(
-  summary_text: "What if no funded spaces are available with my chosen provider?",
-  text: details_text.html_safe
-) %>
--- a/app/views/accounts/_eligible_for_funding_details.html.erb
+++ b/app/views/accounts/_eligible_for_funding_details.html.erb
@@ -1,0 +1,31 @@
+<p class="govuk-!-margin-top-2 govuk-!-margin-bottom-0">
+  This means that if <%= application.lead_provider.name %> confirms that you have a scholarship funded space, you will not need to pay for the course fees.
+</p>
+<p class="govuk-!-margin-top-2">
+  Contact <%= application.lead_provider.name %> to check if they can offer you a scholarship funded space, if you have not done so already.
+</p>
+
+<details class="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
+      What if no funded spaces are available with my chosen provider?
+    </span>
+  </summary>
+  <div class="govuk-details__text">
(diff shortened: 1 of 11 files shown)


app/views/accounts/_eligible_for_funding_details.html.erb
```diff
@@ -0,0 +1,31 @@
+<p class="govuk-!-margin-top-2 govuk-!-margin-bottom-0">
+  This means that if <%= application.lead_provider.name %> confirms that you have a scholarship funded space, you will not need to pay for the course fees.
+</p>
+<p class="govuk-!-margin-top-2">
+  Contact <%= application.lead_provider.name %> to check if they can offer you a scholarship funded space, if you have not done so already.
+</p>
+
+<details class="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
```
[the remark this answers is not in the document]
TBenedyk: Ah should've checked there was a component for it. I went straight to the design system. Will remember for next time 👍 Have pushed an update

app/views/accounts/_eligible_for_funding_details.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      explore other payment options, including paying for the course yourself or checking
+      if your workplace or trust can cover the cost
+    </li>
+  </ul>
+<% end %>
+
+<%= govuk_details(
+  summary_text: "What if no funded spaces are available with my chosen provider?",
+  text: details_text.html_safe
+) %>
```
TBenedyk, line 29: I don't love doing it this way with a method in the view. I would've done it with a heredoc but it doesn't like `html_safe` on it. I saw it's done this way in other places so thought this was probably the best option.

peteryates, line 29: Ah, the component takes a block, this should work 🙂 

```suggestion
<%= govuk_details(summary_text: "What if no funded spaces are available with my chosen provider?") do %>
  <p class="govuk-!-margin-bottom-2">You can:</p>

  <ul class="govuk-list govuk-list--bullet">
    <li>
      register with this provider again at a later date (courses are run twice a year,
      except for the Early headship coaching offer which varies by provider and is available year-round)
    </li>
    <li>
      register with a different provider
    </li>
    <li>
      explore other payment options, including paying for the course yourself or checking
      if your workplace or trust can cover the cost
    </li>
  </ul>
<% end %>
```

peteryates: It just means that if/when the [Details component](https://design-system.service.gov.uk/components/details/) changes upstream we don't have to go and find all the hardcoded instances of it and fix them manually.

peteryates (approved):
Nice!

If ever you find yourself writing `govuk-` in a template it's definitely worth checking [govuk-components](https://govuk-components.netlify.app/) or [govuk-form-builder](https://govuk-form-builder.netlify.app/) first - there's a good chance they have you covered

TBenedyk:
> Nice!
> 
> If ever you find yourself writing `govuk-` in a template it's definitely worth checking [govuk-components](https://govuk-components.netlify.app/) or [govuk-form-builder](https://govuk-form-builder.netlify.app/) first - there's a good chance they have you covered

@peteryates I will do in future. I missed that there was a block syntax. Agree its much better 👍

PR #1330, opened 2024-04-26
Fix pagination

slawosz:


app/views/admin/reopening_email_subscriptions/index.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="govuk-grid-column-full">
     <%= render "admin/layout", title: "Email subscription list" %>
 
-    <h2 class="govuk-heading-l">Users who've requested emails about registration reopening (<%= @users.count %>)</h2>
+    <h2 class="govuk-heading-l">Users who've requested emails about registration reopening (<%= @all_users.count %>)</h2>
```
TBenedyk, line 5: I think for the html page you might be better off doing `@users.count(:all)` which should give the total count. It might mean you end up doing one less query. I could be wrong. Maybe check the logs both ways. Looks fine otherwise

slawosz: It's lazy loading (`ActiveRecord::Relation`), so it doesn't query unless its being called to return data or count. And `@all_users` is required for `CSV` export.

TBenedyk: I just thought this would mean pagy is doing its own offset query for pagination and then theres a second AR query for this. I thought maybe using pagy's `@users.count(:all)` would be more optimised. I know there has to be two variables with @users and @all_users for html and csv so Im not saying to remove @all_users.

I was just curious if doing it the other way made any difference to the way the queries were performed. I don't really know how pagy does things under the hood. Just thought it was worth checking. Theres a good chance it makes no difference.

TBenedyk (approved):

PR #1333, opened 2024-04-29
[CPDLP-3010] Add NPQ application csv endpoint

leandroalemao:
### Context

Ticket: [CPDLP-2965](https://dfedigital.atlassian.net/browse/CPDLP-3010)

After the applications endpoint has been added we need to add its CSV counterparts in v1 and v2 (see [documentation](https://manage-training-for-early-career-teachers.education.gov.uk/api-reference/reference-v1.html#api-v1-npq-applications-csv-get)).

- Add the `/api/v1/npq-applications.csv` endpoint with the relevant filters
- Add the `/api/v2/npq-applications.csv` endpoint with the relevant filters
- Update swagger documentation with the 2 new endpoints above.

API swagger docs versions with the new endpoints can be accessed via:
- [v1](https://npq-registration-review-1307-web.test.teacherservices.cloud/api/docs/v1)
- [v2](https://npq-registration-review-1307-web.test.teacherservices.cloud/api/docs/v2)


app/controllers/api/v2/applications_controller.rb
```diff
@@ -1,41 +1,5 @@
...
-
-      def application_params
-        params.permit(:ecf_id, filter: %i[cohort updated_since])
-      end
-
-      def to_json(obj)
-        ApplicationSerializer.render(obj, root: "data")
-      end
-    end
+    class ApplicationsController < V1::ApplicationsController; end
```
ethax-ross, line 3: I think we should avoid this kind of coupling in favour of duplication but happy to get a wider opinion

app/serializers/api/application_csv_serializer.rb
```diff
@@ -0,0 +1,102 @@
...
+        scope.each do |record|
+          csv << to_row(record)
+        end
+      end
+    end
+
+  private
+
+    def csv_headers
+      %w[
```
ethax-ross, line 26: Frozen const for these?

app/controllers/api/v2/applications_controller.rb
leandroalemao: Hi @ethax-ross, well api v1 & v2 are pretty much the same with minor diferences.. v3 can inherit most of v1 as well.. with the main diferences in the serializers & extra params. Could revert the changes and happy to get a wider opinion as well.

cwrw: I think I'd prefer keeping the duplication, as we agreed in the thread previously in the separation channel, even though they are similar we want to avoid this coupling especially when we deprecate versions. However I'm happy to see how this evolves as we go, and how we can tidy this up further as we build more endpoints with duplication initially in place. Let me know your thoughts :)

leandroalemao: Fair enough guys.. just have reverted in https://github.com/DFE-Digital/npq-registration/pull/1333/commits/bc7fbbe185096007c2c7e161b8aca2d67af2fcd3 👍 😄

public/api/docs/v3/swagger.yaml
```diff
@@ -445,6 +458,213 @@ components:
               - npq-leadership-spring
               - npq-specialist-autumn
               - npq-specialist-spring
+    ApplicationCsv:
```
cwrw, line 461: csv endpoints aren't in v3 could we remove it from here please?

public/api/docs/v2/swagger.yaml
```diff
@@ -353,7 +383,223 @@ components:
               nullable: true
               type: boolean
               example: true
-    IDAttribute:
-      description: The unique ID of the resource.
-      type: uuid
-      example: d0b4a32e-a272-489e-b30a-cb17131457fc
+    ApplicationsCsvResponse:
```
cwrw, line 386: not sure why this is rendering correctly, see https://npq-registration-review-1333-web.test.teacherservices.cloud/api/docs/v2

cwrw (approved):
looks great 🙌 just the one comment around the docs

cwrw (approved):
thanks @leandroalemao

PR #1336, opened 2024-04-29
[CPDLP-3009] change ecf_id in NPQ from text to uuid

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3009

* Migration to changed `ecf_id` to `uuid` type.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_04_22_135109) do
+ActiveRecord::Schema[7.1].define(version: 2024_04_29_195521) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "citext"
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -345,7 +345,7 @@
     t.string "email", default: "", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.text "ecf_id"
+    t.uuid "ecf_id"
```
cwrw, line 348: will this affect any indices? will it work ok with `nil` values? also can we test this on the migration env by deploying this just invase?

cwrw: also a few failing specs to consider :)

mooktakim: The indices should continue to work and it works with `nil` values. I've tested it locally.
Will run on migration env before prod merge.

db/schema.rb
```diff
@@ -204,7 +204,7 @@
     t.text "name", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.text "ecf_id"
```
cwrw, line 207: can we change the factory to set up `SecureRandom.uuid`?

cwrw (commented):
looks great, thanks Mook, just one last comment :)

db/schema.rb
mooktakim: Updated now 👍

cwrw (approved):
Thanks @mooktakim! After testing on the migration env and deploying there happy for this to go in :)

PR #1338, opened 2024-04-30
Clean up database migrations

ethax-ross:
[Jira-3008](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3008)

Remove migrations that alter data as these are causing issues and are no longer needed.

Update some migrations to be consistent with the db schema.

Remove unused `allow_closed_registration` as it doesn't appear to be referenced anywhere and doesn't exist in the migrations.


db/schema.rb
```diff
@@ -353,15 +352,14 @@
...
-    t.string "feature_flag_id"
     t.string "provider"
     t.string "uid"
     t.jsonb "raw_tra_provider_data"
+    t.string "feature_flag_id"
     t.boolean "get_an_identity_id_synced_to_ecf", default: false
     t.datetime "updated_from_tra_at", precision: nil
     t.string "trn_lookup_status"
     t.boolean "notify_user_for_future_reg", default: false
-    t.boolean "allow_closed_registration"
```
ethax-ross, line 364: We should manually drop this from the databases as it doesn't seem to be used/never had a migration in the first place

cwrw: @slawosz this seems recent, just confirming if this was a wrong commit at some point for closed registration?

db/migrate/20230802120824_create_teaching_primary_mathematics_course_to_courses.rb
```diff
@@ -1,6 +0,0 @@
-class CreateTeachingPrimaryMathematicsCourseToCourses < ActiveRecord::Migration[6.1]
```
cwrw, line 1: @TBenedyk ah this is the data migration for maths, before it gets deleted :D

cwrw (commented):
looks good, question around dropping the column as relates to closing registration, will wait for @slawosz

db/schema.rb
slawosz: It could be wrong commit, yes.

cwrw (approved):

PR #1339, opened 2024-04-30
Initial guidance page drafts

mark-avery-dfe:
### Context

[What has been changed?]

failures:


app/views/api/guidance/get_started.md
```diff
@@ -0,0 +1,49 @@
...
+
+Emails us [INSERT EMAIL LINK]if you do not already have this token. We'll send it via secure email. 
+
+The tokens do not expire. 
+
+[WARNING TEXT] Providers must not share tokens in publicly accessible documents or repositories. 
+
+### How to use an authentication token 
+
+Include an authentication token in all requests to the API by adding an <code>Authorization</code> request header (not as part of the URL) in the following format:
```
peteryates, line 19: You should be able to use backticks here instead of `<code>` tags.

```suggestion
Include an authentication token in all requests to the API by adding an `Authorization` request header (not as part of the URL) in the following format: 
```

mark-avery-dfe: It seems to depend on the context. For example, within tables the backticks don't work...

PR #1340, opened 2024-04-30
[CPDLP-3015] Improve application endpoint performance (`previously_funded?` N+1)

leandroalemao:
### Context

Ticket: [CPDLP-3015](https://dfedigital.atlassian.net/browse/CPDLP-3015)

In the current ECF implementation of `ineligible_for_funding_reason` to surface the value `previously_funded?` we had to optimise in the API applications endpoint [here in ECF](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/v3/npq_applications_query.rb#L34).

Fix `previously_funded?` N+1 query then improving the application endpoint performance.

[CPDLP-3015]: https://dfedigital.atlassian.net/browse/CPDLP-3015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -56,6 +56,10 @@ class Application < ApplicationRecord
   }
 
   def previously_funded?
+    # This is an optimization used by the API Applications::Query in order
+    # to speed up the bulk-retrieval of Applications.
+    return transient_previously_funded if respond_to?(:transient_previously_funded)
+
     @previously_funded ||= user.applications
       .where.not(id:)
       .where(course: course.rebranded_alternative_courses)
--- a/app/services/applications/query.rb
+++ b/app/services/applications/query.rb
@@ -6,15 +6,24 @@ module Applications
     attr_reader :scope, :sort
 
     def initialize(lead_provider: :ignore, cohort_start_years: :ignore, updated_since: :ignore, participant_ids: :ignore, sort: nil)
-      @scope = Application
-        .includes(
-          :course,
-          :user,
-          :school,
-          :cohort,
-          :private_childcare_provider,
-          :itt_provider,
-        )
+      # The subquery is an optimization so that we don't have to perform
+      # a separate query for each record as part of Application#previously_funded?
+      @scope = all_applications.select(
+        "applications.*",
+        "EXISTS(
+          WITH json_data(alt_courses) AS (VALUES ('#{ActiveRecord::Base.sanitize_sql(alternative_courses)}'::jsonb))
+          SELECT 1 AS one FROM applications AS a, json_data
+            WHERE a.id != applications.id AND
+                  a.user_id = applications.user_id AND
+                  a.eligible_for_funding = true AND
+                  a.lead_provider_approval_status = 'accepted' AND
+                  a.course_id IN (
+                    SELECT jsonb_array_elements_text(alt_courses->(applications.course_id::text))::bigint
+                    FROM json_data
+                  )
+            LIMIT 1
+        ) AS transient_previously_funded",
+      )
       @sort = sort
 
       where_lead_provider_is(lead_provider)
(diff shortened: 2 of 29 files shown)


app/services/applications/query.rb
```diff
@@ -11,15 +11,24 @@ def initialize(lead_provider: nil, cohort_start_years: nil, updated_since: nil,
...
-          :course,
-          :user,
-          :school,
-          :cohort,
-          :private_childcare_provider,
-          :itt_provider,
-        )
+      # The subquery is an optimization so that we don't have to perform
+      # a separate query for each record as part of Application#previously_funded?
+      scope = all_applications.select(
```
cwrw, line 16: I was wondering if there was a better way of modelling this data rather than doing it through a query, as we avoided porting this over initially due to the complexity and cost of the query, maybe we can look into possible options of modelling first before we fallback to this solution?

app/models/application.rb
```diff
@@ -18,6 +18,19 @@ class Application < ApplicationRecord
   belongs_to :cohort, optional: true
 
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
+  has_many :previously_funded_applications, lambda {
+    includes(:course).eligible_for_funding.accepted.where("EXISTS(
+    WITH json_data(alt_courses) AS (VALUES ('#{ActiveRecord::Base.sanitize_sql("{\"1\":[1],\"8\":[8,7],\"11\":[11],\"3\":[3],\"4\":[4],\"5\":[5],\"6\":[6],\"7\":[7,8],\"9\":[9],\"2\":[2],\"10\":[10]}")}'::jsonb))
```
peteryates, line 23: Let's revisit this next week. I don't think we should be relying on JSON in queries whenever possible.

config/initializers/courses.rb
```diff
@@ -1,6 +1,7 @@
 module Courses
```
peteryates, line 1: ... this should totally just be a page in the admin section.

app/models/application.rb
peteryates: Having had a think, my understanding of the query above is this:

1. each course is mapped to itself except `7` and `8`, which overlap because they are "Additional Support Offer for new headteachers" and "Early Headship Coaching Offer" respectively, which sound similar so you probably don't qualify for funding on one if you've done the other 
    ```
    {
      1: [1],
      2:[2],
      3:[3],
      4:[4],
      5:[5],
      6:[6],
      7:[7,8],
      8:[8,7],
      9:[9],
      10:[10],
      11:[11],
      }
    ```
2. if a user has done the course (or a similar one in the case of `7` and `8`) they can't be funded for it again, so this matches and the record is returned

The thing that springs to mind is we should be able to solve this cleanly with data.

image

In the schema we added `CourseGroup` - this might have been intended for something else but a model _like_ this, where we had a "Headship" group that "Additional Support Offer for new headteachers" and "Early Headship Coaching Offer" belonged to, we could write the query without a CTE, without hardcoded IDs, without JSON  and without a subquery.

cwrw: So this came out of trying to move away from our solution in ECF to write a query for this which might be confusing. The original query in ECF is the following:
```
participant_identity
      .npq_applications
      .where.not(id:)
      .where(npq_course: npq_course.rebranded_alternative_courses)
      .where(eligible_for_funding: true)
      .accepted
      .exists?
```
which is basically checking if we have accepted applications that have been previously funded for the same course for the same user.
@leandroalemao let's chat about how we can maybe have something in the db which maybe is a flag that's set at certain stages of acceptance (which can come later) which we can just read on the record itself to confirm if someone has been accepted or not, as @peteryates mentioned attempting to solve with data itself. Give me a shout so we can chat :)

app/services/applications/query.rb
peteryates: Yeah let's have a chat about this at some point.

cwrw (approved):

PR #1349, opened 2024-05-07
Fix missing translations

peteryates:
### Context

Some translations are missing and we're unaware. Hopefully running through the entire test suite with `config.i18n.raise_on_missing_translations` enabled should highlight a few.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -179,6 +179,13 @@ en:
       pural_title: "Your NPQ registrations"
 
   admin:
+    application_submissions:
+      ecf_sync_request_log:
+        sync_type:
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -494,3 +506,5 @@ en:
         trn: "Teacher reference number (TRN)"
         full_name: "Full name"
         national_insurance_number: "National Insurance number (optional)"
+
+        date_of_birth: What is your date of birth?
```
peteryates, line 510: @mattonfoot I _think_ this is the error message that has its translation missing error read out in the clip you sent me.

cwrw (approved):

PR #1354, opened 2024-05-09
Trigger review app - there is possible bug there

slawosz:


peteryates:
Pretty sure this is hitting the review app memory limits. Do we need _all_ the schools for review apps?

peteryates:
Closing this for now, re-open if necessary.

PR #1353, opened 2024-05-08
[CPDLP-3003] Review migrator performance

leandroalemao:
### Context

Ticket: [CPDLP-3003](https://dfedigital.atlassian.net/browse/CPDLP-3003)

Currently the migrator completes all table migrations however whilst looking at the performance on the CPU usage is quite high.

### Changes proposed in this pull request

Review/improve migrator performance.

[CPDLP-3003]: https://dfedigital.atlassian.net/browse/CPDLP-3003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1355, opened 2024-05-09
[CPDLP-3003] Review migrator performance

leandroalemao:
### Context

Ticket: [CPDLP-3003](https://dfedigital.atlassian.net/browse/CPDLP-3003)

Currently the migrator completes all table migrations however whilst looking at the performance on the CPU usage is quite high.

Review/improve migrator performance.

[CPDLP-3003]: https://dfedigital.atlassian.net/browse/CPDLP-3003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/npq_separation/migration/migrations_controller.rb
+++ b/app/controllers/npq_separation/migration/migrations_controller.rb
@@ -8,8 +8,7 @@ class NpqSeparation::Migration::MigrationsController < ApplicationController
   end
 
   def create
-    Migration::Migrator.prepare_for_migration
-    MigrationJob.perform_later
+    ActiveJob.perform_all_later([MigrationJob.new, MigrationJob.new])
 
     redirect_to npq_separation_migration_migrations_path
   end
--- a/app/helpers/migration_helper.rb
+++ b/app/helpers/migration_helper.rb
@@ -45,7 +45,7 @@ module MigrationHelper
                "green"
              end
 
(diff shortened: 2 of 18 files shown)


leandroalemao:
Had to close the other [PR](https://github.com/DFE-Digital/npq-registration/pull/1353) as the review app in there was failing to be created. Opening another PR seems to have fixed the issue.

app/controllers/npq_separation/migration/migrations_controller.rb
```diff
@@ -8,8 +8,7 @@ def index
   end
 
   def create
-    Migration::Migrator.prepare_for_migration
-    MigrationJob.perform_later
+    ActiveJob.perform_all_later([MigrationJob.new, MigrationJob.new])
```
cwrw, line 11: Can we chat about this tomorrow, as Im not sure I understand, are we starting migrations in parallel?

cwrw:
discussed in person with @leandroalemao: some migrators are dependant on each other while some are not. We would like to differentiate those as we cannot run them all in parallel right now. Since we don't have capacity to focus on this we can close this for now and I'll move the ticket to the backlog to revisit again later when we do. Thanks @leandroalemao :)

PR #1361, opened 2024-05-10
[CPDLP-3004] TS1 - Regression - Statement migrator no importing all s…

mooktakim:
…tatements

### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3004

- Updated statement migrator to correctly find statement based on `ecf_id`


cwrw (commented):
@mook can we add tests that confirm this won't be an issue again if there are multiple statements with the same month/year please?

cwrw (approved):

PR #1366, opened 2024-05-14
[CPDLP-3047] TS4 Add reject applications endpoint

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3047

* `Applications::Reject` service class added
* Endpoints added:
  * `/api/v1/npq-applications/:ecf_id/reject`
  * `/api/v2/npq-applications/:ecf_id/reject`
  * `/api/v3/npq-applications/:ecf_id/reject`

### DIFF
--- a/app/controllers/api/v1/applications_controller.rb
+++ b/app/controllers/api/v1/applications_controller.rb
@@ -25,7 +25,11 @@ module API
       def reject
         service = Applications::Reject.new(application:)
 
-        render_from_service(service)
+        if service.reject
+          render json: to_json(service.application)
+        else
+          render json: API::Errors::Response.from(service), status: :unprocessable_entity
+        end
       end
 
     private
--- a/app/controllers/api/v1/applications_controller.rb
+++ b/app/controllers/api/v1/applications_controller.rb
@@ -17,11 +17,16 @@ module API
       end
 
       def show
-        render json: to_json(applications_query.application(ecf_id: application_params[:ecf_id]))
+        render json: to_json(application)
       end
 
       def accept = head(:method_not_allowed)
-      def reject = head(:method_not_allowed)
+
+      def reject
+        service = Applications::Reject.new(application:)
+
+        render_from_service(service)
+      end
 
     private
 
@@ -29,6 +34,10 @@ module API
         conditions = { lead_provider: current_lead_provider, cohort_start_years:, updated_since: }
 
         Applications::Query.new(**conditions.compact)
+      end
+
+      def application
+        applications_query.application(ecf_id: application_params[:ecf_id])
       end
 
       def cohort_start_years
--- a/spec/requests/api/v1/applications_spec.rb
+++ b/spec/requests/api/v1/applications_spec.rb
@@ -52,29 +52,12 @@ RSpec.describe "Application endpoints", type: :request do
   end
 
   describe "POST /api/v1/npq-applications/:ecf_id/reject" do
-    let(:application) { create(:application, lead_provider: current_lead_provider) }
+    let(:service_class) { Applications::Reject }
 
-    context "when application is pending" do
-      it "returns successfully" do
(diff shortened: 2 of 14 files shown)


spec/requests/api/v1/applications_spec.rb
```diff
@@ -31,14 +31,35 @@ def create_resource(**attrs)
...
     specify { expect(response).to(be_method_not_allowed) }
   end
 
-  describe("reject") do
-    before { api_post(api_v1_application_reject_path(123)) }
+  describe "POST /api/v1/npq-applications/:ecf_id/reject" do
+    let(:application) { create(:application, lead_provider: current_lead_provider) }
 
-    specify { expect(response).to(be_method_not_allowed) }
+    context "when application is pending" do
```
ethax-ross, line 42: Worth extracting these to a shared context?

ethax-ross (approved):
Looks good 👏 just one suggestion

app/controllers/api/v1/applications_controller.rb
```diff
@@ -9,11 +9,16 @@ def index
...
+        render json: to_json(application)
       end
 
       def accept = head(:method_not_allowed)
-      def reject = head(:method_not_allowed)
+
+      def reject
+        service = Applications::Reject.new(application:)
+
+        render_from_service(service)
```
peteryates, line 20: If we want to wrap the `service.call` in a database transaction with some other stuff (say recording an event) would it make sense to do it like this - putting the render within the transaction feels a bit weird:

```suggestion
        Application.transaction do
          render_from_service(service)
          SomeEvent.create("foo")
        end
```

It might make more sense if the `call` was done within `#reject` instead:

```suggestion
        Application.transaction do
          service.call
          SomeEvent.create("foo")
        end
        
        render_from_service(service)
```

peteryates (commented):
Looks good to me. I think as per my comment I'd be a bit more comfortable with the _calling_ of the class done within the `#reject` though.

Also that means we don't need to line the services up with the same method name.

app/controllers/api/v1/applications_controller.rb
mooktakim: I would say the transaction and events should be part of the service and `service.call` method, instead of it being in the controller.

mooktakim: eg:
```ruby
    def call
      return self unless valid?
      Application.transaction do
          application.update!(lead_provider_approval_status: "rejected")
          SomeEvent.create("foo")
      end
      application
    end
```

mooktakim:
> Looks good to me. I think as per my comment I'd be a bit more comfortable with the _calling_ of the class done within the `#reject` though.
> 
> Also that means we don't need to line the services up with the same method name.

@peteryates 
What do you think about the following:
```ruby
      def reject
        service = Applications::Reject.new(application:)
        if service.reject
          render json: to_json(service.application)
        else
          render json: API::Errors::Response.from(service), status: :unprocessable_entity
        end
      end
```

spec/requests/api/v1/applications_spec.rb
cwrw: 👍  with following the context pattern we have established now for other services

app/controllers/api/v1/applications_controller.rb
cwrw: I prefer this too, makes it easier to test as well to avoid those being in the request spec

peteryates: > I would say the transaction and events should be part of the service and service.call method, instead of it being in the controller.

Yeah you're right, I misread 🤦🏽‍♂️

peteryates:
Yeah I think that's super clean and easy to follow.

peteryates (approved):

cwrw (approved):
👍

ethax-ross (approved):

PR #1368, opened 2024-05-15
[CPDLP-3046] Add accept applications endpoint

leandroalemao:
### Context

Ticket: [CPDLP-3046](https://dfedigital.atlassian.net/browse/CPDLP-3046)

We have the application endpoints already built now in the separation branch, we need to implement the application endpoints for providers.

Add the POST accept application in all API versions.

[CPDLP-3046]: https://dfedigital.atlassian.net/browse/CPDLP-3046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/routes.rb
```diff
@@ -124,7 +124,7 @@
         defaults format: :json do
           resources :applications, path: "npq-applications", only: %i[index show], param: :ecf_id do
             post :reject, path: "reject"
-            post :accept, path: "accept"
+            post :accept, path: "accept", on: :member
```
peteryates, line 127: I'm not sure `on: :member` is necessary here.

It just changes the route method from `accept_api_v3_application` to `api_v3_application_accept` and the param name from `:ecf_id` to `:application_ecf_id` - it doesn't actually affect the route itself.

config/routes.rb
```diff
@@ -147,7 +147,7 @@
     namespace :v2, defaults: { format: :json }, constraints: ->(_request) { Rails.application.config.npq_separation[:api_enabled] } do
       resources :applications, path: "npq-applications", only: %i[index show], param: :ecf_id do
         post :reject, path: "reject"
-        post :accept, path: "accept"
+        post :accept, path: "accept", on: :member
```
peteryates, line 150: Or here.

config/routes.rb
```diff
@@ -173,7 +173,7 @@
     namespace :v3, defaults: { format: :json }, constraints: ->(_request) { Rails.application.config.npq_separation[:api_enabled] } do
       resources :applications, path: "npq-applications", only: %i[index show], param: :ecf_id do
         post :reject, path: "reject"
-        post :accept, path: "accept"
+        post :accept, path: "accept", on: :member
```
peteryates, line 176: Here too.

app/services/applications/accept.rb
```diff
@@ -0,0 +1,121 @@
...
+      @cohort ||= application.cohort
+    end
+
+    def user
+      @user ||= application.user
+    end
+
+    def course
+      @course ||= application.course
+    end
```
mooktakim, line 42: Lets use `delegate` for these

app/controllers/api/base_controller.rb
```diff
@@ -22,5 +22,13 @@ def bad_request_response(exception)
       Sentry.capture_exception(exception)
       render json: { errors: API::Errors::Response.new(error: I18n.t(:bad_request), params: exception.message).call }, status: :bad_request
     end
+
+    def render_from_service(service)
+      if service.valid?
+        render json: to_json(service.call)
```
cwrw, line 28: let's touch base on this to keep it similar with @mooktakim [Reject](https://github.com/DFE-Digital/npq-registration/pull/1366) PR

app/services/api/errors/response.rb
```diff
@@ -21,6 +21,15 @@ def call
           detail: params,
         }]
       end
+
+      def self.from(service)
```
cwrw, line 25: same here re @mooktakim pr, whoever merges first I guess 😅

config/routes.rb
mooktakim: I think putting these actions inside `member` is the correct way as its a singular action on a resource.
Would prefer a block though:
```ruby
member do
  post :reject, path: "reject"
  post :accept, path: "accept"
end
```

spec/factories/courses.rb
peteryates, line 1: I have mixed feelings about this. We should really seed this data because it's essentially static.

[The courses are effectively hardcoded](https://github.com/DFE-Digital/npq-registration/blob/main/app/models/course.rb).

I wonder if it would make more sense for them to live in the code instead of the database (where their identifiers could be represented by an enumerated type).

cwrw: I think this is a really good question, although Im not keen on making those changes yet to this due to the moving parts in adding new courses now and separation, but after the dust settles after the current registration happy to revisit. 

Part of me feels like it would be nice for someone to do this in the admin interface, where they can add a course easily and decide which type of NPQ it is etc. vs having it in the code, but right now it is a bit messy I agree

config/routes.rb
peteryates: It's still a member route by default though, adding `member` just changes the id param from `:ecf_id` to `:application_ecf_id`.

peteryates: Happy either way really, but if we use `:member` it should apply to both.

leandroalemao: I think naming the id param as `:ecf_id` makes more sense as we have ecf_id in the line above for the index/show actions, so we reuse the same strong params method across all endpoints in the controller as well.

```resources :applications, path: "npq-applications", only: %i[index show], param: :ecf_id do```

```      
def application_params
   params.permit(:ecf_id, filter: %i[cohort updated_since])
end
```

File: app/serializers/api/application_serializer.rb [hunk 7]
```diff
@@ -38,6 +38,7 @@ class AttributesSerializer < Blueprinter::Base
           a.updated_at,
         ].compact.max
       end
+      field(:funded_place)
```
cwrw, line 41: ⭐

cwrw: actually can we move this over the timestamps if possible?

app/services/applications/accept.rb
```diff
@@ -0,0 +1,111 @@
...
+
+    def set_funded_place_on_npq_application!
+      return unless cohort&.funding_cap?
+
+      application.update!(funded_place:)
+    end
+
+    def eligible_for_funded_place
+      return if errors.any?
```
cwrw, line 95:
```suggestion
      return if errors.any?
      return unless cohort&.funding_cap?
```

spec/requests/api/v1/applications_spec.rb
```diff
@@ -46,9 +46,15 @@ def create_resource(**attrs)
   end
 
   describe("accept") do
```
mooktakim, line 48: Can you update the describe name to `POST /api/v1/npq-applications/:ecf_id/accept`, to follow the naming format.

mooktakim: same for v2, v3

mooktakim (approved):
👍

cwrw (approved):
thanks @leandroalemao

spec/factories/cohorts.rb
```diff
@@ -6,5 +6,17 @@
...
       Cohort.find_by(start_year:) || new(**attributes)
     end
+
+    trait :current do
+      start_year { Date.current.year - (Date.current.month < 9 ? 1 : 0) }
+    end
+
+    trait :next do
+      start_year { Date.current.year + (Date.current.month < 9 ? 0 : 1) }
+    end
```
peteryates, line 16: Should we just be explicit about years when building cohorts - this feels like it could lead to unpredictable behaviour.

Also I think this logic could be expressed a little more simply, eg:

```
    trait :next do
      start_year { Date.current.month < 9 ? Date.current.year : Date.current.year.succ
    end
```

cwrw (approved):

cwrw (approved):

PR #1373, opened 2024-05-16
[CPDLP-3048] Add change funding application endpoint

leandroalemao:
### Context

Ticket: [CPDLP-3048](https://dfedigital.atlassian.net/browse/CPDLP-3048)

We would like to add a specific endpoint to toggle an application funded place attribute value.

Add new API endpoint for all API versions: PUT api/v{1/2/3}/npq-application/{application-id}/change-funded-place​ which accepts the following request:

```
{​
  "data": {​
    "type": "npq-application-change-funded-place",​
    "attributes": {​
      "funded_place": true || false​
    }​
  }​
}
```

[CPDLP-3048]: https://dfedigital.atlassian.net/browse/CPDLP-3048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/applications/change_funded_place.rb
```diff
@@ -0,0 +1,52 @@
...
+module Applications
+  class ChangeFundedPlace
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :application
+    attribute :funded_place, :boolean
+
+    validates :application, presence: { message: I18n.t("application.missing_application") }
+    validate :funded_place_not_nil
```
ethax-ross, line 12: `validates presence`?

spec/services/applications/change_funded_place_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+             eligible_for_funding: true,
+             course:,
+             lead_provider:,
+             cohort:)
+    end
+    let(:course) { create(:course, :sl) }
+    let(:lead_provider) { create(:lead_provider) }
+    let(:cohort) { create(:cohort, :current, :with_funding_cap) }
+
+    context "when application funded_place is false" do
```
ethax-ross, line 29: Can move updating the application in these specs into the before block? It reads oddly to me being "it sets the funded place to true" and we then change it to false right after this 😅 this order makes more sense in my head:

```
context "when the application funded_place is false" do
  before do
    params.merge!(funded_place: true)
    application.update!(funded_place: false)
  end

  it { expect { service.change }.to change { application.reload.funded_place }.to(true) }
end
```

spec/support/shared_examples/api_change_application_funded_plance_endpoint_support.rb
```diff
@@ -0,0 +1,128 @@
...
+
+      it "returns error in response" do
+        api_put(path(application_id), params:)
+
+        expect(parsed_response).to be_key("errors")
+        expect(parsed_response.dig("errors", 0, "detail")).to eql("The entered '#/funded_place' is missing from your request. Check details and try again.")
+      end
+    end
+
+    context "when application is not accepted" do
```
ethax-ross, line 53: Are we over testing the change service here? the validation of the service is already tested in the service class specs so maybe mocking it out and having a single test that covers when there is an error it returns a 422 would suffice?

app/services/applications/change_funded_place.rb
```diff
@@ -0,0 +1,52 @@
...
+      errors.add(:application, I18n.t("application.cannot_change_funded_status_non_eligible"))
+    end
+
+    def funded_place_not_nil
+      errors.add(:application, I18n.t("application.missing_funded_place")) if funded_place.nil?
+    end
+
+    def cohort_has_funding_cap
+      return if errors.any?
+      return if funded_place.nil?
```
ethax-ross, line 46: Not sure I follow why we need these two checks in here?

```
return if errors.any?
return if funded_place.nil?
```

mooktakim (approved):

spec/support/shared_examples/api_change_application_funded_plance_endpoint_support.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "an API change application funded place endpoint" do
```
mooktakim, line 3: I just noticed the file name has spelling error: "plance" => "place"

spec/support/shared_examples/api_change_application_funded_plance_endpoint_support.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    it "responds with 200 and representation of the application" do
+      api_put(path(application_id), params:)
+
+      expect(response).to be_successful
+      expect(parsed_response.dig("data", "attributes", "funded_place")).to be_truthy
+    end
+
+    context "when funded_place is nil in the params" do
+      let(:params) { { data: { type: "npq-application-accept", attributes: { funded_place: nil } } } }
```
mooktakim, line 20: Should the type be "npq-application-change-funded-place"?

app/services/applications/change_funded_place.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    delegate :cohort, to: :application
+
+    def accepted_application
+      return if application.accepted?
+
+      errors.add(:application, I18n.t("application.cannot_change_funded_status_from_non_accepted"))
+    end
+
+    def eligible_for_funding
```
cwrw, line 37: I think we have an extra check added on the other pr if you don't mind adding it here please https://github.com/DFE-Digital/early-careers-framework/pull/4738/files#diff-9312911905848cf62a214e3cbc8cfd98f3308145ff4a9b6123d8cfe1d5e0cbc9R13

cwrw (approved):

app/services/applications/change_funded_place.rb
```diff
@@ -0,0 +1,51 @@
...
+      return unless funded_place
+      return if application.eligible_for_funding?
+
+      errors.add(:application, I18n.t("application.cannot_change_funded_status_non_eligible"))
+    end
+
+    def cohort_has_funding_cap
+      return if errors.any?
```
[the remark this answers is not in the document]
cwrw: we would like to avoid chaining too many errors for providers, it gets confusing if we surface all errors although the main attribute it missing in a request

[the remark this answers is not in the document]
ethax-ross: Ah that makes sense. One for another day, but it doesn't make sense to do it here IMO; this is likely to get accidentally removed at some point. I think it would be better implemented in `API::Errors::Response` if its a general rule.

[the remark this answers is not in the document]
cwrw: I think its not general sometimes we want to chain as it makes sense, and sometimes we don't. However happy to revisit if there is a nicer way of doing this after we add more actions

PR #1374, opened 2024-05-16
Seed the courses instead of relying on factories

peteryates:
### Context

This change puts things we expect to be in the database there from the start. The app won't work if these records aren't present so we should seed them.


db/seeds.rb
```diff
@@ -1,24 +1,26 @@
...
-Faker::Config.locale = "en-GB"
-
 def load_base_file(file)
   base_file = Rails.root.join("db", "seeds", "base", file)
 
   load(base_file)
 end
 
+Rails.logger.info("seeding add_courses.rb")
+load_base_file("add_courses.rb")
```
cwrw, line 11: a bit confused, why are we moving this before the env checks? also might be safer to at least have a prod check?

cwrw (commented):
I wonder if this could be quite confusing, as I would always expect factories to be set within or before the test suite to set up test data in rspec. I feel like seeds aren't intended for the test suite and I think we should keep them separate as we will be amending these as we go. Maybe a middle ground is to set the definition loader for courses in teh before suite, as I think it was set like this before?

peteryates:
Ok, I've moved the definition loader into `rails_helper.rb` too but that still feels a bit weird. Course records are needed for the app to function, they're even [hardcoded in the locales](https://github.com/DFE-Digital/npq-registration/blob/main/config/locales/en.yml#L275).

Additionally, using `initialize_with` to make factories find or create instead of just create doesn't feel right.

Finally, because we're doing a `before_suite` which doesn't tidy up after itself, we're now kind of seeding the database outside of the seeds and in RSpec's config, which also feels a bit wrong.

I don't think it's unreasonable to seed the test database with data we know we need and it's how Rails works by default.

leandroalemao:
A bit confusing to me.. as I'm used to use factories when creating test data.. not using a service class to load or prepare data for tests. Also using `Course.find_by` everywhere in tests seems a bit weird to me.. would prefer using FactoryBot.. `create(:course)`.

peteryates:
But it was calling `Course.find_by(identifier:)` in the tests, just doing it via `FactoryBot.create` and intercepting via `initialize_with`.

That, along with 'seeding' the database in the RSpec setup, is adding several layers of complication over the facts that the application won't work properly without those 11 course records.

I think these code smells are indicating we need to have a discussion about whether courses should be represented in code or data. Maybe one for next week?

cwrw:
> But it was calling `Course.find_by(identifier:)` in the tests, just doing it via `FactoryBot.create` and intercepting via `initialize_with`.
> 
> That, along with 'seeding' the database in the RSpec setup, is adding several layers of complication over the facts that the application won't work properly without those 11 course records.
> 
> I think these code smells are indicating we need to have a discussion about whether courses should be represented in code or data. Maybe one for next week?

I agree its a smell and something isn't right, but I think its best leave this till the dust settles a bit, we are at capacity and both teams are working on adding new courses for the new cohort, I'd like to keep what we have and we can definitely revisit after things calm down a bit in a couple of weeks hopefully

peteryates:
OK. I'm going to park this for now, let's revisit when we're under a little less pressure.

PR #1377, opened 2024-05-17
Ensure course identifiers are present and unique

peteryates:
### Context

I think the application would misbehave quite significantly if we had repeated course identifiers. In the longer term we need to have a think about how to decouple the data from the business logic, but for now let's just ensure they only exist once.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,5 +1,8 @@
 class Course < ApplicationRecord
   validates :name, presence: true
+  validates :identifier,
+            presence: { message: "Enter a identifier" },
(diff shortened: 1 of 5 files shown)


cwrw (commented):
looks good, we might need to cleanup the staging and separation envs as they have duplicate maths courses. I think this is coming from the seeds and might need a tidy up as I have the same issue locally. After that and the failure happy to have this 👍

peteryates:
I've removed the duplicate `npq-leading-primary-mathematics` courses, choosing the ones with 0 applications each time.

cwrw (approved):
🙌

PR #1380, opened 2024-05-17
[CPDLP-3049] TS4 - Add swagger docs for all three new action endpoint…

mooktakim:
…s in all versions

### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3049

* API doc specs for new accept/reject/change funded place endpoints


spec/support/shared_examples/api_change_application_funded_place_endpoint_documentation_support.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      parameter name: :params,
+                in: :body,
+                style: :deepObject,
+                required: false,
+                schema: {
+                  "$ref": "#/components/schemas/ApplicationChangeFundedPlaceRequest",
+                }
+
+      response "200", "The NPQ application being accepted" do
```
ethax-ross, line 26: The NPQ application after changing the funded place

spec/swagger_schemas/attributes/id.rb
```diff
@@ -1,5 +1,5 @@
 ID_ATTRIBUTE = {
   description: "The unique ID of the resource.",
-  type: :uuid,
+  type: :string,
```
ethax-ross, line 3: Why is this changing, out of interest?

spec/swagger_schemas/requests/application_accept_attributes_request.rb
```diff
@@ -0,0 +1,13 @@
+APPLICATION_ACCEPT_ATTRIBUTES_REQUEST = {
+  description: "A NPQ application acceptance request attributes",
+  type: :object,
+  required: %i[funded_place],
+  properties: {
+    funded_place: {
+      description: "This field indicates whether the application is funded",
```
ethax-ross, line 7: I think this should be an `id` attribute?

spec/swagger_helper.rb
```diff
@@ -54,6 +54,12 @@
           ApplicationResponse: APPLICATION_RESPONSE[version],
           ApplicationsResponse: APPLICATIONS_RESPONSE[version],
           Application: APPLICATION[version],
+          ApplicationAcceptRequest: APPLICATION_ACCEPT_REQUEST,
```
ethax-ross, line 57: Is it worth just nesting these into a single file (one for accept, one for reject and one for funded place)? I think having them split out when they aren't being shared across other schemas makes it harder to digest

spec/swagger_schemas/attributes/id.rb
leandroalemao: I used https://editor.swagger.io/ to validate the swagger.yaml file and there was an error on the type: :uuid which is not a valid type according to https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#data-types so had to add `type: :string, format: :uuid` instead.

ethax-ross (approved):
Looks good to me! 👍

PR #1384, opened 2024-05-20
[CPDLP-3072] Add rake task to create good test data for lead providers in separation environment

leandroalemao:
### Context

Ticket: [CPDLP-3072](https://dfedigital.atlassian.net/browse/CPDLP-3072)

We have some early bird providers which would like to test with us before the full set of endpoints are complete.

Add rake task to create good test data for lead providers in separation environment.

[CPDLP-3072]: https://dfedigital.atlassian.net/browse/CPDLP-3072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/data/statements.csv
```diff
@@ -0,0 +1,153 @@
+name,cohort,deadline_date,payment_date,output_fee
+
+January 2022,2021,2021-12-25,2022-1-25,true
```
cwrw, line 3: Although this is close to prod data in ECF I wonder if we can try to avoid csvs and services that read from them. Statements don't need to be exactly like prod in sandbox, and we can probably use a few random ones for each cohort, happy to chat through this

lib/tasks/valid_test_data_generators/applications_populater.rb
```diff
@@ -0,0 +1,141 @@
...
+        email: Faker::Internet.email(name:),
+        date_of_birth: Date.new(1990, 1, 1),
+        trn:,
+        trn_verified: true,
+        trn_auto_verified: true,
+        ecf_id: SecureRandom.uuid,
+        get_an_identity_id: SecureRandom.uuid,
+      )
+
+      application = Application.create!(
```
cwrw, line 68: discussed in person: thinking about using factories vs model itself, we can compare notes after

lib/tasks/valid_test_data_generators/applications_populater.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require "tasks/valid_test_data_generators/school_urn_generator"
+require "tasks/valid_test_data_generators/trn_generator"
+
+module ValidTestDataGenerators
```
cwrw, line 6: discussed in person, maybe we can move this to services instead, same with statement

lib/tasks/valid_test_data_generators/applications_populater.rb
```diff
@@ -0,0 +1,141 @@
...
+      application.reload
+    end
+
+    def reject_application(application)
+      Applications::Reject.new(application:).reject
+      application.reload
+    end
+
+    def create_school_with(urn:)
+      school = School.find_or_create_by!(urn:) do |s|
```
cwrw, line 133: discussed in person: reuse existing schools for now to avoid creating too many schools

lib/tasks/valid_test_data_generators/trn_generator.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+ALL_TRNS = (1111..9_999_999) unless defined?(ALL_TRNS)
+
+class TRNGenerator
```
ethax-ross, line 5: Could we not just get away with using Faker instead of rolling our own service? its unlikely we would hit the same TRN for the seed data using this:

```
Faker::Number.unique.number(digits: 7)
```

And if we did it would raise an error in Faker instead of some random spec failure (similar to your 'TRN available list exhausted' error):

https://github.com/faker-ruby/faker/blob/main/lib/helpers/unique_generator.rb#L22

lib/tasks/valid_test_data_generators/separation_data.rake
```diff
@@ -0,0 +1,21 @@
...
+  task :seed_statements_and_applications, %i[lead_provider_name cohort_start_year] => :environment do |_t, args|
+    return unless Rails.env.in?(%w[development separation])
+
+    lead_provider = LeadProvider.find_by(name: args[:lead_provider_name])
+    raise "LeadProvider not found: #{args[:lead_provider_name]}" if args[:lead_provider_name] && !lead_provider
+
+    cohort = Cohort.find_by(start_year: args[:cohort_start_year])
+    raise "Cohort not found: #{args[:cohort_start_year]}" if args[:cohort_start_year] && !cohort
+
+    (lead_provider ? [lead_provider] : nil || LeadProvider.all).each do |lp|
```
ethax-ross, line 14: `Array.wrap(lead_provider || LeadProvider.all).each do`

`Array.wrap(cohort || Cohort.all).each do`

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -0,0 +1,103 @@
...
+                                      cohort:,
+                                      course:,
+                                      headteacher_status: Application.headteacher_statuses.keys.sample,
+                                      lead_mentor: [true, false].sample,
+                                      eligible_for_funding: [true, false].sample,
+                                      itt_provider: IttProvider.currently_approved.order("RANDOM()").first)
+
+      methods = %i[accept_application reject_application]
+
+      return if [true, false].sample
```
ethax-ross, line 68: Feel free to ignore, but faker has nice methods for this that let you tweak the ratio:

```
Faker::Boolean.boolean(true_ratio: 0.2)
```

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      application = FactoryBot.create(:application,
+                                      :pending,
+                                      :with_random_work_setting,
+                                      school:,
+                                      lead_provider:,
+                                      user:,
+                                      cohort:,
+                                      course:,
```
cwrw, line 81: could we use a different course here, maybe a random course except for the one selected previously to avoid having multiple applications with the same course accepted?

spec/services/valid_test_data_generators/applications_populater_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+require "rails_helper"
+
+RSpec.describe ValidTestDataGenerators::ApplicationsPopulater do
+  let(:lead_provider) { create(:lead_provider) }
+  let(:cohort) { create(:cohort) }
+
+  before { allow(Rails).to receive(:env) { environment.inquiry } }
+
+  subject { described_class.new(lead_provider:, cohort:, number_of_participants: 50) }
```
cwrw, line 11: for testing purposes could we set this to 2 to avoid creating too many and slowing things down?

lib/tasks/valid_test_data_generators/trn_generator.rb
cwrw: that's a fair point, happy to only use faker unless we hit issues later. Maybe we use find or create in the service for that?

cwrw (commented):
looks good and tested locally, just a few minor comments otherwise 👍

cwrw (approved):
thanks @leandroalemao

ethax-ross (approved):
Looks good 👍

PR #1385, opened 2024-05-21
User pp50 to determine eligibility

slawosz:
https://dfedigital.atlassian.net/browse/CPDNPQ-1649


app/controllers/registration_wizard_controller.rb
```diff
@@ -33,6 +33,26 @@ def update
...
+      current_step: :get_an_identity_callback,
+      store: session["registration_store"],
+      params: {},
+      request:,
+      current_user: user,
+    )
+
+    redirect_to registration_wizard_show_path(wizard.next_step_path)
+  end
+
```
TBenedyk, line 55: What does this do?

config/initializers/pp50_insitituions.rb
```diff
@@ -0,0 +1,13 @@
...
+fe = CSV.table("config/data/pp50further_education.csv")
+fe_ukprn = fe.each_with_object({}) { |school, hash| hash[school[:ukprn].to_s] = true }
+ey = CSV.table("config/data/early-years-all-settings.csv")
+ey_ofsted_urn = ey.each_with_object({}) { |school, hash| hash[school[:ofsted_urn].to_s] = true }
+childminders = CSV.table("config/data/childminders.csv")
+childminders_urn = childminders.each_with_object({}) { |school, hash| hash[school[:ofsted_urn].to_s] = true }
+PP50_SCHOOLS_URN_HASH = pp50_schools
+PP50_FE_UKPRN_HASH = fe_ukprn
+EY_OFSTED_URN_HASH = ey_ofsted_urn
+CHILDMINDERS_OFSTED_URN_HASH = childminders_urn
```
TBenedyk, line 13: Why do it like this and not use SQL tables?

app/controllers/registration_wizard_controller.rb
slawosz: Helps to avoid interaction with GAI while testing. When you need to do lot of testing, speeds up a little :)

config/initializers/pp50_insitituions.rb
slawosz: This is static data, and I find it easier that way. In theory, we could add flags in the SQL, but this way is faster. It could be an improvement later and `pp50_institutions.rb` could become a base for script that loads data to SQL.

TBenedyk: Just feels like a lot of data to live in memory. I suppose it's fine this way for now.

app/controllers/registration_wizard_controller.rb
TBenedyk: I'll have to try it out!

TBenedyk (approved):

app/views/registration_wizard/ineligible_for_funding/_not_a_eligible_childminder.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<p class="govuk-body">
+  For example you could pay for the course fees yourself or ask your workplace or trust to:
+
+  <ul class="govuk-list govuk-list--bullet">
+    <li>cover the full cost of the NPQ</li>
+    <li>share the cost of the course with you</li>
+  </ul>
+</p>
+
+
```
peteryates, line 17: ✂️

app/views/registration_wizard/ineligible_for_funding/_not_a_pp50_institution.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<p class="govuk-body">
+  For example you could pay for the course fees yourself or ask your workplace or trust to:
+
+  <ul class="govuk-list govuk-list--bullet">
+    <li>cover the full cost of the NPQ</li>
+    <li>share the cost of the NPQ with you</li>
+  </ul>
+</p>
+
+
```
peteryates, line 18: ✂️

app/views/registration_wizard/ineligible_for_funding/_not_entitled_ey_institution.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <br>
+  This means that you would need to pay for the course another way.
+</p>
+<p class="govuk-body">
+  For example you could pay for the course fees yourself or ask your workplace or trust to:
+
+<ul class="govuk-list govuk-list--bullet">
+  <li>cover the full cost of the NPQ</li>
+  <li>share the cost of the NPQ with you</li>
+</ul>
```
peteryates, line 15:
```suggestion
  <ul class="govuk-list govuk-list--bullet">
    <li>cover the full cost of the NPQ</li>
    <li>share the cost of the NPQ with you</li>
  </ul>
```

config/initializers/pp50_insitituions.rb
peteryates: I echo @TBenedyk's thoughts here.

In addition to the additional memory we also need to keep an eye on the 'sustainability' of the project - if in 5 years time there are 0 devs, could it be operated?

Can we have a ticket near the top of the backlog to
* move this into the database
* update the GIAS importer to set any additional fields when it runs
* or, if the data isn't in the GIAS export add an admin page where a CSV can be uploaded and processed (maybe - open to any nicer solution than this)

peteryates: Also it feels a little funny that they're hashes only with `true`. Would an array not make more sense, then you can check with:

```ruby
"123456".in?(pp50_schools)
```

PR #1390, opened 2024-05-22
Fix a bug caused by differences in sorting

peteryates:
### Context

PostgreSQL sorts alphatetically but Ruby sorts by hex value. Fixed this and left a comment explaining _why_ we're sorting this way.


peteryates:
I actually think this is caused by me upgrading PostgreSQL and somehow my locales were affected, which changed sorting.

PR #1396, opened 2024-05-28
[CPDLP-3131] Add participant id changes table

leandroalemao:
### Context

Ticket: [CPDLP-3131](https://dfedigital.atlassian.net/browse/CPDLP-3131)

We will need to add the participant id changes table to surface the correct structure in the participants npq endpoint in the API.

Add participant id changes table with correct fields/validations.

[CPDLP-3131]: https://dfedigital.atlassian.net/browse/CPDLP-3131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/participant_id_change.rb
+++ b/app/models/participant_id_change.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class ParticipantIdChange < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :user
+  belongs_to :from_participant, class_name: "User"
+  belongs_to :to_participant, class_name: "User"
+end
--- a/app/models/participant_id_change.rb
+++ b/app/models/participant_id_change.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class ParticipantIdChange < ApplicationRecord
-  has_paper_trail
-
   belongs_to :user
   belongs_to :from_participant, class_name: "User"
   belongs_to :to_participant, class_name: "User"
--- a/spec/factories/applications.rb
+++ b/spec/factories/applications.rb
@@ -81,5 +81,13 @@ FactoryBot.define do
     trait :with_random_user do
       user { FactoryBot.build(:user, :with_random_name) }
     end
+
+    trait :with_participant_id_change do
+      after(:create) do |application|
+        user = application.user
+
+        FactoryBot.create(:participant_id_change, from_participant: user, user:)
+      end
+    end
   end
 end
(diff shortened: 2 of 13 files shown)


app/models/participant_id_change.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ParticipantIdChange < ApplicationRecord
+  has_paper_trail
```
cwrw, line 4: could we postpone adding papertrail for now, we will do a sweep in the end to add it to an audited list of tables if that's ok?

spec/factories/applications.rb
```diff
@@ -81,5 +81,13 @@
     trait :with_random_user do
       user { FactoryBot.build(:user, :with_random_name) }
     end
+
+    trait :with_participant_id_change do
+      after(:create) do |application|
+        user = application.user
+
+        FactoryBot.create(:participant_id_change, from_participant: user, user:)
```
cwrw, line 89: could we amend this to move from a previous other user to the current user? maybe something like this:
`FactoryBot.create(:participant_id_change, to_participant: user, user:)`

cwrw (approved):
looking good just some minor comments

peteryates:
Looks good.

A couple of questions.

I'm slightly confused about how we use/surface this data, and does tracking the entire history of changes introduce querying complexity?

i.e. if a participant's changed more than once to get all their previous ids we'd need a recursive query:

```mermaid
flowchart TD
    123 --> 234
    234 --> 345
```

Like here, `345` used to be both `123` and `234`.

Would we be at a disadvantage if we just saved the `former_participant_ids` in an array?

```yaml
Participant:
  id: 345
  former_ids: [123, 234]
```

Do we need to know when the changes happened?

cwrw:
> i.e. if a participant's changed more than once to get all their previous ids we'd need a recursive query

We link everything to the current user, so we just have to look at the current user and list the participant id changes 
```
object.participant_id_changes
```

> Would we be at a disadvantage if we just saved the former_participant_ids in an array?

We do surface the `changed_at` in serialisers so we can't lose that data unfortunately

peteryates:
Ok - I missed that but that means we have 3 foreign keys on `participant_id_changes` pointing at the `users.id`, and two end with `participant_id`.

It wasn't clear to me what they meant and there's no other mention of 'participant' in the database.

cwrw:
> Should it be user_id_changes?
Good question, I think we went with the participant in the name as it maps identically to what's in the API and avoids confusion for us.

PR #1398, opened 2024-05-29
[CPDLP-3079] Update Application table with fields for participant profile from ECF

leandroalemao:
### Context

Ticket: [CPDLP-3079](https://dfedigital.atlassian.net/browse/CPDLP-3079)

We will need to add the participant profile fields to applications as we agreed to merge both tables.

Update `application` table with fields for participant profile from ECF and add `application_states` table that will work as the existing `participant_profile_states` from ECF.

[CPDLP-3079]: https://dfedigital.atlassian.net/browse/CPDLP-3079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/application_state.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class ApplicationState < ApplicationRecord
+  belongs_to :application, touch: true
```
cwrw, line 4: could we not include the touch here? We're going to rethink the whole touch model :)

cwrw (approved):
just the one comment otherwise 👍

PR #1402, opened 2024-05-31
[CPDLP-3014] TS5 - Add schedules table

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3014

* New `CourseGroup` model
* New `Schedule` model
* Removed `schedule_` from column names, eg: `applies_from` instead of `schedule_applies_from`, add your thoughts.
* Added seed files
* Updated tests


app/models/course_group.rb
```diff
@@ -0,0 +1,5 @@
+class CourseGroup < ApplicationRecord
+  has_many :courses
+
+  validates :name, presence: true, uniqueness: true
```
peteryates, line 4: Could we have some error messages on the validations, even if it's a generic "Enter a unique course group name"? Just to prevent the uninformative Rails defaults ("is taken") popping up when we build the admin UI.

spec/factories/course_groups.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :course_group do
+    sequence(:name) { |n| "Course Group #{n}" }
+
+    initialize_with do
```
peteryates, line 5: Not a fan of this as it's making `FactoryBot.create` a bit misleading. I'd rather see some `#find_by` in the specs really - no harm in relying on seeds in specs.

peteryates:
Looking good overall, left a couple of really minor comments.

app/models/course_group.rb
mooktakim: Added error message

spec/factories/course_groups.rb
mooktakim: Removed `initialize_with` and added `find_by` 👍

app/services/migration/migrators/course.rb
```diff
@@ -1,11 +1,29 @@
...
+      "npq-ehco-november" => "ehco",
+      "npq-aso-november" => "support",
+      "npq-aso-december" => "support",
+      "npq-aso-march" => "support",
+      "npq-aso-june" => "support",
+    }.freeze
+
     def call
       migrate(ecf_npq_courses, :course) do |ecf_npq_course|
+        course_group = ::CourseGroup.find_or_initialize_by(name: COURSE_GROUPS[ecf_npq_course.identifier])
```
cwrw, line 20: I think those are not course identifiers but schedules unless I'm mistaken?

db/seeds/base/add_schedules.rb
```diff
@@ -0,0 +1,39 @@
...
+      schedule.applies_from = row["schedule-applies-from"]
+      schedule.applies_to = row["schedule-applies-to"]
+      schedule.allowed_declaration_types = row["declaration-types"].split("|")
+
+      schedule.save!
+    end
+  end
+
+  def rows
+    @rows ||= CSV.read(
```
cwrw, line 31: Is it possible to avoid using CSVs in NPQ reg? We're hoping to move away from this setup and do most cohort set up in the admin, so I was thinking is there an easier way we can create schedules using factories instead? I think we did something similar in statements?

spec/factories/schedules.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.define do
+  factory :schedule do
+    course_group
+    cohort
+
+    sequence(:name) { |n| "Schedule #{n}" }
+    sequence(:identifier) { |n| "schedule-#{n}" }
+
+    applies_from { nil }
```
cwrw, line 9: aren't those applies from/to required 🤔

app/models/course.rb
```diff
@@ -1,4 +1,6 @@
 class Course < ApplicationRecord
+  belongs_to :course_group
```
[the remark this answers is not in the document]
mooktakim: I've set it to optional

app/services/migration/migrators/course.rb
mooktakim: Ah yes. I'll remove this for now.

cwrw (approved):
thanks @mooktakim 🙌

PR #1403, opened 2024-06-03
[CPDLP-3080] Add NPQ participant profiles endpoint

leandroalemao:
### Context

Ticket: [CPDLP-3080](https://dfedigital.atlassian.net/browse/CPDLP-3080)

We would like to add the NPQ participant profiles endpoint to NPQ separation

- Add new API endpoint for all API versions: `GET api/v{1/2/3}/participants/npq​`.

[CPDLP-3080]: https://dfedigital.atlassian.net/browse/CPDLP-3080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v1/participants_controller.rb
+++ b/app/controllers/api/v1/participants_controller.rb
@@ -1,13 +1,35 @@
 module API
   module V1
     class ParticipantsController < BaseController
-      def index = head(:method_not_allowed)
+      include Pagination
+      include ::API::Concerns::FilterByUpdatedSince
+
+      def index
+        render json: to_json(paginate(participants_query.participants))
+      end
+
       def show = head(:method_not_allowed)
       def change_schedule = head(:method_not_allowed)
       def defer = head(:method_not_allowed)
       def withdraw = head(:method_not_allowed)
       def resume = head(:method_not_allowed)
       def outcomes = head(:method_not_allowed)
+
+    private
+
+      def participants_query
+        conditions = { lead_provider: current_lead_provider, updated_since: }
+
+        ::Participants::Query.new(**conditions.compact)
+      end
+
+      def participant_params
+        params.permit(filter: %i[updated_since])
+      end
+
+      def to_json(obj)
+        ParticipantSerializer.render(obj, view: :v1, root: "data", lead_provider: current_lead_provider)
+      end
     end
   end
 end
--- a/app/serializers/api/participant_serializer.rb
+++ b/app/serializers/api/participant_serializer.rb
@@ -1,0 +1,105 @@
+module API
+  class ParticipantSerializer < Blueprinter::Base
+    identifier :ecf_id, name: :id
+    field(:type) { "npq-participant" }
+
+    class AttributesSerializer < Blueprinter::Base
+      exclude :id
+
+      field(:email)
+      field(:full_name)
+      field(:trn, name: :teacher_reference_number)
+      field(:updated_at)
+
+      view :v1 do
+        field(:npq_courses) do |object, options|
+          applications(object, options).map { |application| application.course.identifier }
+        end
+
+        field(:funded_places) do |object, options|
+          applications(object, options).map do |application|
+            {
+              npq_course: application.course.identifier,
+              funded_place: application.funded_place,
+              npq_application_id: application.ecf_id,
+            }
+          end
+        end
+      end
+
+      view :v2 do
+        field(:npq_enrolments) do |object, options|
+          applications(object, options).map do |application|
+            {
+              course_identifier: application.course.identifier,
+              # TODO: Add when schedules are implemented
+              schedule_identifier: nil,
+              cohort: application.cohort&.start_year&.to_s,
+              npq_application_id: application.ecf_id,
+              eligible_for_funding: application.eligible_for_funding,
+              training_status: application.training_status,
+              school_urn: application.school&.urn,
+              targeted_delivery_funding_eligibility: application.targeted_delivery_funding_eligibility,
+              funded_place: application.funded_place,
+            }
+          end
+        end
+      end
+
+      view :v3 do
+        field(:npq_enrolments) do |object, options|
+          applications(object, options).map do |application|
+            {
+              email: object.email,
+              course_identifier: application.course.identifier,
+              # TODO: Add when schedules are implemented
+              schedule_identifier: nil,
+              cohort: application.cohort&.start_year&.to_s,
+              npq_application_id: application.ecf_id,
+              eligible_for_funding: application.eligible_for_funding,
+              training_status: application.training_status,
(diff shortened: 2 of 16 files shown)


config/application.rb
```diff
@@ -29,6 +29,7 @@ class Application < Rails::Application
 
     # don't use AJAX/XHR to submit forms by default
     config.action_view.form_with_generates_remote_forms = false
+    config.action_controller.always_permitted_parameters = %w( controller action format )
```
cwrw, line 32: curious why we needed to amend this?

app/services/participants/query.rb
```diff
@@ -0,0 +1,65 @@
+module Participants
+  class Query
+    include API::Concerns::Orderable
+
+    attr_reader :scope, :sort
+
+    def initialize(lead_provider: :ignore, updated_since: :ignore, training_status: :ignore, from_participant_id: :ignore, sort: nil)
```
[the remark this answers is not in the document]
cwrw: actually probably keep them separate, but we need this to only look at accepted applications

app/services/participants/query.rb
```diff
@@ -50,7 +50,7 @@ def order_by
 
     def all_participants
       User
-        .joins(:applications)
+        .joins(:applications).merge(Application.accepted)
```
cwrw, line 53: could we add a test that it doesn't surface pending/rejected applications?

spec/factories/users.rb
```diff
@@ -23,13 +23,13 @@
       trn_verified { true }
     end
 
-    trait :with_application do
+    factory :participant do
```
cwrw, line 26: I'm wondering about the naming, can we avoid participant factories possibly, as it's not a concept in NPQ, happy to chat about other options

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,103 @@
+module API
+  class ParticipantSerializer < Blueprinter::Base
```
cwrw, line 2: can we add a spec for this please?

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,103 @@
...
+    identifier :ecf_id, name: :id
+    field(:type) { "npq-participant" }
+
+    class AttributesSerializer < Blueprinter::Base
```
cwrw, line 6: could you add the `participant_id` please?

cwrw: also could we ensure the order is similar to ecf please if possible with timestamps at the end?

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,103 @@
...
+              training_status: application.training_status,
+              school_urn: application.school&.urn,
+              targeted_delivery_funding_eligibility: application.targeted_delivery_funding_eligibility,
+              funded_place: application.funded_place,
+            }
+          end
+        end
+      end
+
+      view :v3 do
```
cwrw, line 48: in v3 email is only in the enrolments not in the higher level attributes, if you can please amend that, thanks

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,103 @@
...
+              email: object.email,
+              course_identifier: application.course.identifier,
+              schedule_identifier: application&.schedule&.identifier,
+              cohort: application.cohort&.start_year&.to_s,
+              npq_application_id: application.ecf_id,
+              eligible_for_funding: application.eligible_for_funding,
+              training_status: application.training_status,
+              school_urn: application.school&.urn,
+              targeted_delivery_funding_eligibility: application.targeted_delivery_funding_eligibility,
+              withdrawal: nil,
```
cwrw, line 61: could we add the withdrawal and deferral objects please?

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,103 @@
...
+              from_participant_id: participant_id_change.from_participant.ecf_id,
+              to_participant_id: participant_id_change.to_participant.ecf_id,
+              changed_at: participant_id_change.created_at.rfc3339,
+            }
+          end
+        end
+      end
+
+      def self.applications(object, options)
+        scope = object.applications
```
cwrw, line 81: there are a few n+1s in the API responses could we make sure the API responses are performant? Also I was wondering if we can avoid doing extra queries here and instead do something similar like in ECF with checking the lead provider is equal without a query?

app/controllers/api/v1/participants_controller.rb
```diff
@@ -1,13 +1,35 @@
...
       def withdraw = head(:method_not_allowed)
       def resume = head(:method_not_allowed)
       def outcomes = head(:method_not_allowed)
+
+    private
+
+      def participants_query
+        conditions = { lead_provider: current_lead_provider, updated_since: }
+
+        ::Participants::Query.new(**conditions.compact)
```
mooktakim, line 23: There's no need to separate the params for `Query`, its clearer to just:
```ruby
::Participants::Query.new(lead_provider: current_lead_provider, updated_since:)
```

app/models/application.rb
```diff
@@ -62,7 +62,7 @@ class Application < ApplicationRecord
     active: "active",
     deferred: "deferred",
     withdrawn: "withdrawn",
-  }
+  }, _prefix: "training_status"
```
cwrw, line 65: thinking about this, I think keeping it without a prefix is preferable as the actions are quite clear, but happy to chat through this

cwrw (approved):
Looks great, just the one comment around removing the prefix, happy to chat through that :)

app/controllers/api/v1/participants_controller.rb
leandroalemao: Yeah I know I'm just trying to follow the pattern already being used in the other controllers.

cwrw (approved):

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,146 @@
...
+      class << self
+        def applications(object, options)
+          return Application.none unless options[:lead_provider]
+
+          object.applications.select { |application| application.lead_provider_id == options[:lead_provider].id }
+        end
+
+        def withdrawal(application:, lead_provider:)
+          if application.withdrawn?
+            # We are doing this in memory to avoid running those as queries on each request
```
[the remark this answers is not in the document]
cwrw: Chatted in person: good point, I'll create a tech debt ticket for this, we can pick up next sprint and for this pr we can leave for now, thanks Pete :)

PR #1413, opened 2024-06-05
[CPDNPQ-1831] Funding eligibility rules for other

TBenedyk:
### Context

https://dfedigital.atlassian.net/browse/CPDNPQ-1831

The funding rules for registrations with ‘other’ as a setting are changing. Up til now all registrations using this route were given a ‘in review’ funding outcome, because the policy team had to double check information about them. But some of these will now have a funding outcome given during registration. 

See the ticket for the specific funding eligibility rule changes required.


app/helpers/funding_helper.rb
```diff
@@ -9,6 +9,11 @@ def scholarship_funding_eligibility(application)
   end
 
   def scholarship_eligibility_in_review?(application)
+    return false if application.eligible_for_funding
+    return false if !application.eligible_for_funding && application.funding_choice.present?
+    return false if application.employment_type == "other"
+    return true if application.teacher_catchment == "england" && application.course.ehco? && new_headteacher?(application)
```
slawosz, line 15: Could `application.teacher_catchment == "england"` be separate check like `return false if application.teacher_catchment != "england"` and maybe we could have method like `application.teacher_catchment_in_england?`?

TBenedyk: Updated

slawosz (approved):

PR #1421, opened 2024-06-07
Fix deprecation warning with Rails7 usage of ActiveSupport::Deprecation singleton

leandroalemao:
### Context

Ticket: N/A

Rails 7 deprecated ActiveSupport::Deprecation singleton usage.
The gem `rails_semantic_logger`[ is still using it](https://github.com/reidmorrison/rails_semantic_logger/blob/93fed9edd0ba21490044e5fc7474c7b83eb2676c/lib/rails_semantic_logger/extensions/action_dispatch/debug_exceptions.rb#L10). So a warning is displayed when running tests locally.

```Rails -- DEPRECATION WARNING: Calling silence on ActiveSupport::Deprecation is deprecated and will be removed from Rails (use Rails.application.deprecators.silence instead)```

Create a patch to fix the issue using `Rails.application.deprecators.silence` instead.

### DIFF
--- a/lib/rails_semantic_logger/extensions/action_dispatch/debug_exceptions.rb
+++ b/lib/rails_semantic_logger/extensions/action_dispatch/debug_exceptions.rb
@@ -1,0 +1,15 @@
+# Log actual exceptions, not a string representation
+require "action_dispatch"
+
+module ActionDispatch
+  class DebugExceptions
+  private
+
+    undef_method :log_error
+    def log_error(_request, wrapper)
+      Rails.application.deprecators.silence do
+        ActionController::Base.logger.fatal(wrapper.exception)
+      end
+    end
+  end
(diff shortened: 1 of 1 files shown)


peteryates:
Hmm it's not very reassuring when the author of the gem [isn't responsive to fixes for this](https://github.com/reidmorrison/rails_semantic_logger/pull/204).

I made the switch on the advice from the platform team, perhaps there's a more modern (supported) alternative?

leandroalemao:
Yeah I had seen that PR.. that's why decided to create the patch.. not sure when the author of the gem will get it fixed. Maybe we could fork the repo, add a fix/customise then use the forked version?

leandroalemao:
Reopening another PR for this as this one seems to have a broken build.

PR #1418, opened 2024-06-06
[CPDLP-3074] TS5 - Allow schedule change on acceptance

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3074

### Changes proposed in this pull request

* Updated `/api/v3/npq-applications/{id}/accept` to allow schedule change
* Updated `Applications::Accept` service class
* Updated tests
* Updated swagger

PR #1422, opened 2024-06-07
[CPDLP-3083] Add seed data for participant profiles/accepted applications

leandroalemao:
### Context

Ticket: [CPDLP-3083](https://dfedigital.atlassian.net/browse/CPDLP-3083)

We've added accepted applications as well as funded place logic we need to ensure that the seed data for review environments is up to date with this data.

Add seed data for participant profiles/accepted applications.

[CPDLP-3083]: https://dfedigital.atlassian.net/browse/CPDLP-3083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -9,4 +9,17 @@ class Schedule < ApplicationRecord
 
   validates :applies_from, presence: true
   validates :applies_to, presence: true
+
+  def self.default_for(course_group:, cohort: Cohort.current)
+    case course_group.name
+    when "specialist"
+      find_by!(cohort:, identifier: "npq-specialist-spring")
+    when "leadership"
+      find_by!(cohort:, identifier: "npq-leadership-spring")
+    when "support"
+      find_by!(cohort:, identifier: "npq-aso-december")
+    when "ehco"
+      find_by!(cohort:, identifier: "npq-ehco-june")
+    end
+  end
 end
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -9,17 +9,4 @@ class Schedule < ApplicationRecord
 
   validates :applies_from, presence: true
   validates :applies_to, presence: true
-
-  def self.default_for(course_group:, cohort: Cohort.current)
-    case course_group.name
-    when "specialist"
-      find_by!(cohort:, identifier: "npq-specialist-spring")
-    when "leadership"
-      find_by!(cohort:, identifier: "npq-leadership-spring")
-    when "support"
-      find_by!(cohort:, identifier: "npq-aso-december")
-    when "ehco"
-      find_by!(cohort:, identifier: "npq-ehco-june")
-    end
-  end
 end
--- a/db/seeds/base/add_applications.rb
+++ b/db/seeds/base/add_applications.rb
@@ -21,7 +21,7 @@ LeadProvider.find_each do |lead_provider|
     FactoryBot.create_list(
       :application,
       4,
-      :with_random_lead_provider_approval_status,
+      %i[accepted rejected].sample,
       :with_random_participant_outcome_state,
       :with_random_work_setting,
       user: FactoryBot.create(:user, :with_random_name),
@@ -34,11 +34,11 @@ LeadProvider.find_each do |lead_provider|
     FactoryBot.create_list(
       :application,
       4,
+      :accepted,
       :with_random_user,
       :with_random_work_setting,
       lead_provider:,
       course: Course.all.sample,
-      lead_provider_approval_status: "accepted",
       participant_outcome_state: "passed",
       cohort: Cohort.all.sample,
     )
@@ -47,11 +47,11 @@ LeadProvider.find_each do |lead_provider|
     FactoryBot.create_list(
       :application,
       4,
+      :rejected,
       :with_random_user,
       :with_random_work_setting,
       lead_provider:,
       course: Course.all.sample,
-      lead_provider_approval_status: "rejected",
       participant_outcome_state: "failed",
       cohort: Cohort.all.sample,
     )
@@ -61,9 +61,9 @@ LeadProvider.find_each do |lead_provider|
       :application,
       4,
       :deferred,
+      %i[accepted rejected].sample,
       :with_random_user,
       :with_random_work_setting,
-      :with_random_lead_provider_approval_status,
       :with_random_participant_outcome_state,
       lead_provider:,
       course: Course.all.sample,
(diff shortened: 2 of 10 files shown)


db/seeds/base/add_applications.rb
```diff
@@ -21,7 +21,7 @@
     FactoryBot.create_list(
       :application,
       4,
-      :with_random_lead_provider_approval_status,
+      %i[accepted rejected].sample,
```
cwrw, line 24: can we keep those as previously set as that trait is doing exactly that?

spec/services/applications/accept_spec.rb
```diff
@@ -251,7 +251,7 @@
     end
 
     describe "NPQ capping" do
-      let(:cohort) { create(:cohort, :current, :with_funding_cap) }
+      let(:cohort) { create(:cohort, :next, :with_funding_cap) }
```
cwrw, line 254: I think it's ok to keep to current for this and below as we are just testing and prefer not to test for next as that will always change

spec/models/schedule_spec.rb
```diff
@@ -11,4 +11,53 @@
     it { is_expected.to validate_presence_of(:applies_from) }
     it { is_expected.to validate_presence_of(:applies_to) }
   end
+
+  describe "instance methods" do
```
cwrw, line 15: can we avoid this extra nesting please, as adding `.name` defines a test for an instance method?

app/models/schedule.rb
```diff
@@ -9,4 +9,17 @@ class Schedule < ApplicationRecord
 
   validates :applies_from, presence: true
   validates :applies_to, presence: true
+
+  def self.default_for(course_group:, cohort: Cohort.current)
```
cwrw, line 13: @mooktakim any thoughts on this, yet to review your pr but just want to make sure this is aligned with you?

db/seeds/base/add_applications.rb
```diff
@@ -75,13 +75,80 @@
...
       :with_random_user,
       :with_random_work_setting,
-      :with_random_lead_provider_approval_status,
       :with_random_participant_outcome_state,
       lead_provider:,
       course: Course.all.sample,
       cohort: Cohort.all.sample,
     )
+
+    # users with one eligible for funded place application each (cohort funding_cap true)
```
cwrw, line 87: it feels like we're adding a bit too many applications here, and we're not looking at the withdrawn statuses yet in this ticket. Could we just maybe add 2 applications in a cohort with funded cap set with true/false and accepted so they are valid and maybe that's enough?

spec/factories/cohorts.rb
```diff
@@ -16,7 +16,11 @@
     end
 
     trait :with_funding_cap do
-      funding_cap { true }
+      after(:build) do |cohort|
+        next unless cohort.start_year >= 2024
```
cwrw, line 20: I think best avoid the 2024 value and just go with setting the funding cap to true false to determine if that cohort is funding cap ready as not keen on having hard values. Let me know your thoughts

spec/models/schedule_spec.rb
peteryates: Usually `#name` for instance and `.name` for class.

spec/factories/applications.rb
```diff
@@ -60,13 +60,16 @@
       eligible_for_funding { true }
     end
 
+    trait :with_random_funding_cap_cohort do
+      cohort { association :cohort, :with_funding_cap, start_year: Faker::Number.between(from: 2024, to: 2029) }
```
cwrw, line 64: same here around adding set numbers, just setting the funding cap value to true should be enough to indicate that in a cohort

spec/models/cohort_spec.rb
```diff
@@ -62,9 +62,5 @@
 
       expect(Cohort.current(Date.new(2022, 4, 10))).to eq(current_cohort)
     end
-
-    it "raises an error when there is no cohort for the current year" do
```
cwrw, line 66: why did this need to be removed?

spec/rails_helper.rb
```diff
@@ -118,6 +118,25 @@
     Course::IDENTIFIERS.each do |identifier|
       FactoryBot.create(identifier)
     end
+    (2021..2029).each do |start_year|
```
cwrw, line 121: same question here @mooktakim around what you are doing in your pr, thanks :)

cwrw (commented):
having a think I think we should probably defer any schedule logic to after this is merged https://github.com/DFE-Digital/npq-registration/pull/1418/files as this is the pr introducing schedules to acceptance. Happy to chat through that cc @mooktakim

db/seeds/base/add_applications.rb
```diff
@@ -73,15 +73,53 @@
...
+    FactoryBot.create_list(
+      :application,
+      2,
+      :eligible_for_funded_place,
+      :with_random_user,
+      :with_random_work_setting,
+      :with_participant_id_change,
+      lead_provider:,
+      course: Course.all.sample,
+      cohort: Cohort.where.not("start_year = ?", 2024).sample,
```
cwrw, line 109: instead of using 2024, can we use `funding_cap: false` instead here to avoid hardcoding years?

db/seeds/base/add_applications.rb
```diff
@@ -73,15 +73,53 @@
...
+    # users with one eligible for funded place application each (cohort funding_cap true)
+    FactoryBot.create_list(
+      :application,
+      2,
+      :eligible_for_funded_place,
+      :with_random_user,
+      :with_random_work_setting,
+      :with_participant_id_change,
+      lead_provider:,
+      course: Course.all.sample,
```
cwrw, line 96: could we set the cohort here to funding cap true?

db/seeds/base/add_applications.rb
```diff
@@ -73,15 +73,53 @@
...
+      funded_place: false,
```
cwrw, line 122: can we ensure that the application is accepted to be able to set this value? also that the cohort has funding cap: true? I think something to revisit across all the setup is:
1) If an application is accepted/withdraw/deferred and it is in a funding cap true cohort, we set the funded place true/false dependent on the eligible for funding field
2) If an application is accepted/withdrawn/deferred and is not in a funded cap true cohort, the funded place value remains false

Happy to jump on a call if there are questions

spec/services/participants/resume_spec.rb
```diff
@@ -35,7 +35,7 @@
     end
 
     context "when a matching application does not exist (different lead provider)" do
-      let(:lead_provider) { create(:lead_provider) }
+      let(:lead_provider) { LeadProvider.all.excluding(application.lead_provider).sample }
```
cwrw, line 38: I think @ethax-ross has fixed this in his pr :P https://github.com/DFE-Digital/npq-registration/pull/1430/files#diff-27987830d045f99f9132c178a2e7640f50db10a8e4f50e580e8b53fd8fc00c0bR28

spec/factories/cohorts.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :cohort do
-    start_year { Faker::Number.between(from: 2021, to: 2029) }
+    start_year { Faker::Number.between(from: 2021, to: Date.current.month < 9 ? Date.current.year : Date.current.year.succ) }
```
cwrw, line 3: can we keep this and avoid the ranges as they trip us up in ECF?

spec/rails_helper.rb
```diff
@@ -118,6 +118,27 @@
     Course::IDENTIFIERS.each do |identifier|
       FactoryBot.create(identifier)
     end
+    # create cohorts since 2021 with default schedule
```
cwrw, line 121: I think I would rather avoid this for now since some tests are creating specific cohorts and schedules, we can revisit this in a later ticket if that's ok?

db/seeds/base/add_applications.rb
```diff
@@ -73,15 +73,66 @@
...
-      :with_random_lead_provider_approval_status,
       :with_random_participant_outcome_state,
       lead_provider:,
       course: Course.all.sample,
       cohort: Cohort.all.sample,
     )
+
+    # users with one eligible for funded place application each (cohort funding_cap true)
+    FactoryBot.create_list(
+      :application,
```
cwrw, line 89: can this application be accepted please?

cwrw (approved):

cwrw (approved):
🙌

PR #1425, opened 2024-06-11
Add funded place logic to previously funded and eligibility

padv2010:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3150

### Description

Add support for `eligible_for_dfe_funding?` in applications, so we take into consideration
the `funded_place` attribute for capping.

Related to this PR (https://github.com/DFE-Digital/early-careers-framework/pull/4871).

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -65,6 +65,19 @@ class Application < ApplicationRecord
     withdrawn: "withdrawn",
   }
 
+  # `eligible_for_dfe_funding?`  takes into consideration what we know
+  # about user eligibility plus if it has been previously funded. We need
+  # to keep this method in place to keep consistency during the split between
+  # ECF and NPQ. In the mid term we will perform this calculation on NPQ and
+  # store the value in the `eligible_for_funding` attribute.
+  def eligible_for_dfe_funding?
+    if previously_funded?
+      false
+    else
+      eligible_for_funding
+    end
+  end
+
   def previously_funded?
     # This is an optimization used by the API Applications::Query in order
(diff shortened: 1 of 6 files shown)


app/models/application.rb
```diff
@@ -64,6 +64,17 @@ class Application < ApplicationRecord
     withdrawn: "withdrawn",
   }
 
+  # eligible_for_funding is solely based on what NPQ app knows
+  # eg school, course etc
+  # here we need to account for previous enrollments too
+  def eligible_for_dfe_funding?
```
padv2010, line 70: @cwrw following the same approach as in https://github.com/DFE-Digital/early-careers-framework/pull/4871, we would need to consider two scenarios: with funded place and without funded place. 
But this method is not used yet, so it is unclear to me how we should tackle this.

It is also unclear to me the difference between `eligible_for_dfe_funding` and `eligible_for_funding`. 

- `eligible_for_funding` is calculated in NPQ app based on the applicant responses
- `eligible_for_dfe_funding?` takes into consideration the LP decision about the applicant 

New developers coming to the code might find this confusing, I would aim for a different method name?

What about
- `funded_by_dfe?`
- `requested_for_dfe_funding?`

cwrw (commented):
@padv2010 has this been updated since our chat? Let me know if you have any further questions

app/models/application.rb
padv2010: Discussed in person, we will handle it after separation by including this logic into the eligibility calculation.

cwrw (approved):

PR #1428, opened 2024-06-13
Add `Participant::Resume` service and endpoints

ethax-ross:
[Jira-3088](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3088)

### Context

We want to add the participant resume endpoint from ECF, bringing across anything relevant to NPQ.

- Add Participant::Resume service

Add the service to move a participant back to the `active` state. This is lifted from ECF; instead of looking for the matching profile we find the matching application.

- Add /participants/npq/:id/resume endpoint

Adds the endpoint to enable providers to resume the training status of a participant.

### DIFF
--- a/app/controllers/api/v1/participants_controller.rb
+++ b/app/controllers/api/v1/participants_controller.rb
@@ -12,10 +12,19 @@ module API
         render json: to_json(participant)
       end
 
+      def resume
+        service = ::Participants::Resume.new(participant_action_params)
+
+        if service.resume
+          render json: to_json(service.participant)
+        else
+          render json: API::Errors::Response.from(service), status: :unprocessable_entity
+        end
+      end
+
       def change_schedule = head(:method_not_allowed)
       def defer = head(:method_not_allowed)
       def withdraw = head(:method_not_allowed)
-      def resume = head(:method_not_allowed)
       def outcomes = head(:method_not_allowed)
 
     private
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/participants_controller.rb
```diff
@@ -27,7 +36,18 @@ def participants_query
       end
 
       def participant_params
-        params.permit(:ecf_id, filter: %i[updated_since])
```
cwrw, line 30: could we keep the updated since here?

ethax-ross: It doesn't appear to be used? we pull in a concern which exposes the `updated_since`:

https://github.com/DFE-Digital/npq-registration/blob/main/app/controllers/api/v1/participants_controller.rb#L5
https://github.com/DFE-Digital/npq-registration/blob/main/app/controllers/api/concerns/filter_by_updated_since.rb#L8

spec/support/shared_examples/api_defer_participant_endpoint_support.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "an API defer participant endpoint" do
```
cwrw, line 3: ✂️ 😅

cwrw (approved):

spec/support/shared_examples/api_defer_participant_endpoint_support.rb
ethax-ross: Whoops, juggling too many PRs 😬

PR #1430, opened 2024-06-13
Add `Participant::Withdraw` service and endpoints

ethax-ross:
[Jira-3087](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3087)

### Context

We want to add the participant withdraw endpoint from ECF, bringing across anything relevant to NPQ.

- Add Participant::Withdraw service

Add the service to move a participant to a withdrawn state. This is lifted from ECF; instead of looking for the matching profile we find the matching application.

- Add /participants/npq/:id/withdraw endpoint

Adds the endpoint to enable providers to withdraw the training status of a participant.

- DRY up participant action services

Add `Participants::Action` abstract class that individual actions can inherit from. Move specs into a shared context.

- DRY up participant action request specs

Move specs into a single shared context and pass in the service.

- Rescue malformed request body and return 400

When the request contains a malformed body (without data/attributes) we want to return a 400 with an error message.

- Fix flakey participant action spec

This was flakey because the lead provider was occasionally being instantiated and then selected in the application anyway (via `sample`). To avoid this we ensure the application lead provider is created first (by referencing it in the name of the other lead provider).

### Guidance to review

Best reviewed by-commit.


config/locales/en.yml
```diff
@@ -28,11 +28,21 @@ en:
 
   participant: &participant
     blank: Your update cannot be made as the '#/participant_id' is not recognised. Check participant details and try again.
-
+    already_active: The participant is already active
+    already_deferred: The participant is already deferred
+    already_withdrawn: The participant is already withdrawn
+    no_declarations: You cannot defer an NPQ participant that has no declarations
+    already_withdrawn: The participant is already withdrawn
```
cwrw, line 35: ✂️

cwrw (approved):
🙌 very cleanly and nicely done!

PR #1432, opened 2024-06-13
[CPDLP-3085] TS6 - Add declarations table

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3085

* `Declaration` model added
* `cohort` and `lead_provider` has been included in declaration as this can change on the `application`
* `state_reason` is now on `Declaration` as its used when state is marked `ineligible`, therefore we do not need the `declaration_states` table. (reason is used for serializer)

### DIFF
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -1,0 +1,30 @@
+class Declaration < ApplicationRecord
+  belongs_to :application
+  belongs_to :cohort
+  belongs_to :lead_provider
+  belongs_to :superseded_by, class_name: "Declaration", optional: true
+
+  enum state: {
+    submitted: "submitted",
+    eligible: "eligible",
+    payable: "payable",
+    paid: "paid",
+    voided: "voided",
+    ineligible: "ineligible",
+    awaiting_clawback: "awaiting_clawback",
+    clawed_back: "clawed_back",
+  }, _suffix: true
+
+  enum declaration_type: {
+    started: "started",
+    "retained-1": "retained-1",
+    "retained-2": "retained-2",
+    completed: "completed",
+  }, _suffix: true
+
+  enum state_reason: {
+    duplicate: "duplicate",
+  }, _suffix: true
+
+  validates :declaration_date, :declaration_type, presence: true
+end
--- a/app/models/statement_item.rb
+++ b/app/models/statement_item.rb
@@ -3,7 +3,7 @@ class StatementItem < ApplicationRecord
   REFUNDABLE_STATES = %w[awaiting_clawback clawed_back].freeze
 
   belongs_to :statement
-  # belongs_to :declaration
+  belongs_to :declaration
 
   scope :billable, -> { where(state: BILLABLE_STATES) }
   scope :refundable, -> { where(state: REFUNDABLE_STATES) }
--- a/db/migrate/20240613151304_create_declarations.rb
+++ b/db/migrate/20240613151304_create_declarations.rb
@@ -1,0 +1,24 @@
+class CreateDeclarations < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :declaration_states, %w[submitted eligible payable paid voided ineligible awaiting_clawback clawed_back]
+    create_enum :declaration_types, %w[started retained-1 retained-2 completed]
+    create_enum :declaration_state_reasons, %w[duplicate]
+
+    create_table :declarations do |t|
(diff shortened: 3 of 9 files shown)


app/models/declaration.rb
```diff
@@ -0,0 +1,30 @@
+class Declaration < ApplicationRecord
+  belongs_to :application
+  belongs_to :cohort
+  belongs_to :lead_provider
+  belongs_to :superseded_by, class_name: "Declaration", optional: true
```
cwrw, line 5: can we also add the declaration to statement line items?

spec/factories/declarations.rb
```diff
@@ -0,0 +1,12 @@
+FactoryBot.define do
+  factory :declaration do
+    application
+    lead_provider
```
cwrw, line 4: could this also be application lead provider?

cwrw (commented):
minor comments but looking good

app/models/declaration.rb
ethax-ross: do we need `participant_declaration_attempts` as well or are we dropping that intentionally?

db/migrate/20240613151304_create_declarations.rb
```diff
@@ -0,0 +1,22 @@
+class CreateDeclarations < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :declaration_states, %w[submitted eligible payable paid voided ineligible awaiting_clawback clawed_back]
+    create_enum :declaration_types, %w[started retained-1 retained-2 completed]
```
ethax-ross, line 4: are extended declarations only applicable to ECF?

db/migrate/20240613151304_create_declarations.rb
```diff
@@ -0,0 +1,22 @@
...
+    create_enum :declaration_state_reasons, %w[duplicate]
+
+    create_table :declarations do |t|
```
ethax-ross, line 7: I think we will need `ecf_id` here too?

ethax-ross (commented):
Looks good, just a few questions

app/models/declaration.rb
cwrw: We are dropping them intentionally, as we don't need them yet

db/migrate/20240613151304_create_declarations.rb
cwrw: yes

db/migrate/20240613151304_create_declarations.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.uuid :ecf_id, default: "gen_random_uuid()", null: false, index: true
```
ethax-ross, line 8: @cwrw one for a future PR, but I noticed we default some `ecf_ids` in the migrations at the moment and not others; we should probably make them consistent. I think defaulting is probably going to be the way to go since we will be uncoupled from ECF in the future and need these IDs for a time 🤔

app/models/statement_item.rb
```diff
@@ -3,7 +3,7 @@ class StatementItem < ApplicationRecord
   REFUNDABLE_STATES = %w[awaiting_clawback clawed_back].freeze
 
   belongs_to :statement
-  # belongs_to :declaration
+  belongs_to :declaration
```
ethax-ross, line 6: I think we'll want the inverse relationships on declaration as well as we often look through the declaration to the statement/line items

db/migrate/20240613151304_create_declarations.rb
mooktakim: @ethax-ross 
The `ecf_id`'s we display in the serializers we'll need to default as the new records on NPQ only will still need to display this.  The others maybe we can remove the column after migration deployment.

app/models/statement_item.rb
mooktakim: I was thinking the `has_many` etc can be added as needed

db/migrate/20240613151304_create_declarations.rb
ethax-ross: Yeah that makes sense 👍 so probably update applications and users to default and remove from courses/lead providers after migrating

app/models/statement_item.rb
ethax-ross: Fair enough 👌

ethax-ross (approved):
LGTM

PR #1435, opened 2024-06-17
Prevent funded place change with applicable declarations

padv2010:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3086

### Description

Add logic to check for `applicable_declarations` declarations before removing the `funded_place` in an application.
The API endpoint, as the validation returns `false`, will trigger a 422 with the error message from the validator.

### DIFF
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -16,6 +16,8 @@ class Declaration < ApplicationRecord
   delegate :identifier, to: :course, prefix: true
   delegate :name, to: :lead_provider, prefix: true
 
+  scope :billable, -> { where(state: BILLABLE_STATES) }
+
   enum state: {
     submitted: "submitted",
     eligible: "eligible",
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -1,6 +1,6 @@
 class Declaration < ApplicationRecord
-  BILLABLE_STATES = %w[submitted eligible payable paid].freeze
-  NON_BILLABLE_STATES = %w[voided ineligible awaiting_clawback clawed_back].freeze
+  BILLABLE_STATES = %w[eligible payable paid].freeze
+  CHANGEABLE_STATES = %w[eligible submitted].freeze
 
   belongs_to :application
   belongs_to :cohort
@@ -17,6 +17,8 @@ class Declaration < ApplicationRecord
   delegate :name, to: :lead_provider, prefix: true
 
   scope :billable, -> { where(state: BILLABLE_STATES) }
+  scope :changeable, -> { where(state: CHANGEABLE_STATES) }
+  scope :billable_or_changeable, -> { billable.or(changeable) }
 
   enum state: {
     submitted: "submitted",
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -1,4 +1,7 @@
 class Declaration < ApplicationRecord
+  BILLABLE_STATES = %w[submitted eligible payable paid].freeze
+  NON_BILLABLE_STATES = %w[voided ineligible awaiting_clawback clawed_back].freeze
+
   belongs_to :application
   belongs_to :cohort
   belongs_to :lead_provider
--- a/spec/models/declaration_spec.rb
+++ b/spec/models/declaration_spec.rb
@@ -150,17 +150,30 @@ RSpec.describe Declaration, type: :model do
   end
 
   describe "scopes" do
+    let(:declarations) { Declaration.states.keys.map { |state| create(:declaration, state:) } }
+
     describe ".billable" do
       it "returns declarations with billable states" do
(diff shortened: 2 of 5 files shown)


db/schema.rb
```diff
@@ -97,9 +97,9 @@
     t.boolean "tsf_primary_plus_eligibility", default: false
     t.enum "lead_provider_approval_status", enum_type: "lead_provider_approval_statuses"
     t.text "participant_outcome_state"
-    t.bigint "school_id"
```
ethax-ross, line 100: It doesn't look like these schemas should have changed?

spec/services/applications/change_funded_place_spec.rb
```diff
@@ -92,6 +92,32 @@
             expect(service.errors.messages_for(:application)).to include("You must accept the application before attempting to change the '#/funded_place' setting.")
           end
         end
+
+        context "with declarations" do
+          before do
+            params.merge!(funded_place: false)
+          end
+
+          %w[submitted eligible payable paid].each do |applicable_state|
```
ethax-ross, line 101: I think we should put constants in the `Declaration` model for billable and unbillable states; we reference these all the time and its a mess in ECF, there's no structure around the declaration state groups

db/schema.rb
padv2010: Yes, thanks. I had some changes from another branch.

spec/services/applications/change_funded_place_spec.rb
padv2010: Thank you @ethax-ross. [Great call](https://github.com/DFE-Digital/npq-registration/pull/1435/commits/9255d42e3b2397a1250eb2cfd4c5b62878a36d48) to use `billable` states if this will be of extended use also in `NPQ`. Following the spirit of your comment [I have also extracted the scope](https://github.com/DFE-Digital/npq-registration/pull/1435/commits/843423eeaac28fdb9ef874de4fac68540371e1ba): `billable` which enables us to:

1. Align the name `billable` with the service
2. Increase the probability of not using the constants directly from our codebase (other than tests)

Any thoughts?

db/schema.rb
padv2010: Addressed in https://github.com/DFE-Digital/npq-registration/pull/1435/commits/6c5d7bc408adb05bb25d669f610804bd78773b12

spec/services/applications/change_funded_place_spec.rb
ethax-ross: Yep that sounds good thanks 👍

spec/models/declaration_spec.rb
```diff
@@ -14,4 +14,21 @@
...
+        billable_declarations = Declaration::BILLABLE_STATES.map do |state|
+          create(:declaration, state:)
+        end
+
+        non_billable_declarations = Declaration::NON_BILLABLE_STATES.map do |state|
+          create(:declaration, state:)
+        end
+
+        expect(Declaration.billable).to match_array(billable_declarations)
+        expect(Declaration.billable).not_to include(*non_billable_declarations)
```
ethax-ross, line 30: We shouldn't need this as the `match_array` will ensure it only contains billable declarations.

ethax-ross (approved):
Looks good to me!

spec/models/declaration_spec.rb
padv2010: Addressed in https://github.com/DFE-Digital/npq-registration/pull/1435/commits/784877789af410b288ac63f2563e59419f6909bf

app/models/declaration.rb
```diff
@@ -1,9 +1,14 @@
 class Declaration < ApplicationRecord
+  BILLABLE_STATES = %w[submitted eligible payable paid].freeze
```
cwrw, line 2: hmm not sure about including submitted in billable, as it's not technically a billable state 🤔

cwrw: in ECF there's changeable, maybe we include those and then we can merge scopes? let me know your thoughts

padv2010: Best option is to revert both commits: constant & scopes?
cc/ @cwrw @ethax-ross 

We could extract them based on actual code duplications?

padv2010: Discussed in person (@cwrw @ethax-ross), we have decided to rely on:

```ruby
scope :billable, -> { where(state: %w[eligible payable paid]) }
scope :changeable, -> { where(state: %w[eligible submitted]) }
scope :billable_or_changeable, -> { billable.or(changeable) }
```

 See https://github.com/DFE-Digital/npq-registration/pull/1435/commits/1b12f06c9b373614830cc3b9088e4939708437cb

ethax-ross (approved):
Looks great, just a typo but other than that LGTM

cwrw (approved):
👍  past the comment

PR #1437, opened 2024-06-17
Add Swagger docs for participant actions

ethax-ross:
[Jira-3091](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3091)

### Context

We've recently added participant action endpoints (resume, defer and withdraw) and we want to introduce the corresponding Swagger documentation.

- Add swagger docs for participant action endpoints

Add the swagger documentation for the participant action endpoints (resume/defer/withdraw).

- DRY up application documentation specs

We were creating specific support files for each action; we can instead have a single generic spec file that can be customised to cover any POST request against a resource. The PUT request for change-funded-place can use the
existing generic update shared context.

Minor updates to descriptions/summaries for participants to bring inline with applications.

- Fix response examples for participant actions

In the participant action response examples we want to show the correct `training_status`. In the response schema we have `training_status` as `active`, which was being used for resume/defer/withdraw.

To get a more accurate response example we have to define the whole response manually for each action.

Update shared context to allow customising the response example on a case by case basis (defaulting to the response
schema). Override the response example for participant actions.

- Update response examples for accept/reject application

If we use the default response example here the `status` will always be `accepted`. Instead, we have to manually override
the response example and adjust it for each action.


ethax-ross:
> looking good, for v3 mostly we have to set the withdrawal/deferral nested objects when those endpoints are called, could we populate those in the examples for only those endpoints, we've done something similar in ECF, and I'm keen to move over depending on complexity. Also in v1/v2 if we can set the `training_status` to the deferred/withdrawn values in those endpoint would be great as well :)

I've looked into a few ways of doing this; the most automated/maintainable way going forward IMO is to generate the response/request examples from the spec payloads directly. This removes the need to specify any examples in the schema files and ensures the examples we show are relevant for the endpoint (whereas before a generic response/request example was derived from the schema that was the same regardless of the endpoint).

The potential drawback to this approach is that the examples become coupled to our factories and they will change every time we regenerate the swagger docs. I've excluded the examples from the diff so that the CI check still works/doesn't care about this happening. We could look to make the factories more static/reliable in the data they generate if this is going to be a problem. 

I also looked at adding a generic hook into the shared contexts so we could take the standard response example from the schema and override it on a case-by-case basis, which might be more desirable but its a pain to get working as you have to manually parse the swagger schema into JSON (`rswag` doesn't seem to expose an interface to do this easily).

Let me know your thoughts, [here's the change](https://github.com/DFE-Digital/npq-registration/pull/1437/commits/1a0a58d47e772c3a40d08c8fc0b574e392690a96).

spec/swagger_schemas/requests/participant_withdraw_request.rb
```diff
@@ -0,0 +1,40 @@
...
+      required: %i[type attributes],
+      properties: {
+        type: {
+          description: "The data typed",
+          type: :string,
+          example: "participant-withdraw",
+        },
+        attributes: {
+          description: "A participant withdraw request attributes",
+          type: :object,
```
leandroalemao, line 17: Think we need to add a `required: %i[reason course_identifier],` here so we have a red * next to the required fields in the schema div. Same for the others actions requests schemas.

image

ethax-ross: Good spot 👍 I'll update all the request schemas to be consistent; looks like you need `required` on both the parent attribute (specifying the children) and on the children (set to `true`) to get both the asterisk and `required: true` text.

spec/support/helpers/swagger_example_parser.rb
```diff
@@ -0,0 +1,53 @@
...
+# of our schemas.
+#
+# Example usage to override the `status` attribute example in the `ApplicationResponse` schema:
+#
+# let(:response_example) do
+#  extract_swagger_example(schema: "#/components/schemas/ApplicationResponse", version: :v1).tap do |example|
+#   example[:data][:attributes][:status] = "accepted"
+#  end
+# end
+module Helpers
```
cwrw, line 18: 🌟

spec/requests/api/docs/v3/participants_spec.rb
```diff
@@ -24,4 +35,70 @@
...
+                    "#/components/schemas/ParticipantResponse",
+                    "#/components/schemas/ParticipantDeferRequest" do
+      let(:resource) { participant }
+      let(:type) { "participant-defer" }
+      let(:attributes) { { course_identifier: course.identifier, reason: Participants::Defer::DEFERRAL_REASONS.sample } }
+      let(:invalid_attributes) { { course_identifier: "invalid" } }
+      let(:response_example) do
+        base_response_example.tap do |example|
+          example[:data][:attributes][:npq_enrolments][0][:training_status] = "deferred"
+          example[:data][:attributes][:npq_enrolments][0][:withdrawl] = nil
```
cwrw, line 80:
```suggestion
          example[:data][:attributes][:npq_enrolments][0][:withdrawal] = nil
```

cwrw (approved):
Thank you @ethax-ross, cleanly done 🙌 just the two typos after that looks great, worth presenting to the team maybe today after the other session :)

leandroalemao (approved):
@ethax-ross Nice work tks 👍

PR #1442, opened 2024-06-18
[CPDLP-3173] Add change schedule endpoint

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3173

We would like to add the NPQ participant profile change schedule endpoint to NPQ separation

Add change schedule endpoint.

### DIFF
--- a/spec/support/helpers/swagger_example_parser.rb
+++ b/spec/support/helpers/swagger_example_parser.rb
@@ -42,7 +42,11 @@ module Helpers
           extract_swagger_properties(properties: value[:properties], version:)
         # If the value is an array of items, we need to extract the example from the items properties.
         elsif value[:type] == :array || value.key?(:items)
-          [extract_swagger_properties(properties: value[:items][:properties], version:)]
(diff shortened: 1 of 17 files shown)


spec/support/helpers/swagger_example_parser.rb
```diff
@@ -42,7 +42,11 @@ def extract_swagger_properties(properties:, version:)
           extract_swagger_properties(properties: value[:properties], version:)
         # If the value is an array of items, we need to extract the example from the items properties.
         elsif value[:type] == :array || value.key?(:items)
-          [extract_swagger_properties(properties: value[:items][:properties], version:)]
+          if value[:items].key?(:enum)
```
leandroalemao, line 45: had to make this changes as [npq courses](https://github.com/DFE-Digital/npq-registration/blob/main/spec/swagger_schemas/models/participant.rb#L39) has `type: :array`with no `properties` in `items:` as it’s an array of strings (not objects).

mooktakim (approved):
👍

PR #1445, opened 2024-06-19
Add declarations endpoint

ethax-ross:
[Jira-3092](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3092)

### Context

We want to port the participant declarations endpoint from ECF to NPQ reg.

- Add DeclarationSerializer for all API versions

Adds the `DeclarationSerializer` to replicate the ECF serialization for each API version. Adds convenience methods to `Declaration` to keep the serializer as simple as possible.

- Add Declarations::Query service

Add the `Declarations::Query` service to enable querying declarations. The functionality is lifted directly from ECF and supports all API versions.

- Add /api/:version/declarations endpoint

Add the GET `/api/:version/declarations` endpoint, which calls the `Declarations::Query` and returns via the `Declarations::Serializer`. The functionality for filtering has been lifted from ECF.

- Update /declarations to /participant-declarations

We want to keep parity with the ECF routes.

### Guidance to review

~~The API is slightly different `/declarations` instead of `/participant-declarations`; this was already setup like this in the routes so I assume the difference is intentional?~~

We don't appear to support custom sorting via the API and default to `created_at`.

The `has_passed` can be implemented in the serializer after we have added outcomes.

The only difference in filtering is v3 can filter by cohorts.

The differences in serialization are v1 has:

- `voided`
- `eligible_for_payment` 

And v3 has:

- `created_at`
- `statement_id`
- `clawback_statement_id`
- `uplift_paid`
- `lead_provider_name`
- `ineligible_for_funding_reason`


app/serializers/api/declaration_serializer.rb
```diff
@@ -0,0 +1,44 @@
...
+      field(:declaration_type)
+      field(:course_identifier)
+      field(:declaration_date)
+      field(:updated_at)
+      field(:state) { |declaration| declaration.state.dasherize }
+      # TODO: implement once we have outcomes
+      field(:has_passed) { nil }
+
+      view :v1 do
+        field(:voided_state?, name: :voided)
```
leandroalemao, line 19: Should we remove these 2 as they're deprecated in favour of `state`?

https://manage-training-for-early-career-teachers.education.gov.uk/api-reference/reference-v1.html#schema-participantdeclarationattributes

app/serializers/api/declaration_serializer.rb
```diff
@@ -0,0 +1,44 @@
+module API
```
leandroalemao, line 1: Noticed some differences in the order of some of the fields if we compare with the same endpoint in ECF app (I know we removed some fields as those apply to ECT/Mentors only). Should we follow the same order? Just trying to start a chat on this.. tks

leandroalemao (commented):
Great work @ethax-ross 
Left some minor comments
tks

app/controllers/api/v3/declarations_controller.rb
```diff
@@ -1,10 +1,35 @@
...
+
       def create = head(:method_not_allowed)
       def show = head(:method_not_allowed)
-      def index = head(:method_not_allowed)
       def void = head(:method_not_allowed)
+
+    private
+
+      def declarations_query
+        conditions = { lead_provider: current_lead_provider, updated_since:, participant_ids:, cohort_start_years: }
```
cwrw, line 18: I think we also need `delivery_partner_id` on v3

app/serializers/api/declaration_serializer.rb
cwrw: I think we should order reasonably in a way that makes the most sense, they don't have to match, with the timestamps hopefully at the end :)

app/controllers/api/v3/declarations_controller.rb
ethax-ross: is delivery partner not just ECF though?

app/serializers/api/declaration_serializer.rb
ethax-ross: One for @cwrw but I expect we'll need to keep it in v1 as it would be a breaking change to remove it

app/serializers/api/declaration_serializer.rb
ethax-ross: I've pushed the timestamp to the end, as much as it pains my OCD 😅 making the other fields match ECF would result in multiple view blocks which I think would be 🤮

app/controllers/api/v3/declarations_controller.rb
cwrw: 🤦 yep!

app/serializers/api/declaration_serializer.rb
cwrw: yea I think let's keep it as otherwise will be a schema change

app/services/declarations/query.rb
```diff
@@ -0,0 +1,62 @@
...
+    def where_updated_since(updated_since)
+      return if updated_since == :ignore
+
+      scope.merge!(Declaration.where(updated_at: updated_since..))
+    end
+
+    def where_participant_ids_in(participant_ids)
+      return if participant_ids == :ignore
+
+      scope.merge!(Declaration.where(user: { ecf_id: extract_conditions(participant_ids) }))
```
cwrw, line 38: existing issue in ECF: if you pass in a participant id that isn't in UUID format it returns a 500, one to tackle I guess in a different ticket

app/services/declarations/query.rb
```diff
@@ -0,0 +1,62 @@
...
+    def where_cohort_start_year_in(cohort_start_years)
+      return if cohort_start_years == :ignore
+
+      scope.merge!(Declaration.where(cohort: { start_year: extract_conditions(cohort_start_years) }))
+    end
+
+    def all_declarations
+      Declaration
+        .includes(
+          :cohort,
```
cwrw, line 50: can we include the lead provider here, as the lead provider is stamped on the declaration

app/serializers/api/declaration_serializer.rb
```diff
@@ -0,0 +1,45 @@
...
+      view :v1 do
+        field(:voided_state?, name: :voided)
+        field(:eligible_for_payment?, name: :eligible_for_payment)
+      end
+
+      view :v2 do
+      end
+
+      view :v3 do
+        field(:created_at)
```
cwrw, line 26: any chance we can move this timestamp down till the end of the v3 block 😄

cwrw (approved):
perfect, just the minor comments 🙌

app/services/declarations/query.rb
ethax-ross: Ah good catch; yeah I think pick that up as a separate ticket if its an issue throughout the app

cwrw (approved):

PR #1450, opened 2024-06-20
Add declarations CSV endpoint

ethax-ross:
[Jira-3816](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3186)

### Context

We want to bring the declarations CSV endpoint from ECF across to NPQ reg for v1 and v2 of the API (it doesn't exist in v3).

- Add DeclarationsCsvSerializer

Add `DeclarationsCsvSerializer`; as per ECF this will return the same data as the JSON serializer/API endpoint.

Unlike the other serializers in ECF this one will return an empty CSV if there are no declarations (other serializers return the header rows).

- Add api/:version/participant-declarations.csv endpoint

Add the CSV endpoint for participant declarations in v1 and v2 of the API, as per ECF.

- Add swagger docs for declarations CSV endpoints

Add swagger documentation for the declarations CSV endpoints.

- DRY up the swagger enums/examples

We can reference the enums directly here; this will keep the documentation inline with the code automatically as well.

- DRY up and fix swagger versioning

Previously we were duplicating some swagger schemas instead of building off previous versions. When I cleaned this up I also noticed that the hashes were passing by reference between the versions, so we need to `deep_dup` them prior to modifying, or all versions would end up with the changes.


spec/swagger_schemas/models/application.rb
```diff
@@ -224,249 +191,13 @@
...
-        description: "The data type",
-        type: :string,
-        example: "npq_application",
-        enum: %w[
-          npq_application
-        ],
-      },
-      attributes: {
-        properties: {
-          participant_id: {
```
cwrw, line 246: I think mainly this was done to maintain the order of the fields in the API to match what we surface for that version, losing this will mean that the order in the schema is affected 🤔 happy to chat about this

app/serializers/api/declarations_csv_serializer.rb
```diff
@@ -0,0 +1,52 @@
...
+        declarations_json.each do |declaration_json|
+          csv << to_row(declaration_json, headers:)
+        end
+      end
+    end
+
+  private
+
+    def declarations_json
+      @declarations_json ||= JSON.parse(DeclarationSerializer.render(declarations, view: :v1)).map(&method(:flatten_hash))
```
cwrw, line 31: could we pass in the version here as v1 has a few extra fields to v2?

ethax-ross: whoops, yes, that was the intention its in this class as `view` but I haven't forwarded it to the serializer; good catch I'll add a test as well

ethax-ross: [Fixed](https://github.com/DFE-Digital/npq-registration/pull/1450/commits/5c56525edd42d46467903bee90b8dbe0d895ef25)

cwrw (approved):
🙌

leandroalemao (approved):

PR #1460, opened 2024-06-23
Add `/participant-declarations/:id/void` endpoints

ethax-ross:
[Jira-3095](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3095)

### Context

We want to add the `/participant-declarations/:id/void` endpoints from ECF so that lead providers are able to void and clawback declarations.

- Add convenience methods to Declaration/Statement/LeadProvider

These are the convenience methods we need for finding output fee statements and checking the state of declarations as part of the void/claw back service.

In future we're going to look at introducing a state machine to the `Declaration` model, similar to that in `StatementItem` to manage the state transitions.

- Add Declarations::Void service

Ports across the void declaration logic from ECF. Confusingly the `/void` action of the API is used for both voiding declarations and clawing them back (the action is determined by the state of the declaration). In ECF we split out
the void and clawback services but I opted to combine them here (as the void service calls the clawback service anyway and the additional service just makes it more confusing than it already is, IMO).

The validation rule to ensure only a `paid` declaration can be clawed back turned out to be redundant (as the service treats non-paid declarations as a void intention anyway) so I didn't bring this across.

This also brings across the `StatementAttacher` from ECF, complete with the validation to make sure there is a next output fee statement.

- Add /participant-declarations/:id/void endpoint

Add the `/void` endpoint for participant declarations; this will either void or clawback the declaration as per the logic in ECF.

- Add request specs for voiding declarations

- DRY up request specs with generic shared contexts

We've been adding a shared context for each action, but this isn't really necessary as the actions for update/create on resource are largely identical and we just need to switch out the service being called and the serializer being used.


spec/models/statement_spec.rb
```diff
@@ -69,6 +69,23 @@
...
+    describe ".with_output_fee" do
+      it "selects only output fee statements" do
+        expect(Statement.with_output_fee.to_sql).to include(%(WHERE "statements"."output_fee" = TRUE))
+      end
+    end
+
+    describe ".next_output_fee_statements" do
+      it "selects the output fee statement with the earliest deadline date in the future" do
+        freeze_time do
+          sql = Statement.next_output_fee_statements.to_sql
```
cwrw, line 82: I wonder if we can test those with actual statements like the test in the lead provider spec, mostly because I feel like those can be tested easily without asserting against the SQL?

app/services/declarations/void.rb
```diff
@@ -0,0 +1,75 @@
...
+    def output_fee_statement_available
+      errors.add(:declaration, :no_output_fee_statement, cohort: declaration.cohort) unless statement_attacher.valid?
+    end
+
+    def statement_attacher
+      @statement_attacher ||= StatementAttacher.new(declaration:)
+    end
+
+    def clawing_back?
+      declaration&.paid_state?
```
cwrw, line 68: There is an issue in the old implementation that if someone hits the void endpoint for a paid declaration, which then moved to awaiting clawback, and hits it again we get the voidable error which is confusing. Maybe we can include paid/awaiting clawback/clawed back states here to ensure it goes down the right path and returns the right error, happy to discuss and explain further

app/services/declarations/void.rb
```diff
@@ -0,0 +1,75 @@
...
+  private
+
+    def clawback_declaration
+      declaration.awaiting_clawback_state!
+      statement_attacher.attach
+    end
+
+    def void_declaration
+      declaration.voided_state!
+      declaration.statement_items.with_state(:eligible, :payable).first&.mark_voided!
```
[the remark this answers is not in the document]
ethax-ross: I thought so as well initially but we don't appear to have the `submitted` state for line items:

https://github.com/DFE-Digital/early-careers-framework/blob/d0a2c8a06352a4ac8a7f5d4b8423a1e4c798f619/app/models/finance/statement_line_item.rb#L9-L16

app/services/declarations/void.rb
ethax-ross: I think I follow; I've used the error we had for this scenario in ECF (although the bug meant it was not reachable in ECF). [Here's the change](https://github.com/DFE-Digital/npq-registration/pull/1460/commits/673f811f05ccce26ba428b0cc9b9389382f415ec).

spec/models/statement_spec.rb
ethax-ross: [Updated](https://github.com/DFE-Digital/npq-registration/pull/1460/commits/355e9912dc381169d481721dab2d2afaad23b61a)

cwrw (approved):
🙌

leandroalemao (approved):

PR #1467, opened 2024-06-25
[CPDLP-3197] Option 1

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3197

* `:with_default_schedules` tag to create defaults schedules before each test


peteryates:
➕ for option 1, option 2 is just sneaky seeding.

PR #1453, opened 2024-06-20
[CPDLP-3089] TS6 - Add change schedule participant service

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3089

### Changes proposed in this pull request

* `Participants::ChangeSchedule` service created with tests

PR #1475, opened 2024-06-25
[CPDLP-3094] Add POST declaration endpoint

leandroalemao:
### Context

Ticket: [CPDLP-3099](https://dfedigital.atlassian.net/browse/CPDLP-3099)

We would like to add the POST participant declaration endpoint to NPQ separation.

- Add new API endpoint for all API versions: `POST api/v{1/2/3}/participant-declarations​`
- The endpoint accepts the following request body:
```
{
  "data": {
    "type": "participant-declaration",
    "attributes": {
      "participant_id": "db3a7848-7308-4879-942a-c4a70ced400a",
      "declaration_type": "started",
      "declaration_date": "2021-05-31T02:21:32.000Z",
      "course_identifier": "npq-leading-teaching"
      "has_passed": true # only on completed requests
    }
  }
}
```
- All validations and mapping are similar to ECF
- Creation of outcomes can be ignored for now, however we’ll just validate the value of the has_passed field being passed in
- Some fields are required on some requests but not others
- no documentation in this ticket
- no declaration attempt
- align with creating declaration ticket on statement line attacher logic

[CPDLP-3099]: https://dfedigital.atlassian.net/browse/CPDLP-3099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/support/shared_examples/api_create_declaration_endpoint_support.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "an API create declaration endpoint" do
```
ethax-ross, line 3: Can we make this a generic "an API create endpoint" that just checks it accepts a POST request of the correct body and calls the expected service/serializer? Similar to [this](https://github.com/DFE-Digital/npq-registration/blob/f6ab9971671edf56485560ffc181ca6a43981e92/spec/support/shared_examples/api_update_endpoint_support.rb)

app/services/declarations/create.rb
```diff
@@ -0,0 +1,164 @@
...
+
+    attribute :lead_provider
+    attribute :participant_id
+    attribute :declaration_type, :string
+    attribute :declaration_date, :datetime
+    attribute :course_identifier, :string
+    attribute :has_passed
+
+    validates :lead_provider, presence: true
+    validates :participant_id, presence: { message: I18n.t("declaration.missing_participant_id") }
```
ethax-ross, line 16: We could use implicit translations here to avoid specifying them all

app/services/declarations/create.rb
```diff
@@ -0,0 +1,164 @@
...
+  class Create
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :lead_provider
+    attribute :participant_id
+    attribute :declaration_type, :string
+    attribute :declaration_date, :datetime
+    attribute :course_identifier, :string
+    attribute :has_passed
```
ethax-ross, line 13: I expect the answer is yes, but just checking we're omitting `evidence_held` intentionally here?

app/services/declarations/create.rb
```diff
@@ -0,0 +1,164 @@
...
+
+    validates :lead_provider, presence: true
+    validates :participant_id, presence: { message: I18n.t("declaration.missing_participant_id") }
+    validates :participant, participant_presence: true, participant_not_withdrawn: true
+    validates :course_identifier, inclusion: { in: Course::IDENTIFIERS, message: I18n.t(:invalid_course) }, allow_blank: false, course_for_participant: true
+    validates :declaration_date, declaration_date: true, allow_blank: true
```
ethax-ross, line 19: Don't think we want to `allow_blank` for declaration date? even if its caught by the `declaration_date` validator its misleading having it here

spec/services/declarations/create_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+    }
+  end
+
+  before { create(:contract, course:, cohort:, lead_provider:) }
+
+  subject(:service) { described_class.new(**params) }
+
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:lead_provider) }
+    it { is_expected.to validate_presence_of(:participant_id).with_message("The property '#/participant_id' must be present") }
```
ethax-ross, line 35: Can we avoid hard coding the error messages

spec/services/declarations/create_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+          expect(service.errors.messages_for(:participant_id)).to eq(["This participant withdrew from this course on #{application.application_states.last.created_at.rfc3339}. Enter a '#/declaration_date' that's on or before the withdrawal date."])
+        end
+      end
+    end
+
+    context "when an existing declaration already exists" do
+      before { service.save }
+
+      context "when the state submitted" do
+        it "does create duplicates" do
```
ethax-ross, line 106: `does not create duplicates`

spec/services/declarations/create_spec.rb
```diff
@@ -0,0 +1,164 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Declarations::Create, type: :model do
```
ethax-ross, line 5: I think we're missing a few test scenarios here:

- Output fee statements
- Superseded by
- Eligible/ineligible states

app/services/declarations/create.rb
leandroalemao: Hi @ethax-ross think `evidence_held` is only applicable to declarations for ECF profiles isn't it?

app/models/application.rb
```diff
@@ -66,16 +66,22 @@ class Application < ApplicationRecord
     withdrawn: "withdrawn",
   }
 
+  def funding_eligibility(with_funded_place:)
```
cwrw, line 69: could this be private or do we use it anywhere?

app/models/application.rb
```diff
@@ -66,16 +66,22 @@ class Application < ApplicationRecord
...
+    eligible_for_funding && (funded_place.nil? || funded_place)
+  end
+
   # `eligible_for_dfe_funding?`  takes into consideration what we know
   # about user eligibility plus if it has been previously funded. We need
   # to keep this method in place to keep consistency during the split between
   # ECF and NPQ. In the mid term we will perform this calculation on NPQ and
   # store the value in the `eligible_for_funding` attribute.
-  def eligible_for_dfe_funding?
+  def eligible_for_dfe_funding?(with_funded_place: false)
```
cwrw, line 80: is there a way we can avoid bringing in the complexity from ECF here, do we need those methods to be shared or can they be different, something to think about 🤔

spec/models/application_spec.rb
```diff
@@ -317,4 +317,36 @@ def subject.transient_previously_funded
...
+    end
+
+    context "when it is eligible_for_funding but and has a funded place" do
+      subject { create(:application, eligible_for_funding: true, funded_place: true) }
+
+      it { is_expected.to be_fundable }
+    end
+
+    context "when it is not eligible_for_funding but and has a funded place" do
+      subject { create(:application, eligible_for_funding: false, funded_place: false) }
```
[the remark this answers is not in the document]
cwrw: I need to fix this in ECF as well

app/models/contract.rb
```diff
@@ -0,0 +1,4 @@
+class Contract < ApplicationRecord
```
cwrw, line 1: discussed in person: double checking contract as we opted to leave it out initially from MVP as it is not needed, will confirm shortly

spec/services/declarations/create_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+    context "when an existing declaration already exists" do
+      before { service.save }
+
+      context "when the state submitted" do
+        it "does not create duplicates" do
+          expect { service.save }.not_to change(Declaration, :count)
+        end
+      end
+
+      context "with an fundable participant" do
```
ethax-ross, line 110: I think the nesting has got a little off here; we could do it all in a loop:

```
        %i[eligible payable paid].each do |state|
          context "when the state uis #{state}" do
            before { existing_declaration.update!(state:) }

            it "does not create duplicates" do
              expect { service.save }.not_to change(Declaration, :count)

              expect(existing_declaration).to be_eligible_state
            end
          end
        end
```

spec/models/declaration_spec.rb
```diff
@@ -204,4 +230,46 @@
...
+  describe "#duplication_declarations" do
+    let(:cohort) { create(:cohort, :current) }
+    let(:course_group) { CourseGroup.find_by(name: "leadership") || create(:course_group, name: "leadership") }
+    let(:course) { create(:course, :sl, course_group:) }
+    let(:schedule) { create(:schedule, :npq_leadership_autumn, course_group:, cohort:) }
+    let(:application) { create(:application, :accepted, cohort:, course:) }
+    let(:participant) { application.user }
+    let!(:declaration) { create(:declaration, application:) }
+
+    context "when a user exists with the same TRN" do
```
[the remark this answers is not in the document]
ethax-ross: I think this is still the case? (sorry if you've not got to it yet)

spec/validators/participant_presence_validator_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    context "with no participant" do
+      let(:participant) { nil }
+
+      it "is invalid" do
+        expect(subject).to be_invalid
+      end
+
+      it "has a meaningful error", :aggregate_failures do
+        expect(subject).to be_invalid
+        expect(subject.errors.messages_for(:participant_id)).to include("Your update cannot be made as the '#/participant_id' is not recognised. Check participant details and try again.")
```
ethax-ross, line 45: Can we test on the translation key instead of the message here?

spec/services/declarations/create_spec.rb
ethax-ross: There are still a few hard-coded translations kicking about; you can test the ones with a variable as well using the key like this:

```
    record
      .errors
      .add(:participant_id, :declaration_must_be_before_withdrawal_date, withdrawal_date: latest_state.created_at.rfc3339)
```

```
expect(subject.errors.first).to have_attributes(attribute: :participant_id, type: :declaration_must_be_before_withdrawal_date, options: { withdrawal_date: application.application_states.last.created_at.rfc3339 })

```

ethax-ross:
Looking great 👍 just a few minor comments left

app/services/declarations/create.rb
```diff
@@ -0,0 +1,162 @@
...
+    validates :cohort, contract_for_cohort_and_course: true
+
+    validate :output_fee_statement_available
+    validate :validate_has_passed_field, if: :validate_has_passed?
+    validate :validate_schedule_exists
+    validate :validates_billable_slot_available
+
+    attr_reader :raw_declaration_date
+
+    def save
```
mooktakim, line 31: To maintain consistency with other service classes, can we rename `save` to `create`

app/models/contract.rb
mooktakim: I don't think we should create `Contract` in the PR as we have yet to decide how we'll migrate `NPQContract`.
I think its only needed for validation `contract_for_cohort_and_course: true`, if we ignore that for now, we can remove contracts.

app/models/declaration.rb
```diff
@@ -79,4 +81,25 @@ def ineligible_for_funding_reason
...
+        declaration_type:,
+        superseded_by_id: nil,
+        application: { course: application.course.rebranded_alternative_courses },
+      )
+  end
+
+private
+
+  def declaration_date_not_in_the_future
+    errors.add(:declaration_date, :future_declaration_date) if declaration_date && declaration_date > Time.zone.now
```
mooktakim, line 103: It'll be more readable if we do: `declaration_date && declaration_date.future?`

app/services/declarations/create.rb
mooktakim: rubcop 🤦

mooktakim (changes requested):
Some comments.

app/services/declarations/create.rb
```diff
@@ -0,0 +1,166 @@
...
+        &.includes(:course)
+        &.find_by(lead_provider:, course: { identifier: course_identifier })
+    end
+
+    def participant
+      @participant ||= begin
+        Participants::Query.new(lead_provider:).participant(ecf_id: participant_id)
+      rescue StandardError
+        nil
+      end
```
peteryates, line 63: This is really dangerous.

leandroalemao: Agree.. I've made the rescue exception specific.. tks 👍

spec/models/application_spec.rb
```diff
@@ -317,4 +317,36 @@ def subject.transient_previously_funded
...
+    end
+
+    context "when it is eligible_for_funding but and has a funded place" do
+      subject { create(:application, eligible_for_funding: true, funded_place: true) }
+
+      it { is_expected.to be_fundable }
+    end
+
+    context "when it is not eligible_for_funding but and has a funded place" do
+      subject { create(:application, eligible_for_funding: false, funded_place: true) }
```
cwrw, line 347: we can never have `funded_place` true and `eligible_for_funding` as false, if `eligible_for_funding` is false the `funded_place` will always be false. Could we just amend the context to signify that and change this back to false please?

app/models/declaration.rb
```diff
@@ -79,4 +81,25 @@ def ineligible_for_funding_reason
...
   end
+
+  def duplicate_declarations
+    self
+      .class
+      .joins(application: %i[user course])
+      .where(user: { trn: application.user.trn })
+      .where.not(user: { trn: nil })
+      .where.not(id:)
+      .where(state: %w[submitted eligible payable paid])
```
cwrw, line 92: could we use the scope: `billable_or_changeable `?

app/models/declaration.rb
```diff
@@ -79,4 +81,25 @@ def ineligible_for_funding_reason
...
     end
   end
+
+  def duplicate_declarations
+    self
+      .class
+      .joins(application: %i[user course])
+      .where(user: { trn: application.user.trn })
+      .where.not(user: { trn: nil })
+      .where.not(id:)
```
cwrw, line 91: could we also check that the user isn't the same as per ECF:
```
.where.not(user_id:, id:)
```

app/validators/declaration_date_validator.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def declaration_within_schedule(record)
+    return unless record.schedule && record.declaration_date.present?
+
+    if record.declaration_date < record.schedule.applies_from.beginning_of_day
+      record.errors.add(:declaration_date, :declaration_before_milestone_start)
+    end
+
+    if record.schedule.applies_to.end_of_day <= record.declaration_date
```
cwrw, line 33: I think `milestone_date` in ECF doesn't equate to `applies_to` here as in ECF all milestone dates are `nil` for NPQ, and we only have payment date which we don't validate against in declarations, do you mind confirming @leandroalemao unless I'm missing something

spec/validators/declaration_date_validator_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        subject { model_class.new(declaration_date: "2023-19-01T29:21:55Z") }
+
+        it "has a meaningful error", :aggregate_failures do
+          expect(subject).to be_invalid
+          expect(subject.errors.first).to have_attributes(attribute: :declaration_date, type: :invalid)
+        end
+      end
+    end
+
+    context "when declaration_date is before the milestone start" do
```
cwrw, line 81: could we avoid using milestone in the tests as we don't have that concept in NPQ here and across this test?

app/services/declarations/create.rb
```diff
@@ -0,0 +1,166 @@
...
+                        state: %w[submitted eligible payable paid],
+                      ).exists?
+
+      errors.add(:base, :declaration_already_exists)
+    end
+
+    def validate_has_passed_field
+      self.has_passed = has_passed.to_s
+
+      return unless has_passed.blank? || !%w[true false].include?(has_passed)
```
cwrw, line 151: in ECF we return different messages for blank vs wrong values, should we do the same here?

app/services/declarations/create.rb
```diff
@@ -0,0 +1,166 @@
...
+    def validates_billable_slot_available
+      return if errors.any?
+      return unless participant
+
+      return unless Declaration
+                      .joins(application: %i[user course])
+                      .where(
+                        application: { user: participant, courses: { identifier: course_identifier } },
+                        declaration_type:,
+                        state: %w[submitted eligible payable paid],
```
cwrw, line 142: same here around using a scope please

app/validators/course_for_participant_validator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    record.errors.add(:course_identifier, :invalid_course)
+  end
+
+private
+
+  def has_accepted_application_for_course_given_course_identifier?(record)
+    return if record.participant.blank?
+
+    record.participant.applications.joins(:course).accepted.active.map { |application|
```
cwrw, line 16: I'm unsure about this validation 🤔 do you mind pointing me to where this is in ECF, as I think we just need to validate that the course identifiers exist which I think you're already doing?

app/services/declarations/create.rb
leandroalemao: Hi @cwrw Yeah I noticed that.. but the message content is the same that's why I merged both here..

https://github.com/DFE-Digital/early-careers-framework/blob/main/config/locales/en.yml#L60
https://github.com/DFE-Digital/early-careers-framework/blob/main/config/locales/en.yml#L81

app/validators/course_for_participant_validator.rb
leandroalemao: Hi @cwrw it's [here](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/validators/course_validator.rb#L16) .. it's getting all active participant profiles and then checks the course identifiers.
The idea here is to get all accepted & active applications and do the same with the course identifier?
It's basically checking if the participant has any accepted & active application for the given course..

app/validators/declaration_date_validator.rb
leandroalemao: Ah yeah we're not migrating `milestone_date` over as it's `nil` for NPQ, we did a spike on that. The question is .. should we keep this & validate the the declaration against the payment_date (`applies_to`)? Or in another words.. is it ok to create a declaration after a schedule `applies_to` date?

mooktakim: We have something similar here:
https://github.com/DFE-Digital/npq-registration/pull/1453/files#diff-92e39530416181739ced666c8a1c0e1e14f64d062ea6cabd5819387b76ef6f38R82

ECF had the check with `milestone_date`, but I removed it because we don't have `milestone_date` on NPQ.

cwrw: discussed in person: for now we're going to remove the validation on applies to

app/validators/course_for_participant_validator.rb
cwrw: discussed in person: we can tidy the query here and removed active, but need the invalid course validation. Also think about multiple queries doing the same thing and how we can tidy up

cwrw: discussed in person: we're going to amend the query here

app/models/declaration.rb
mooktakim: I think this:
```
.where.not(user: { trn: nil, id: application.user_id })
```
would only check `user_id` if `trn` is nil and should be two separate where statements:
```
.where.not(user: { trn: nil })
.where.not(user: { id: application.user_id })
```
Or I'm messing up the logic in my head.

cwrw: discussed in person: let's separate just for readability and remove includes

config/locales/en.yml
```diff
@@ -28,6 +28,24 @@ en:
     cohort_does_not_accept_capping: "Leave the '#/funded_place' field blank. It's only needed for participants starting NPQs from autumn 2024 onwards."
     cannot_change_funded_place: "You must void or claw back your declarations for this participant before being able to set '#/funded_place' to false"
 
+  declaration: &declaration
+    future_declaration_date: "The '#/declaration_date' value cannot be a future date. Check the date and try again."
+    mismatch_declaration_type_for_schedule: "The property '#/declaration_type' does not exist for this schedule."
```
cwrw, line 33: I think this should be under declaration type, as testing locally I'm getting a mismatch error?

config/locales/en.yml
```diff
@@ -28,6 +28,24 @@ en:
...
+    declaration_date:
+      blank: "Enter a '#/declaration_date'."
+      invalid: "Enter a valid RCF3339 '#/declaration_date'."
+      declaration_before_schedule_start: "Enter a '#/declaration_date' that's on or after the schedule start."
+    declaration_type:
+      blank: "Enter a '#/declaration_type'."
+    has_passed:
+      invalid: "Enter 'true' or 'false' in the '#/has_passed' field to indicate whether this participant has passed or failed their course."
+    course_identifier:
+      invalid_course: "The entered '#/course_identifier' is not recognised for the given participant. Check details and try again."
```
cwrw, line 47: same for this one I think it needs to move under the declaration/create as coming back as missing

app/models/declaration.rb
```diff
@@ -79,4 +81,26 @@ def ineligible_for_funding_reason
...
+        declaration_type:,
+        superseded_by_id: nil,
+        application: { course: application.course.rebranded_alternative_courses },
+      )
+  end
+
+private
+
+  def declaration_date_not_in_the_future
+    errors.add(:declaration_date, :future_declaration_date) if declaration_date&.future?
```
cwrw, line 104: in ECF we catch those errors after the validation happens, in NPQ we don't have that so the endpoint will fall over after passing the service validation. Maybe we can move this into the service instead, similar to what has been done here? https://github.com/DFE-Digital/npq-registration/pull/1498/files#diff-f633d59f99e20b9c30469f9dff5e59e7612da3eca47b2055f7b9e1b05004635dR76

config/locales/en.yml
```diff
@@ -28,6 +28,24 @@ en:
     cohort_does_not_accept_capping: "Leave the '#/funded_place' field blank. It's only needed for participants starting NPQs from autumn 2024 onwards."
     cannot_change_funded_place: "You must void or claw back your declarations for this participant before being able to set '#/funded_place' to false"
 
+  declaration: &declaration
+    future_declaration_date: "The '#/declaration_date' value cannot be a future date. Check the date and try again."
```
cwrw, line 32: same issue with this translation as well being missing, but hopefully if we move it into the service it will be ok

cwrw (approved):
Thank you for addressing all the comments @leandroalemao 🙌

mooktakim (approved):

PR #1479, opened 2024-06-26
[CPDLP-3197] TS7 - Generating spec and seed data

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3197

* Use `:with_default_schedules` tag for only the specific tests that require default schedules.

### DIFF
--- a/spec/services/applications/accept_spec.rb
+++ b/spec/services/applications/accept_spec.rb
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-RSpec.describe Applications::Accept do
+RSpec.describe Applications::Accept, :with_default_schedules do
   let(:params) do
     {
       application:,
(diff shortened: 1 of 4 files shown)


leandroalemao (commented):
Should we then remove the schedule creation from these lines or should we keep it explicit?

https://github.com/DFE-Digital/npq-registration/blob/a95876e3e5e759830e4df37968182969fd4f7183/spec/services/applications/accept_spec.rb#L21
https://github.com/DFE-Digital/npq-registration/blob/a95876e3e5e759830e4df37968182969fd4f7183/spec/services/applications/accept_spec.rb#L37
https://github.com/DFE-Digital/npq-registration/blob/a95876e3e5e759830e4df37968182969fd4f7183/spec/services/applications/accept_spec.rb#L239
https://github.com/DFE-Digital/npq-registration/blob/a95876e3e5e759830e4df37968182969fd4f7183/spec/services/applications/accept_spec.rb#L325

leandroalemao (approved):
@mooktakim well done! tks

PR #1489, opened 2024-06-28
[CPDLP-3239] Amend accept and change funding service to only accept funded place booleans only

leandroalemao:
### Context

We currently validate against the funded place attribute regardless of type. So if a string is passed in we compute it as true.
This confused providers as they were passing values such as `"false"` and `"null"` and this was computed as `true`

- Ticket: [CPDLP-3239](https://dfedigital.atlassian.net/browse/CPDLP-3239)

Amend the validations in both services to only accept boolean values otherwise returns a validation error, even if it's a string.

### Guidance to review
Maybe we should allow strings and compute those values, but felt better to lock it down to just booleans

[CPDLP-3239]: https://dfedigital.atlassian.net/browse/CPDLP-3239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/applications/accept.rb
+++ b/app/services/applications/accept.rb
@@ -6,15 +6,20 @@ module Applications
     include ActiveModel::Attributes
 
     attribute :application
-    attribute :funded_place, :boolean
+    attribute :funded_place
     attribute :schedule_identifier, :string
 
     validates :application, presence: { message: I18n.t("application.missing_application") }
+    validates :funded_place,
+              inclusion: {
+                in: [true, false],
+                if: :validate_funded_place?,
+                message: I18n.t("application.funded_place_required"),
+              }
     validate :not_already_accepted
     validate :cannot_change_from_rejected
     validate :other_accepted_applications_with_same_course?
     validate :eligible_for_funded_place
-    validate :validate_funded_place
     validate :validate_permitted_schedule_for_course
 
     def accept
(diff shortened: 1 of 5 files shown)


app/services/applications/accept.rb
```diff
@@ -127,11 +129,9 @@ def fallback_schedule
     def validate_permitted_schedule_for_course
       return if errors.any?
       return if schedule_identifier.blank?
-      return unless schedule
+      return if schedule&.course_group&.courses&.include?(course)
```
padv2010, line 132: Can we accept an application not belonging to a `CourseGroup`? Is it also possible to have a `schedule` that is `nil` at this point? If it is not possible, then it is probably better not to protect ourselves against `nil` values because we would be making the investigation of issues a bit more difficult.

leandroalemao: Well said @padv2010 .. I've just rolled back this change 👍 😄  tks

config/locales/en.yml
```diff
@@ -7,24 +7,21 @@ en:
...
   invalid_data_structure: correct json data structure required. See API docs for reference
-  schedule_invalid_for_course: Selected schedule is not valid for the course
 
   errors:
     email:
       invalid: Enter a valid email address
 
-  application:
-    missing_application: "The entered '#/application' is missing from your request. Check details and try again."
+  application: &application
```
cwrw, line 15: do you mind splitting these changes into another pr regarding the locales, as from the declaration pr I'm not sure all are working as expected, and prefer we have a ticket for them so we can test properly, let's chat tomorrow :)

cwrw (approved):
thanks @leandroalemao 🙌

PR #1490, opened 2024-06-28
Add GET `api:/version/participant/npq/outcomes` endpoints

ethax-ross:
[Jira-3097](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3097)

### Context

We want to bring across participant outcomes from ECF, starting with the `index` action. As part of this we will also need the base query and serializer.

- Add ParticipantOutcomes::Query

Add query service to retrieve `ParticipantOutcome` records filtered by `lead_provider` and `created_since`, which will be required for the `index` action.

- Add ParticipantOutcomeSerializer

Add the serializer for `ParticipantOutcome` so that we can return these models from the API endpoints.

- Refactor FilterByUpdatedSince to be more generic

We want to also filter by `created_since` in some API requests; rename to `FilterByDate` and add `created_since` method. Add dedicated unit tests.

- Add GET /api/:version/participants/npq/outcomes endpoints

Add the endpoints to get participant outcomes across all versions.

Slight reworking of the routes and controller names to correct a couple of things.

Adds new shared context for checking filtering by `created_since`.


app/models/participant_outcome.rb
```diff
@@ -6,6 +6,8 @@ class ParticipantOutcome < ApplicationRecord
   validates :completion_date, presence: true
   validate :completion_date_not_in_the_future
 
+  delegate :user, :lead_provider, to: :declaration
```
leandroalemao, line 9: If we add `:course` here as well, in the [serializer](https://github.com/DFE-Digital/npq-registration/pull/1490/files#diff-b9d35c56d14224cecdd555788c30d3ca50d92a62852df03a4a65fd1250f09463R11) we could do:

`field(:course_identifier) { |outcome| outcome.course.identifier }`

app/serializers/api/participant_outcome_serializer.rb
```diff
@@ -0,0 +1,34 @@
...
+    identifier :ecf_id, name: :id
+    field(:type) { "participant-outcome" }
+
+    class AttributesSerializer < Blueprinter::Base
+      exclude :id
+
+      field(:state)
+      field(:completion_date)
+      field(:course_identifier) { |outcome| outcome.declaration.application.course.identifier }
+      field(:participant_id) { |outcome| outcome.user.ecf_id }
```
leandroalemao, line 12: There's a N+1 query here..
think we need to add user table to the [query](https://github.com/DFE-Digital/npq-registration/pull/1490/files#diff-b6c95ab3ddda088375890402e92d6a94b1525517a04cc1668c3b27826edc2e3aR40)

`application: [:course, :user],`

leandroalemao (changes requested):
Great work! Very minor comments.. tkss 👍

app/serializers/api/participant_outcome_serializer.rb
ethax-ross: Good spot, I've updated 👍

leandroalemao (approved):
All good tks 👍 🔝

config/routes.rb
```diff
@@ -138,13 +140,11 @@
               put :withdraw
 
               scope module: :participants do
-                resources :outcomes, only: %i[create index], as: :participant_outcomes
+                resources :outcomes, only: %i[create index], as: :participants_outcomes
```
cwrw, line 143: So I think we need both lists one for a single participant and one for multiple, the decision to separate those was intentional:
- we return a list of all outcomes for all declarations
- we return a list of all outcomes for a single participant
- we create an outcome for a participant
happy to chat through

cwrw (approved):

PR #1498, opened 2024-07-02
Add create participant outcomes endpoint/service

ethax-ross:
[Jira-3099](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3099)

### Context

We want to bring across the create participant outcomes functionality from ECF into NPQ reg. This will not include posting updates to the qualified teachers API when outcomes are created.

-  Add ParticipantOutcomes::Create service

Port the service to create a `ParticipantOutcome` from ECF, replicating the functionality.

The `course_for_participant` validator has been omitted as its redundant; we require a completed declaration for the given `course_identifier` to submit the outcome which implies the participant has an application with the relevant course.

Add transient `course` to declarations factory to more easily create a declaration with a given course.

- Remove presence validation from ecf_id

We generate this at the database level, but if we don't omit the validation here it will prevent us from creating a new `ParticipantOutcome` without specifying an `ecf_id` upfront.

- Add POST api/:version/npq/participants/:id/outcomes endpoints

Add the endpoints to create a `ParticipantOutcome` for a given participant. Calls the `ParticipantOutcomes::Create` service.

### DIFF
--- a/app/services/participant_outcomes/create.rb
+++ b/app/services/participant_outcomes/create.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+module ParticipantOutcomes
+  class Create
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    STATES = %w[passed failed].freeze
+    COURSE_IDENTIFIERS = %w[npq-early-headship-coaching-offer npq-additional-support-offer].freeze
+    COMPLETION_DATE_FORMAT = /\d{4}-\d{2}-\d{2}/
+
+    attr_reader :created_outcome
+
+    attribute :lead_provider
+    attribute :participant
+    attribute :course_identifier
+    attribute :state
+    attribute :completion_date
+
+    validates :lead_provider, presence: true
+    validates :participant, presence: true
+    validates :course_identifier, inclusion: { in: COURSE_IDENTIFIERS }
+    validates :state, inclusion: { in: STATES }
+    validates :completion_date, presence: true, format: { with: COMPLETION_DATE_FORMAT }
+    validate :participant_has_no_completed_declarations
+    validate :completion_date_not_in_the_future
+
+    def create_outcome
+      return false unless valid?
+
+      ApplicationRecord.transaction do
+        @created_outcome = if outcome_already_exists?
+                             latest_existing_outcome
+                           else
+                             build_outcome.tap(&:save!)
+                           end
+      end
+
+      true
+    end
+
+  private
+
+    def outcome_already_exists?
+      return unless latest_existing_outcome
+
+      latest_existing_outcome.slice(:state, :completion_date) == build_outcome.slice(:state, :completion_date)
+    end
+
+    def build_outcome
+      @build_outcome ||= ParticipantOutcome.new(declaration: latest_completed_declaration, state:, completion_date:)
+    end
+
+    def completed_declarations
+      return Declaration.none unless participant
+
+      @completed_declarations ||= participant.declarations
+        .completed
+        .with_lead_provider(lead_provider)
+        .with_course_identifier(course_identifier)
+        .billable_or_voidable
+    end
+
+    def latest_completed_declaration
+      @latest_completed_declaration ||= completed_declarations.order(created_at: :desc).first
+    end
+
+    def latest_existing_outcome
+      @latest_existing_outcome ||= latest_completed_declaration.participant_outcomes.order(created_at: :desc).first
+    end
+
+    def participant_has_no_completed_declarations
+      errors.add(:base, :no_completed_declarations) unless completed_declarations.exists?
+    end
+
+    def completion_date_not_in_the_future
+      errors.add(:completion_date, :future_date) if completion_date&.to_date&.future?
+    end
+  end
+end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -28,6 +28,11 @@ en:
     cohort_does_not_accept_capping: "Leave the '#/funded_place' field blank. It's only needed for participants starting NPQs from autumn 2024 onwards."
     cannot_change_funded_place: "You must void or claw back your declarations for this participant before being able to set '#/funded_place' to false"
 
+  completion_date: &completion_date
+    future_date: The '#/%{attribute}' value cannot be a future date. Check the date and try again.
+    blank: "The '#/%{attribute}' is missing from your request. Please include a completion_date value and try again."
+    format: "The '#/%{attribute}' value must be in the following format: 'yyyy-mm-dd'"
+
   statement: &statement
     no_output_fee_statement: You cannot submit or void declarations for the %{cohort} cohort. The funding contract for this cohort has ended. Get in touch if you need to discuss this with us.
 
@@ -273,8 +278,20 @@ en:
         participant_outcomes:
           attributes:
             completion_date:
-              future_date: The '#/%{attribute}' value cannot be a future date. Check the date and try again.
-              
+              <<: *completion_date
+        participant_outcomes/create:
+          attributes:
+            base:
+              no_completed_declarations: "The participant has not had a 'completed' declaration submitted for them. Therefore you cannot update their outcome."
+            completion_date:
+              <<: *completion_date
+            course_identifier:
+              inclusion: "The entered '#/course_identifier' is not recognised for the given participant. Check details and try again."
+              invalid_course: "The entered '#/course_identifier' is not recognised for the given participant. Check details and try again."
+            state:
+              inclusion: "The attribute '#/%{attribute}' can only include 'passed' or 'failed' values. If you need to void an outcome, you will need to void the associated 'completed' declaration."
+            
+
   omniauth_providers:
     tra_openid_connect: "Get an Identity"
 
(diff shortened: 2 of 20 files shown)


spec/support/shared_examples/api_create_endpoint_support.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "an API create endpoint" do
```
ethax-ross, line 3: This is lifted from another open PR; I'll rebase when its merged in.

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,80 @@
...
+    attr_reader :created_outcome
+
+    attribute :lead_provider
+    attribute :participant
+    attribute :course_identifier
+    attribute :state
+    attribute :completion_date
+
+    validates :lead_provider, presence: true
+    validates :participant, presence: true
```
leandroalemao, line 21: When calling this service with `participant: nil` in the params I'm getting these error messages below. Think it should be another message maybe?

```
["Participant can't be blank",
 "The participant has not had a 'completed' declaration submitted for them. Therefore you cannot update their outcome."].
```

A bit different message when when course_identifier is nil as well:

```
service = ParticipantOutcomes::Create.new(
  lead_provider: Application.accepted.last.lead_provider,
  course_identifier: nil,
  participant: Application.accepted.last.user,
  completion_date: "2022-12-01",
  state: "passed",
)
service.valid?
service.errors.full_messages
```
```
["The participant has not had a 'completed' declaration submitted for them. Therefore you cannot update their outcome."]
```

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,80 @@
...
+    validates :course_identifier, inclusion: { in: COURSE_IDENTIFIERS }
+    validates :state, inclusion: { in: STATES }
+    validates :completion_date, presence: true, format: { with: COMPLETION_DATE_FORMAT }
```
leandroalemao, line 24: I'm seeing an exception being raised when I'm calling the service with an invalid date ie:
```
ParticipantOutcomes::Create.new(
  lead_provider: Application.accepted.last.lead_provider,
  course_identifier: Application.accepted.last.course.identifier,
  participant: Application.accepted.last.user,
  completion_date: "2022-13-01",
  state: "passed",
).valid?
```

leandroalemao (changes requested):
Hi @ethax-ross Added some comments but all good in general. Tks 👍

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    attribute :lead_provider
+    attribute :participant
+    attribute :course_identifier
+    attribute :state
+    attribute :completion_date
+
+    validates :lead_provider, presence: true
+    validates :participant, presence: true
+    validates :course_identifier, inclusion: { in: COURSE_IDENTIFIERS }
```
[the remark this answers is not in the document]
ethax-ross: Whoops, yep, good spot I'll update 👍

app/services/participant_outcomes/create.rb
ethax-ross: Ah, I didn't put a custom validation error in for the participant being blank as the service won't get called in that scenario from the controller. I'll add one in for completeness, same for `lead_provider`.

I couldn't replicate the issue with `course_identifier: nil`, I get the correct error (+ declarations error as the user has no declarations):

```
["Course identifier The entered '#/course_identifier' is not recognised for the given participant. Check details and try again.",
 "The participant has not had a 'completed' declaration submitted for them. Therefore you cannot update their outcome."]
```

app/services/participant_outcomes/create.rb
ethax-ross: Good spot, I'll update 👍

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,92 @@
...
+    def latest_completed_declaration
+      @latest_completed_declaration ||= completed_declarations.order(created_at: :desc).first
+    end
+
+    def latest_existing_outcome
+      @latest_existing_outcome ||= latest_completed_declaration.participant_outcomes.order(created_at: :desc).first
+    end
+
+    def participant_has_no_completed_declarations
+      errors.add(:base, :no_completed_declarations) unless completed_declarations.exists?
```
leandroalemao, line 75: It'd be nice to add a guard clause here `return if errors.any?` so we just return one error message at the time in case it exists.

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    def completion_date_not_in_the_future
+      return if errors.key?(:completion_date)
```
leandroalemao, line 79: Same as above..

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      errors.add(:completion_date, :future_date) if completion_date&.to_date&.future?
+    end
+
+    def completion_date_is_a_valid_date
+      return unless completion_date
```
leandroalemao, line 85: Same as above..

app/services/participant_outcomes/create.rb
leandroalemao, line 75: It'd be nice to add a guard clause here `return if errors.any?` so we just return one error message at the time in case of validation errors.

leandroalemao (changes requested):
Looking much better! Tks Ross 👍 😄

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,92 @@
...
+      return Declaration.none unless participant
+
+      @completed_declarations ||= participant.declarations
+        .completed
+        .with_lead_provider(lead_provider)
+        .with_course_identifier(course_identifier)
+        .billable_or_voidable
+    end
+
+    def latest_completed_declaration
```
cwrw, line 66: maybe we can move those queries into scopes, especially the latest outcome as it will be reused?

app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    attribute :lead_provider
+    attribute :participant
+    attribute :course_identifier
+    attribute :state
+    attribute :completion_date
+
+    validates :lead_provider, presence: true
+    validates :participant, presence: true
+    validates :course_identifier, inclusion: { in: PERMITTED_COURSES }
```
cwrw, line 23: could we also check for course identifier presence as well as state presence?

config/locales/en.yml
```diff
@@ -273,8 +281,24 @@ en:
...
+              <<: *completion_date
+        participant_outcomes/create:
+          attributes:
+            base:
+              no_completed_declarations: "The participant has not had a 'completed' declaration submitted for them. Therefore you cannot update their outcome."
+            completion_date:
+              <<: *completion_date
+            course_identifier:
+              inclusion: "The entered '#/course_identifier' is not recognised for the given participant. Check details and try again."
+              invalid_course: "The entered '#/course_identifier' is not recognised for the given participant. Check details and try again."
```
cwrw, line 293: this seems like duplicate error message to the above, is it needed?

spec/services/participant_outcomes/create_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+    it { is_expected.not_to allow_values(nil, "", 1.day.from_now.strftime(date_format), 1.day.ago.strftime("%d-%m-%Y")).for(:completion_date) }
+    it { is_expected.to validate_inclusion_of(:course_identifier).in_array(described_class::PERMITTED_COURSES) }
+    it { is_expected.to validate_inclusion_of(:state).in_array(described_class::STATES) }
+
+    context "when the completion date is a valid format but invalid date" do
+      let(:completion_date) { "2021-13-01" }
+
+      it "adds an error to the completion_date attribute" do
+        expect(instance).to be_invalid
+        expect(instance.errors.first).to have_attributes(attribute: :completion_date, type: :format)
```
cwrw, line 29: could we start adding the messages to the specs as discussed yesterday?

app/services/participant_outcomes/create.rb
cwrw: I think this isn't the case for all endpoints, we should have a look at ECF as we don't always return one error if all attributes are missing for example we need to return a list of errors, there are a few we avoid as they are confusing if added. So the suggestion is to add the returns where it makes the most sense

spec/support/matchers/error_matcher.rb
```diff
@@ -0,0 +1,7 @@
+RSpec::Matchers.define :have_error do |attribute, type, message|
+  match do |actual|
+    actual.invalid? && actual.errors.any? do |error|
```
cwrw, line 3: 👍

cwrw (approved):
🙌

leandroalemao (approved):
@ethax-ross great work! Tks

PR #1499, opened 2024-07-02
[CPDLP-3198] Add outcomes logic to create declaration endpoint

leandroalemao:
### Context

Ticket: [CPDLP-3198](https://dfedigital.atlassian.net/browse/CPDLP-3198)

When a provider submits a completed declaration we need to capture if they passed in an outcome.

Similar to ECF if a valid completed declaration has been created, create the outcome for that participant.

[CPDLP-3198]: https://dfedigital.atlassian.net/browse/CPDLP-3198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/declarations/create.rb
+++ b/app/services/declarations/create.rb
@@ -38,6 +38,7 @@ module Declarations
         set_eligibility!
 
         statement_attacher.attach unless declaration.submitted_state?
+        create_participant_outcome!
       end
 
       true
@@ -165,7 +166,25 @@ module Declarations
     end
 
     def valid_course_identifier_for_participant_outcome?
-      CourseGroup.joins(:courses).where(name: %w[leadership specialist]).map(&:courses).flatten.uniq.map(&:identifier).include?(course_identifier)
+      CourseGroup.joins(:courses).leadership_or_specialist.where(courses: { identifier: course_identifier }).exists?
+    end
+
+    def create_participant_outcome!
+      return unless validate_has_passed?
+
+      service = ParticipantOutcomes::Create.new(
+        lead_provider:,
+        participant:,
+        course_identifier:,
+        state: has_passed.to_s == "true" ? "passed" : "failed",
+        completion_date: declaration_date.rfc3339,
+      )
+
+      if service.valid?
+        service.create_outcome
+      else
+        raise ArgumentError, I18n.t(:cannot_create_completed_declaration)
+      end
     end
   end
 end
--- a/spec/models/declaration_spec.rb
+++ b/spec/models/declaration_spec.rb
@@ -203,7 +203,7 @@ RSpec.describe Declaration, type: :model do
         completed_declaration.dup.update!(declaration_type: "retained-1")
 
         # Declaration on another provider.
-        completed_declaration.dup.update!(lead_provider: create(:lead_provider, name: "Other lead provider"))
+        completed_declaration.dup.update!(lead_provider: LeadProvider.where.not(id: lead_provider.id).first)
 
         # Declaration with different course.
         completed_declaration.dup.update!(application: create(:application, course: create(:course, identifier: "other-course")))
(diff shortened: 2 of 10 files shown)


app/services/participant_outcomes/create.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module ParticipantOutcomes
```
cwrw, line 3: @leandroalemao this is already implemented in https://github.com/DFE-Digital/npq-registration/pull/1498/files#diff-f633d59f99e20b9c30469f9dff5e59e7612da3eca47b2055f7b9e1b05004635dR4 could we please hold off on this pr till that gets merged or branch off that one please?

leandroalemao: Putting this PR on hold for now till we get Ross' PR merged. Tks for flagging up @cwrw

app/models/participant_outcome.rb
```diff
@@ -1,6 +1,12 @@
 class ParticipantOutcome < ApplicationRecord
   belongs_to :declaration
 
+  enum state: {
```
cwrw, line 4: note that this has already been added in another pr https://github.com/DFE-Digital/npq-registration/pull/1513/files#diff-bd21f06a16ff911a3e6127ef0154f4712f292420f86ce416314274a32bb0dee3R11

leandroalemao: Hi @cwrw Remove it from here then tksss 👍 😄

app/services/declarations/create.rb
```diff
@@ -167,5 +168,23 @@ def validate_has_passed?
     def valid_course_identifier_for_participant_outcome?
       CourseGroup.joins(:courses).where(name: %w[leadership specialist]).map(&:courses).flatten.uniq.map(&:identifier).include?(course_identifier)
```
mooktakim, line 169: @leandroalemao I know this isn't part of this PR, I was thinking this would be better to do this in SQL:
```ruby
CourseGroup.joins(:courses).where(name: %w[leadership specialist], courses: {identifier: course_identifier}).exists?
```
Also maybe with scopes:
```ruby
CourseGroup.joins(:courses).leadership_or_specialist.where(courses: {identifier: course_identifier}).exists?
```

mooktakim (approved):

spec/models/declaration_spec.rb
```diff
@@ -203,7 +203,7 @@
         completed_declaration.dup.update!(declaration_type: "retained-1")
 
         # Declaration on another provider.
-        completed_declaration.dup.update!(lead_provider: create(:lead_provider, name: "Other lead provider"))
+        completed_declaration.dup.update!(lead_provider: LeadProvider.where.not(id: lead_provider.id).first)
```
cwrw, line 206: I've seen this change multiple times, can I ask why we are changing this?

app/services/declarations/create.rb
```diff
@@ -165,7 +166,25 @@ def validate_has_passed?
...
+        participant:,
+        course_identifier:,
+        state: has_passed.to_s == "true" ? "passed" : "failed",
+        completion_date: declaration_date.rfc3339,
+      )
+
+      if service.valid?
+        service.create_outcome
+      else
+        raise ArgumentError, I18n.t(:cannot_create_completed_declaration)
```
cwrw, line 186: in ECF we have been using this pattern however in NPQ we were hoping to move towards this failing at the initial `valid?` stage of the create declaration service rather than fail from within the transaction, is it possible to instantiate and check this service is valid when we do all other validations instead?

spec/models/declaration_spec.rb
leandroalemao: Hi @cwrw you ok? Still getting flakey tests when doing `create(:lead_provider, name: "Other lead provider")`.. and then no flakey tests with this change.. hope it's all good. Tks.

app/services/declarations/create.rb
leandroalemao: Hi @cwrw I tried that idea but the outcomes create service class relies on the completed declaration to exists as part of its own validation. If we move this code to be a validation in the declarations create class, the completed declaration won't yet be present and the outcomes service class will be always invalid.

cwrw (approved):

PR #1511, opened 2024-07-04
Upgrade Ruby to 3.3.4

leandroalemao:
### Context

Ticket: N/A

As Ruby 3.3.4 was [released](https://www.ruby-lang.org/en/news/2024/07/09/ruby-3-3-4-released/) we need to keep it as updated as possible.

Upgrade Ruby to 3.3.4

To install ruby 3.3.4 (M1 chip):
- `rvm install "ruby-3.3.4" --with-openssl-dir=$(brew --prefix openssl) --with-readline-dir=$(brew --prefix readline) --with-libyaml-dir=$(brew --prefix libyaml) --disable-dtrace --disable-docs`
- `gem install iconv -- --with-cflags="-Wno-incompatible-function-pointer-types"`
- `bundle install`


leandroalemao:
Think it needs to get node version upgraded firstly.

ethax-ross (approved):
Looks good! nothing stands out in the Ruby release notes that would cause us issues, but it might be worth someone on NPQ reg eyeballing this as well 👍

.github/workflows/lint.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         name: Checkout Code
 
       - name: Set up Ruby
-        uses: ruby/setup-ruby@v1.154.0
```
cwrw, line 28: any reason why we're removing those here and below?

leandroalemao: Just following recommendation to get latest ruby version always.. https://github.com/ruby/setup-ruby?tab=readme-ov-file#versioning

.github/workflows/lead_provider_openapi_check.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       DB_PORT: 5432
       DATABASE_URL: postgres://postgres:postgres@localhost:5432
       SECRET_KEY_BASE: abc123
+      NODE_OPTIONS: --openssl-legacy-provider
```
peteryates, line 31: As it's not obvious, this looks like it's needed [because of a bug in WebPack](https://github.com/webpack/webpack/issues/14532#issuecomment-947515866).

peteryates (approved):

cwrw (approved):

.github/workflows/lead_provider_openapi_check.yml
leandroalemao: 👍 Yeah took a good while to figure this out as the build was broken without it

mooktakim (approved):

leoapost:
#1526 should replace this PR.

cwrw:
@leoapost could we merge this separately and you can continue with your work?

leoapost:
> @leoapost could we merge this separately and you can continue with your work?

#1526 is ready for review. I don't think it worth merging this PR anymore.

leandroalemao:
Merging this for now as altho @leoapost's PR includes same changes as this... think it's a bit more complex and big as it replaces webpacker and sure requires more testing.. and it's better anyway to have all changes split in two anyway.

PR #1513, opened 2024-07-05
Add `has_passed` to `DeclarationSerializer`

ethax-ross:
[Jira-3236](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3236)

### Context

We want to bring the `has_passed?` logic from ECF into `ParticipantOutcome` and expose in the `DeclarationSerializer`. The `has_passed` will be `nil` if the outcome does not exist or is `voided`, otherwise `true`/`false` for `passed` and `failed`.

- Add has_passed to DeclarationSerializer

### DIFF
--- a/app/models/participant_outcome.rb
+++ b/app/models/participant_outcome.rb
@@ -8,8 +8,20 @@ class ParticipantOutcome < ApplicationRecord
 
   delegate :user, :lead_provider, :course, to: :declaration
 
+  enum state: {
+    passed: "passed",
+    failed: "failed",
+    voided: "voided",
+  }, _suffix: true
+
   def self.latest
     order(created_at: :desc).first
+  end
+
+  def has_passed?
+    return nil if voided_state?
+
+    passed_state?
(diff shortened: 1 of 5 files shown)


app/models/participant_outcome.rb
```diff
@@ -8,10 +8,22 @@ class ParticipantOutcome < ApplicationRecord
 
   delegate :user, :lead_provider, :course, to: :declaration
 
+  enum state: {
```
leandroalemao, line 11: Maybe added a test for this on `spec/models/participant_outcome_spec.rb`?

```
  describe "enums" do
    it {
      expect(subject).to define_enum_for(:state).with_values(
        passed: "passed",
        failed: "failed",
        voided: "voided",
      ).backed_by_column_of_type(:enum).with_suffix
    }
  end
```

leandroalemao (approved):

mooktakim (approved):

PR #1524, opened 2024-07-12
[CPDLP-3199] TS8 - Add outcomes logic to void declaration endpoint

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3199

* Created `ParticipantOutcomes::Void` service
* Updated `Declarations::Void` to run `ParticipantOutcomes::Void`

### DIFF
--- a/app/services/declarations/void.rb
+++ b/app/services/declarations/void.rb
@@ -22,6 +22,8 @@ module Declarations
 
       ApplicationRecord.transaction do
         clawing_back? ? clawback_declaration : void_declaration
+
+        void_participant_outcome
       end
 
       true
@@ -70,6 +72,10 @@ module Declarations
 
     def voiding?
       !clawing_back?
+    end
+
+    def void_participant_outcome
+      ParticipantOutcomes::Void.new(declaration:).void_outcome
     end
   end
 end
--- a/app/services/participant_outcomes/void.rb
+++ b/app/services/participant_outcomes/void.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module ParticipantOutcomes
+  class Void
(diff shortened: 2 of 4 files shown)


app/models/course_group.rb
```diff
@@ -4,6 +4,8 @@ class CourseGroup < ApplicationRecord
 
   validates :name, presence: { message: "Enter a unique course group name" }, uniqueness: { message: "Course name already exist, enter a unique name" }
 
+  scope :leadership_or_specialist, -> { where(name: %w[leadership specialist]) }
```
leandroalemao, line 7: Hi @mooktakim Could you please add a spec for this scope? I followed your suggestion on the other PR and added a spec in there.. so whatever PR gets merged first, the scope will be in.. tks
https://github.com/DFE-Digital/npq-registration/pull/1499/files#diff-be233ac6eae795dbfb4a1a24629eef68a767d6875377e03a5c7d1b7cc84c3dd6R80

mooktakim: @leandroalemao i was planning on merging your ticket first as its almost ready to be merged. 👍

mooktakim: I'll copy in the test anyway, it should be easy to merge.

cwrw: This has been added now in three prs, @leandroalemao maybe we can open a separate pr for this scope and merge so the others can rebase those changes?

cwrw (approved):

app/models/course_group.rb
leandroalemao: 👍 I've got a PR to merge tomorrow which includes this changes .. so it should be good after that.

PR #1519, opened 2024-07-10
[CPDLP-3102] Add outcomes seed data to separation

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3102

We have a service to create seed data for providers in separation which we need to keep up to date.

### Changes proposed in this pull request

Add outcomes to some accepted applications with completed declarations in the seed data.

PR #1525, opened 2024-07-12
[CPDLP-3269] Add data required to support NPQ MVP testing

leandroalemao:
### Context

Ticket: [CPDLP-3269](https://dfedigital.atlassian.net/browse/CPDLP-3269)

We need to revisit the rake task to setup seed data to cover all the set up specs below.

New data needs to be set up to support NPQ testing, some of which has already been set up in the rake task, but we’d like to beef up the scripts as per in the ticket description.

[CPDLP-3269]: https://dfedigital.atlassian.net/browse/CPDLP-3269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/valid_test_data_generators/applications_populater.rb
+++ b/app/services/valid_test_data_generators/applications_populater.rb
@@ -18,6 +18,7 @@ module ValidTestDataGenerators
       logger.info "ApplicationsPopulater: Started!"
 
       ActiveRecord::Base.transaction do
+        prepare_cohort!
         create_participants!
       end
 
@@ -41,6 +42,13 @@ module ValidTestDataGenerators
         travel_to(rand(3.years.ago..Time.zone.now)) do
           create_participant(school: School.open.order("RANDOM()").first)
         end
+      end
+    end
+
+    def prepare_cohort!
+      cohort.tap do |c|
+        c.funding_cap = cohort.start_year >= 2024
+        c.save!
       end
     end
 
@@ -143,7 +151,7 @@ module ValidTestDataGenerators
       return unless completed_declaration
       return unless CourseGroup.joins(:courses).leadership_or_specialist.where(courses: { identifier: application.course.identifier }).exists?
 
-      ParticipantOutcome.states.keys.sort.each do |state_trait|
+      ParticipantOutcomes::Create::STATES.reverse.each do |state_trait|
         FactoryBot.create(
           :participant_outcome,
           state_trait,
@@ -151,7 +159,7 @@ module ValidTestDataGenerators
           completion_date: completed_declaration.declaration_date,
         )
 
-        break if Faker::Boolean.boolean(true_ratio: 0.3)
+        break if Faker::Boolean.boolean(true_ratio: 0.2)
       end
     end
   end
--- a/app/services/valid_test_data_generators/applications_populater.rb
+++ b/app/services/valid_test_data_generators/applications_populater.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
+require "active_support/testing/time_helpers"
+
 module ValidTestDataGenerators
   class ApplicationsPopulater
+    include ActiveSupport::Testing::TimeHelpers
+
     class << self
       def populate(lead_provider:, cohort:, number_of_participants: 50)
         new(lead_provider:, cohort:, number_of_participants:).populate
@@ -33,7 +37,11 @@ module ValidTestDataGenerators
     end
 
     def create_participants!
-      number_of_participants.times { create_participant(school: School.open.order("RANDOM()").first) }
+      number_of_participants.times do
+        travel_to(rand(3.years.ago..Time.zone.now)) do
+          create_participant(school: School.open.order("RANDOM()").first)
+        end
+      end
     end
 
     def create_participant(school:)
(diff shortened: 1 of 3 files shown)


spec/models/course_group_spec.rb
```diff
@@ -76,4 +76,15 @@
       end
     end
   end
+
+  describe "scopes" do
+    describe ".leadership_or_specialist" do
+      let!(:leadership_group) { CourseGroup.find_by(name: "leadership") || create(:course_group, name: "leadership") }
+      let!(:specialist_group) { CourseGroup.find_by(name: "specialist") || create(:course_group, name: "specialist") }
```
ethax-ross, line 83: Can we add a course group that isn't leadership/specialist to ensure its not returning?

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -33,7 +37,11 @@ def initialize(lead_provider:, cohort:, number_of_participants:, logger: Rails.l
     end
 
     def create_participants!
-      number_of_participants.times { create_participant(school: School.open.order("RANDOM()").first) }
+      number_of_participants.times do
+        travel_to(rand(3.years.ago..Time.zone.now)) do
```
ethax-ross, line 41: TIL you can use date times with `rand`!

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -94,17 +110,48 @@ def reject_application(application)
...
           :declaration,
-          :submitted_or_eligible,
+          state: Declaration.states.keys.sample,
           application:,
           declaration_type:,
         )
       end
     end
+
+    def create_outcomes(application)
```
ethax-ross, line 140: Is this used?

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -94,17 +110,48 @@ def reject_application(application)
...
+      completed_declaration = application.declarations.eligible_for_outcomes(lead_provider, application.course.identifier).first
+      return unless completed_declaration
+      return unless CourseGroup.joins(:courses).leadership_or_specialist.where(courses: { identifier: application.course.identifier }).exists?
+
+      %i[passed voided failed].each do |state_trait|
```
cwrw, line 145: could we use the create states similar to https://github.com/DFE-Digital/npq-registration/pull/1519/files#diff-38f0cbe9d749dd85fe13a256c7c104dfc5a070dbf1a7e0e205b6c9b80df3ff19R115 :)

app/services/valid_test_data_generators/applications_populater.rb
leandroalemao: 👍  Rebased with main and now it's all good.. tks @ethax-ross

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -111,13 +141,15 @@ def create_declarations(application)
     def create_outcomes(application)
       completed_declaration = application.declarations.eligible_for_outcomes(lead_provider, application.course.identifier).first
       return unless completed_declaration
+      return unless CourseGroup.joins(:courses).leadership_or_specialist.where(courses: { identifier: application.course.identifier }).exists?
 
-      ParticipantOutcomes::Create::STATES.reverse.each do |state|
```
cwrw, line 115: can we use the old version (this one) please, as otherwise we would create voided outcomes without creating a previous outcome first, also a voided outcome should only be attached to a voided declaration?

ethax-ross (commented):
Looks good, just a few cases missing from the ticket when I checked the data:

```
Applications with 2025 cohort (should only be 21-24 according to ticket?)
Funded place differs slightly; true/false for 2021-2025, nil for 2024-2025
Participant outcome missing `nil` state
Declarations missing retained-3/4
```

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -44,6 +45,13 @@ def create_participants!
       end
     end
 
+    def prepare_cohort!
+      cohort.tap do |c|
+        c.funding_cap = cohort.start_year >= 2024
```
ethax-ross, line 50: Can we use something like `Cohort.current.start_year` here instead of hard coding?

cwrw (approved):

ethax-ross (commented):
I'm still seeing some discrepancies with the ticket specs?

Participant outcome missing `voided` state (the ticket mentions `null` but I don't think that's actually valid?):

```
[3] pry(main)> ParticipantOutcome.pluck(:state).uniq
=> ["failed", "passed"]
```
In the ticket it mentions true/false for funded place in 2024 and `nil` for 21-23; we have this but we also have `nil` for 2024 and true/false for 21-23, not sure if that's a problem or not:

```
[10] pry(main)> Application.all.map { |a| [a.cohort.start_year, a.funded_place] }.sort { |a, b| a.f
irst <=> b.first }.uniq
[2021, nil],
[2021, true],
[2022, nil],
[2022, false],
[2022, true],
[2023, nil],
[2023, true],
[2024, false],
[2024, nil],
[2024, true],
[2025, nil],
[2025, true]
```

Declarations are missing retained-3/4, but they don't appear to be in the enum either so maybe that's fine?

```
[12] pry(main)> Declaration.pluck(:declaration_type).uniq
=> ["started", "retained-1", "retained-2", "completed"]
```

cwrw:
will wait for @ethax-ross comments to be addressed sorry :)

cwrw:
@ethax-ross @leandroalemao just regarding Ross's comment, we don't have retained 3/4 in NPQ so that can be ignored, also funded place if accepted for 2024 will have to be true/false otherwise is `nil`, for 2021-2023 needs to be `nil` always regardless of accepted or not.
For voided outcomes we need to void completed declarations which then will void outcomes only

ethax-ross (approved):

PR #1527, opened 2024-07-16
Add `bullet` gem

ethax-ross:
### Context

We want to use `bullet` to identify and alert on N+1 queries, so that we don't have to manually check for them.

- Add `bullet` gem

Add `bullet` to identify and report N+1 queries. Enabled in `development` and will raise an error in `test` if an N+1 is detected.

- Precompile assets using prod environment

We bundle the production assets and run the precompile using the test env, which doesn't make sense. It causes an issue now that we have `bullet` referenced in the `test.rb` as it installed installed with the production gems.

Switch to compiling assets using the prod environment. We need to specify a `SECRET_KEY_BASE` for this to complete, but it doesn't matter what it is.

### Guidance to review

Awkwardly due to the way we are mocking out the underlying services in our request specs and the way our queries are structured (to perform the includes), `bullet` won't actually catch many N+1s from the API 😞 we may want to rethink if we want a robust way to detect them, but there's no harm in adding it in.

Off the top of my head we could hit the query services as part of the serializer specs and that would cover _most_ N+1's we're likely to see.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,7 @@ RUN yarn install --frozen-lockfile
 COPY . /app/
 
 # Compile assets and run webpack
-# Run in rails test environment to avoid loading development gems
-RUN RAILS_ENV=test bundle exec rails assets:precompile
+RUN RAILS_ENV=production SECRET_KEY_BASE=key bundle exec rails assets:precompile
 
 # Cleanup to save space in the production image
 RUN rm -rf node_modules log tmp && \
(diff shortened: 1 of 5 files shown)


peteryates:
> bullet won't actually catch many N+1s from the API 😞 we may want to rethink if we want a robust way to detect them, but there's no harm in adding it in.

Improving this would be a worthwhile post-separation task.

mooktakim (approved):

cwrw:
> Improving this would be a worthwhile post-separation task.
definitely, will capture that for our testing improvements

PR #1528, opened 2024-07-17
Add X-With-Server-Date header support

ethax-ross:
### Context

In ECF we allow the server-time to be changed per-request by specifying an `X-With-Server-Header` header.

We want to replicate this functionality in NPQ for all environments apart from production.

- Add X-With-Server-Date header support

In ECF we allow the server-time to be changed per-request by specifying an `X-With-Server-Header` header.

We want to replicate this functionality in NPQ for all environments apart from production.

Add `time_traveler` middleware from ECF and update it to use `ActiveSupport::Testing::TimeHelpers` instead of `TimeCop`. Load middleware into all environments, but disable in production.

### Guidance to review

I opted to load the middleware in all environments by default to avoid having to manually include it into new environments individually. The middleware itself will only run if its in a non-production environment.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -34,5 +34,9 @@ module NpqRegistration
     config.action_mailer.notify_settings = {
       api_key: ENV["GOVUK_NOTIFY_API_KEY"]
     }
+
+    # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
+    require "middleware/time_traveler"
+    config.middleware.use Middleware::TimeTraveler
   end
 end
--- a/lib/middleware/time_traveler.rb
+++ b/lib/middleware/time_traveler.rb
@@ -1,0 +1,19 @@
+require "active_support/testing/time_helpers"
+
+module Middleware
+  class TimeTraveler
+    include ActiveSupport::Testing::TimeHelpers
+
+    def initialize(app)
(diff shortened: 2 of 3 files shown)


cwrw (approved):

config/application.rb
```diff
@@ -34,5 +34,9 @@ class Application < Rails::Application
     config.action_mailer.notify_settings = {
       api_key: ENV["GOVUK_NOTIFY_API_KEY"]
     }
+
+    # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
+    require "middleware/time_traveler"
```
mooktakim, line 39: Can we move this to the top of the file as its more common location

mooktakim (approved):

config/application.rb
ethax-ross: You get an error if you do this, I think because we're in the `application.rb` file the Rails auto loader hasn't setup the load paths for our application files yet (but it will by the time its evaluating the class). You can `require_relative` at the top of the file, but then you have to put the full `../lib/middleware` path which didn't feel as nice as this.

mooktakim: Ah ok, thats fine 👍

leandroalemao (approved):

PR #1532, opened 2024-07-18
Fix query scopes for `show` actions

ethax-ross:
### Context

We were using the same `Query` instance for both the `index` and `show` actions; this meant the filtering was being applied to the `show` endpoints when it shouldn't be (although this would only happen if a filter parameter was passed in).

- Fix query scopes for show actions

Extract the query `conditions` so that we only surface them in the `index` actions.


leandroalemao (approved):
Looking good tks 👍 🔝

PR #1535, opened 2024-07-19
Clean up service error tests and translations

ethax-ross:
[Jira-3244](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3244)

### Context

Update the API service class specs to use the `have_error` matcher.

Clean up the translations file to make it consistent.

- Clean up service error tests and translations

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -33,6 +33,8 @@ class Application < ApplicationRecord
   scope :accepted, -> { where(lead_provider_approval_status: "accepted") }
   scope :eligible_for_funding, -> { where(eligible_for_funding: true) }
 
+  validate :schedule_cohort_matches
+
   enum kind_of_nursery: {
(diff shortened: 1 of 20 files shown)


cwrw:
@ethax-ross a few failures I think from the other pr possibly?

leandroalemao (changes requested):
@ethax-ross that's good! Just a small comment.. Tkss

https://github.com/DFE-Digital/npq-registration/blob/6e12b7638850eaecac082998746eafb35d56f776/spec/services/applications/accept_spec.rb#L210

```
expect(service).to have_error(:application, :cannot_change_from_rejected, "Once rejected an application cannot change state")
```

cwrw (approved):

leandroalemao (approved):
Tkss @ethax-ross

PR #1540, opened 2024-07-22
Clean up factories

ethax-ross:
[Jira-3283](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3283)

### Context

We want to tidy up the factories with consistent trait naming, associations, filenames, etc.

- Clean up factories

- Add spec to ensure factories are valid

Add spec to ensure all factories can be built and are valid with all traits.

Fix flakey `course_group` factory (for some reason `find_by || create` was giving errors; `find_or_create_by` seems more reliable, but I'm not sure why).

Fix ECF factory that requires an accept application to be valid (so create one by default).

- Remove abbreviations from course names

Use full course names for clarity.


db/seeds/base/add_applications.rb
```diff
@@ -125,7 +125,7 @@
       :accepted,
       :with_random_user,
       :with_random_work_setting,
-      :with_random_eligibility_for_funding,
+      :with_random_eligibile_for_funding,
```
mooktakim, line 128: Spelling mistake

spec/factories/courses.rb
```diff
@@ -9,90 +9,90 @@
       Course.find_by(identifier:) || new(**attributes)
     end
 
-    trait :ehco do
+    trait :early_headship_coaching_offer do
       sequence(:name) { |n| "NPQ Early Headship Coaching Offer Course #{n}" }
       identifier { "npq-early-headship-coaching-offer" }
-      course_group { CourseGroup.find_by(name: "ehco") || create(:course_group, name: "ehco") }
+      course_group { CourseGroup.find_or_create_by(name: "ehco") }
```
mooktakim, line 15: As we're using `initialize_with` for a lot of the factories now, can we also use it for `CourseGroup`, that way we can just use `create(:course_group, name: "ehco")` here and the other traits below.

spec/factories/statements.rb
```diff
@@ -4,8 +4,8 @@
     year { Faker::Number.between(from: 2021, to: 2024) }
     deadline_date { Faker::Date.forward(days: 30) }
     payment_date { Faker::Date.forward(days: 30) }
-    cohort { association :cohort }
-    lead_provider { association :lead_provider }
+    cohort
```
mooktakim, line 7: I think we should default to `cohort { create(:cohort, :current) }` otherwise some validations can break when a random year is picked from the future.

spec/factory_integrity_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe "Factory integrity check" do
+  FactoryBot.factories.each do |factory|
```
cwrw, line 4: 🌟

cwrw (approved):

db/seeds/base/add_course_groups.rb
```diff
@@ -3,6 +3,6 @@
   specialist
   support
   ehco
-].each do |n|
-  CourseGroup.find_or_create_by!(name: n)
+].each do |name|
+  create(:course_group, name:)
```
leandroalemao, line 7: Think we need explicitly add FactoryBot here otherwise rails seeds task fails.

`FactoryBot.create(:course_group, name:)`

leandroalemao (changes requested):
All good!! added only a small comment 👍 🔝 tksss

db/seeds/base/add_course_groups.rb
ethax-ross: Good spot, added 👍

leandroalemao (approved):

mooktakim (approved):

PR #1542, opened 2024-07-22
[CPDLP-3235] Enable sentry on separation env

leandroalemao:
### Context

Ticket: [CPDLP-3235](https://dfedigital.atlassian.net/browse/CPDLP-3235)

We need to track logs and debug exceptions on the new separation environment. Also we need to turn off integration with ECF Api.

- Enable Sentry and Logit on separation environment.
- Disable the ECF API only in separation ENV and remove the keys from the keyvault/setting to nil
- Set NPQ funding eligibility to read from previously funded if ECF API is disabled

[CPDLP-3235]: https://dfedigital.atlassian.net/browse/CPDLP-3235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/environments/migration.rb
+++ b/config/environments/migration.rb
@@ -11,6 +11,6 @@ Rails.application.configure do
     admin_portal_enabled: true,
     api_enabled: true,
     migration_enabled: true,
-    ecf_api_disabled: false,
+    ecf_api_disabled: true,
   }
 end
--- a/config/environments/migration.rb
+++ b/config/environments/migration.rb
@@ -11,5 +11,6 @@ Rails.application.configure do
     admin_portal_enabled: true,
     api_enabled: true,
     migration_enabled: true,
+    ecf_api_disabled: false,
   }
 end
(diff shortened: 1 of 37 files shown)


config/environments/migration.rb
```diff
@@ -11,5 +11,6 @@
     admin_portal_enabled: true,
     api_enabled: true,
     migration_enabled: true,
+    ecf_api_disabled: false,
```
cwrw, line 14: could we disable it in migration as well, it might be a good idea :)

cwrw (commented):
thanks @leandroalemao looking good, I have two final comments:
- Could you add tests for all the ECF services for when the API is disabled please?
- Could we also add a return and a test here please: https://github.com/DFE-Digital/npq-registration/blob/main/app/models/user.rb#L119 and https://github.com/DFE-Digital/npq-registration/blob/main/app/services/participant_validator.rb#L11

cwrw (approved):
thanks @leandroalemao 🙌

PR #1544, opened 2024-07-23
[CPDLP-3282] NPQ MVP Prep - Review and add missing validations for services and models

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3282

* Moved validations from service classes to the model related to the specific columns.
* Updated factories to be valid

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -72,6 +72,11 @@ class Application < ApplicationRecord
     withdrawn: "withdrawn",
   }
 
+  validates :funded_place,
+            inclusion: { in: [true, false] },
+            if: :validate_funded_place?
+  validate :eligible_for_funded_place
+
   # `eligible_for_dfe_funding?`  takes into consideration what we know
   # about user eligibility plus if it has been previously funded. We need
   # to keep this method in place to keep consistency during the split between
@@ -181,5 +186,18 @@ private
 
   def schedule_cohort_matches
     errors.add(:schedule, :cohort_mismatch) if schedule && schedule.cohort != cohort
+  end
+
+  def validate_funded_place?
+    accepted? && errors.blank? && cohort&.funding_cap?
+  end
+
+  def eligible_for_funded_place
+    return if errors.any?
+    return unless cohort&.funding_cap?
+
+    if funded_place && !eligible_for_funding
+      errors.add(:funded_place, :not_eligible)
+    end
   end
 end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -76,6 +76,7 @@ class Application < ApplicationRecord
             inclusion: { in: [true, false] },
             if: :validate_funded_place?
   validate :eligible_for_funded_place
+  validate :validate_permitted_schedule_for_course
 
   # `eligible_for_dfe_funding?`  takes into consideration what we know
   # about user eligibility plus if it has been previously funded. We need
@@ -198,6 +199,15 @@ private
 
     if funded_place && !eligible_for_funding
       errors.add(:funded_place, :not_eligible)
+    end
+  end
+
+  def validate_permitted_schedule_for_course
+    return if errors.any?
+    return unless accepted? && schedule && course
+
+    unless schedule.course_group.courses.include?(course)
+      errors.add(:schedule, :invalid_for_course)
     end
   end
 end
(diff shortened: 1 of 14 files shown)


spec/models/declaration_spec.rb
```diff
@@ -16,6 +16,39 @@
...
+      before { subject.declaration_date = subject.application.schedule.applies_from.prev_week }
+
+      it "has a meaningful error" do
+        expect(subject).to be_invalid
+        expect(subject).to have_error(:declaration_date, :declaration_before_schedule_start, "Enter a '#/declaration_date' that's on or after the schedule start.")
+      end
+    end
+
+    context "when declaration_date is at the schedule start" do
+      let(:schedule_applies_from_date) { declaration_date }
```
ethax-ross, line 48: This isn't used anywhere; I think this should be something like:

```
before { subject.declaration_date = subject.application.schedule.applies_from }
```

I'm surprised rubocop didn't flag it 🤔

app/services/valid_test_data_generators/applications_populater.rb
```diff
@@ -39,9 +39,7 @@ def initialize(lead_provider:, cohort:, number_of_participants:, logger: Rails.l
 
     def create_participants!
```
ethax-ross, line 40: Why is the applications populater changing as part of this?

app/models/application.rb
```diff
@@ -176,4 +181,17 @@ def funding_eligibility(with_funded_place:)
 
     eligible_for_funding && (funded_place.nil? || funded_place)
   end
+
+  def validate_funded_place?
+    accepted? && errors.blank? && cohort&.funding_cap?
+  end
+
+  def eligible_for_funded_place
```
ethax-ross, line 189: Is it worth having these shared validations in a concern so we're not duplicating them?

ethax-ross (commented):
I think there are more validation rules that could be moved to the models; I'm not sure how far we want to go with it though... some feel like they would make sense, for example:

- application -> validate_permitted_schedule_for_course (accept)
- participant outcome -> completion_date_not_in_the_future (create)

I'm less sure about the more service-specific ones though, such as:

- application -> cannot_change_from_rejected (accept)
- application -> cannot_change_from_accepted (reject)

app/services/valid_test_data_generators/applications_populater.rb
mooktakim: As we're creating seed data for different cohorts, validations related to dates get triggered for declarations. I had to move the `travel_to` so we create declarations that are after `schedule.applies_from`.

app/models/application.rb
mooktakim: There is slight differences to both methods, including how the errors are added. I'm not sure if its worth splitting it out.

mooktakim:
> I think there are more validation rules that could be moved to the models; I'm not sure how far we want to go with it though... some feel like they would make sense, for example:
> 
> * application -> validate_permitted_schedule_for_course (accept)
> * participant outcome -> completion_date_not_in_the_future (create)
> 
> I'm less sure about the more service-specific ones though, such as:
> 
> * application -> cannot_change_from_rejected (accept)
> * application -> cannot_change_from_accepted (reject)

The `completion_date_not_in_the_future` for participant outcome already exists.

`validate_permitted_schedule_for_course` will look to include this

app/services/valid_test_data_generators/applications_populater.rb
ethax-ross: Makes sense 👍 I think we need to keep some of the randomness here though; so previously this was creating participants on random dates over the last 3 years whereas now the seeds would all have the same created at?

mooktakim: Sure, I've tried to maintain some randomness with this line:
https://github.com/DFE-Digital/npq-registration/pull/1544/files#diff-38f0cbe9d749dd85fe13a256c7c104dfc5a070dbf1a7e0e205b6c9b80df3ff19R61

ethax-ross (approved):
Looks good 👍 I'm not sure if we need to have a think about if we should be adding some of the more service-specific validations into the core models or not, but that could be a follow-up PR.

cwrw (approved):
thanks @mooktakim, I think might be worth testing the data generator on review app for the PO review incase it changes anything with the randomness, but looks good otherwise

PR #1549, opened 2024-07-24
Fix N+1 in DeclarationSerializer

ethax-ross:
[Jira-3336](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3336)

### Context

The `DeclarationSerializer` was performing a second query for each declaration in order to find the latest outcome. Instead, we can preload the outomes in the query and perform an in-memory `sort_by` (instead of `order` which triggers another query) to find the latest outcome.

- Fix N+1 in DeclarationSerializer

### DIFF
--- a/app/models/participant_outcome.rb
+++ b/app/models/participant_outcome.rb
@@ -14,10 +14,6 @@ class ParticipantOutcome < ApplicationRecord
     voided: "voided",
   }, _suffix: true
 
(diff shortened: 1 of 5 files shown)


ethax-ross:
Closing as github doesn't like the `+` in the branch name - will re-open

PR #1551, opened 2024-07-24
Setup Rack::Attack for API requests

ethax-ross:
[Jira-3337](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3337)

### Context

We want to mirror the rate limiting for API requests from ECF.

- Add rate limiting for `/api` requests that matches the ECF configuration.

Clean up and improve the testing around rate limiting.

Remove redundant protected routes (they no longer appear to exist).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -330,9 +330,9 @@ GEM
     nokogiri (1.16.6)
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
-    nokogiri (1.16.6-x86_64-darwin)
+    nokogiri (1.16.6-arm64-darwin)
       racc (~> 1.4)
-    nokogiri (1.16.6-x86_64-linux)
+    nokogiri (1.16.6-x86_64-darwin)
       racc (~> 1.4)
     notifications-ruby-client (5.4.0)
       jwt (>= 1.5, < 3)
@@ -463,6 +463,8 @@ GEM
     rouge (4.3.0)
     rspec-core (3.13.0)
       rspec-support (~> 3.13.0)
+    rspec-default_http_header (0.0.6)
+      rspec-rails (> 3.0)
     rspec-expectations (3.13.1)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.13.0)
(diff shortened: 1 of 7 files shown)


Gemfile.lock
```diff
@@ -647,9 +649,9 @@ GEM
     zlib (3.1.0)
 
 PLATFORMS
+  arm64-darwin-23
   ruby
   x86_64-darwin-20
-  x86_64-linux
```
mooktakim, line 652: I think we need `x86_64-linux` for the azure deployment environment

config/initializers/rack_attack.rb
```diff
@@ -2,26 +2,32 @@
...
   end
 
   # Throttle /csp_reports requests by IP (5rpm)
   throttle("csp_reports req/ip", limit: 5, period: 1.minute) do |req|
     req.ip if req.path == "/csp_reports"
   end
+
+  # Throttle /api requests by auth token (1000 requests per 5 minutes)
+  throttle("API requests by auth token", limit: 1000, period: 5.minutes) do |request|
+    if request.path.starts_with?("/api/")
```
cwrw, line 24: do we need to worry about the internal APIs here or this won't affect anything?

ethax-ross: I think we're ok; I checked all the /api/ routes and the only others that stood out were api docs and guidance, so I've added exclusions for those (they will fall back to the general requests by ip).

config/initializers/rack_attack.rb
```diff
@@ -2,26 +2,32 @@
...
+      request.get_header("HTTP_AUTHORIZATION")
+    end
```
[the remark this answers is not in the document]
ethax-ross: Good spot; I've added a rule to rate limit this by IP address; previously it would have been caught by the general 300 per 5 minute rule but I've kept it consistent with our 1000 per 5 minute rule for other API requests.

Gemfile.lock
ethax-ross: I'm not sure why this was removed when I did `bundle install` 🤔 I think it would be added back in as part of the docker build as that does its own `bundle install` but I'll add it back in manually

mooktakim (approved):

cwrw (approved):

PR #1553, opened 2024-07-24
[CPDLP-3342] Do not use the fallback schedule if wrong identifier is sent in the params to api v3 accept application endpoint

leandroalemao:
### Context

Ticket: [CPDLP-3342](https://dfedigital.atlassian.net/browse/CPDLP-3342)

In ECF if the `schedule_identifier` passed into the accept endpoint is invalid we return the error message: 

`Selected schedule is not valid for the course`

in NPQ it is instead using the fallback schedule.

Amend the accept application service to return the above error if the wrong schedule is passed in in v3. Ensure that if the schedule is empty we do fallback, only return error if it’s invalid.

[CPDLP-3342]: https://dfedigital.atlassian.net/browse/CPDLP-3342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/applications/accept_spec.rb
+++ b/spec/services/applications/accept_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Applications::Accept, :with_default_schedules, type: :model do
   describe "#accept" do
     let(:trn) { rand(1_000_000..9_999_999).to_s }
     let(:user) { create(:user, :with_verified_trn) }
-    let(:course_group) { CourseGroup.find_by(name: "leadership") || create(:course_group, name: "leadership") }
+    let(:course_group) { create(:course_group, name: "leadership") }
     let(:course) { create(:course, :senior_leadership, course_group:) }
     let(:lead_provider) { create(:lead_provider) }
     let(:cohort) { create(:cohort, :current) }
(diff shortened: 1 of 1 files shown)


spec/services/applications/accept_spec.rb
```diff
@@ -372,6 +370,17 @@
         let(:new_course) { create(:course, :sl, course_group: new_course_group) }
         let(:new_schedule) { create(:schedule, :npq_leadership_spring, course_group: new_course_group, cohort:) }
```
ethax-ross, line 371: Could use `let!(:new_schedule)` instead of having `before { new_schedule }`

spec/services/applications/accept_spec.rb
```diff
@@ -372,6 +370,17 @@
         let(:new_course) { create(:course, :sl, course_group: new_course_group) }
```
ethax-ross, line 370: I think you'll want to rebase before merging as these course identifiers have changed to remove the abbreviations. We can also drop the find on the course group creation above now as we use `initialize_with`, so `create(:course_group, name: "specialist")` will do a find/create for us.

It took me a minute to figure out what we're testing here; would it make sense to rename `new_course/group` to something like `wrong_course/group`?

ethax-ross (commented):
Looks good, just minor comments

cwrw (approved):

PR #1556, opened 2024-07-25
Ignore empty filters in queries

ethax-ross:
[Jira-3341](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3341)

### Context

Currently, if we add a blank filter to a query, for example `?filter[cohort_start_years]=` it will return no results. Instead, we want to treat the filter as if it wasn't specified, which is how ECF functions.

- Treat blank filters as if they are not specified

Add `FilterIgnorable` concern to centralise the logic around filtering. Treat empty values as ignored as well as the `:ignore` symbol. Retain `nil` as a valid/filterable value.

### DIFF
--- a/app/services/api/concerns/filter_ignorable.rb
+++ b/app/services/api/concerns/filter_ignorable.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module API
+  module Concerns
+    module FilterIgnorable
+      extend ActiveSupport::Concern
+
+      def ignore?(filter:)
+        filter == :ignore || (!filter.nil? && filter.blank?)
+      end
+    end
+  end
+end
--- a/app/services/applications/query.rb
+++ b/app/services/applications/query.rb
@@ -2,6 +2,7 @@ module Applications
   class Query
     include API::Concerns::Orderable
     include Queries::ConditionFormats
+    include API::Concerns::FilterIgnorable
 
     attr_reader :scope, :sort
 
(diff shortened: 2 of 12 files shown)


leandroalemao (commented):
Quick question please:

Trying to do some calls to 
`localhost:3000/api/v3/npq-applications?filter[cohort]=""`
`localhost:3000/api/v3/npq-applications?filter[cohort]=" "`
And no results are returned.
Not sure if I understand well but shouldn't we ignore those filters as well?

ethax-ross:
> Quick question please:
> 
> Trying to do some calls to `localhost:3000/api/v3/npq-applications?filter[cohort]=""` `localhost:3000/api/v3/npq-applications?filter[cohort]=" "` And no results are returned. Not sure if I understand well but shouldn't we ignore those filters as well?

@leandroalemao so here you're passing in a filter values of `""` and `" "`, what we're trying to ignore is an empty filter, so for example:

```
http://0.0.0.0:3000/api/v3/npq-applications?filter[cohort]=
http://0.0.0.0:3000/api/v3/npq-applications?filter[cohort]=   &filter[blah]=somethingelse
```

leandroalemao (approved):
Tks @ethax-ross !!

cwrw (approved):

PR #1558, opened 2024-07-26
[CPDLP-3349] Add accepted at date field to applications

leandroalemao:
### Context

Ticket: [CPDLP-3349](https://dfedigital.atlassian.net/browse/CPDLP-3349)

In ECF on the NPQ participant endpoints we surface the created at of npq application but that doesn’t really capture the same concept as ECF which is where we create a participant profile on acceptance. We have agreed to add a new timestamp to the application to capture this value.

- Add a new accepted at date datetime field to application
- Populate this value on acceptance
- Surface this value in the serialiser instead of the NPQ application value for npq participants endpoints and any csv endpoints (enrolments) 
- Amend any filters that read from the created at for participants npq endpoint, as well as ordering

[CPDLP-3349]: https://dfedigital.atlassian.net/browse/CPDLP-3349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/api/application_serializer.rb
+++ b/app/serializers/api/application_serializer.rb
@@ -32,7 +32,12 @@ module API
       field(:itt_provider) { |a| a.itt_provider&.legal_name }
       field(:lead_mentor)
       field(:funded_place)
-      field(:created_at)
+      field(:created_at) do |a|
+        [
+          a.accepted_at,
+          a.created_at,
+        ].compact.max
+      end
       field(:updated_at) do |a|
         [
           a.user.updated_at,
--- a/app/serializers/api/application_serializer.rb
+++ b/app/serializers/api/application_serializer.rb
@@ -32,12 +32,7 @@ module API
       field(:itt_provider) { |a| a.itt_provider&.legal_name }
       field(:lead_mentor)
       field(:funded_place)
-      field(:created_at) do |a|
-        [
-          a.accepted_at,
-          a.created_at,
-        ].compact.max
-      end
+      field(:created_at)
       field(:updated_at) do |a|
         [
           a.user.updated_at,
--- a/app/serializers/api/applications_csv_serializer.rb
+++ b/app/serializers/api/applications_csv_serializer.rb
@@ -75,7 +75,7 @@ module API
         application.works_in_school,
         application.employer_name,
         application.employment_role,
-        created_at(application),
+        application.created_at.rfc3339,
         updated_at(application),
         application.cohort&.start_year&.to_s,
         application.ineligible_for_funding_reason,
@@ -96,13 +96,6 @@ module API
       [
         application.user.updated_at,
         application.updated_at,
-      ].compact.max.rfc3339
-    end
-
-    def created_at(application)
-      [
-        application.accepted_at,
-        application.created_at,
       ].compact.max.rfc3339
     end
   end
--- a/app/serializers/api/applications_csv_serializer.rb
+++ b/app/serializers/api/applications_csv_serializer.rb
@@ -75,7 +75,7 @@ module API
         application.works_in_school,
         application.employer_name,
         application.employment_role,
-        application.created_at.rfc3339,
+        created_at(application),
         updated_at(application),
         application.cohort&.start_year&.to_s,
         application.ineligible_for_funding_reason,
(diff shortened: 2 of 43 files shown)


spec/serializers/api/applications_csv_serializer_spec.rb
```diff
@@ -76,5 +80,22 @@
...
+    describe "created_at serialization" do
+      let(:parsed_created_at_attribute) { Time.zone.parse(first_row[:created_at]) }
+
+      it { expect(parsed_created_at_attribute).to be_within(1.second).of(first_application.created_at) }
+
+      context "when the application has been accepted" do
+        before do
+          ActiveRecord::Base.no_touching do
+            first_application.update!(created_at: 5.days.ago)
+            first_application.update!(accepted_at: 1.day.ago)
```
mooktakim, line 93: We should combine these

app/serializers/api/application_serializer.rb
```diff
@@ -32,7 +32,12 @@ class AttributesSerializer < Blueprinter::Base
       field(:itt_provider) { |a| a.itt_provider&.legal_name }
       field(:lead_mentor)
       field(:funded_place)
-      field(:created_at)
+      field(:created_at) do |a|
+        [
+          a.accepted_at,
```
ethax-ross, line 37: Is it worth sticking this logic in `Application` as something like `latest_created_or_accepted_at` so we're not duplicating it in the serializers and we can add dedicated tests?

spec/services/applications/accept_spec.rb
```diff
@@ -102,6 +102,7 @@
       it "rejects other_application" do
         service.accept
         expect(application.reload.lead_provider_approval_status).to eql("accepted")
+        expect(application.reload.accepted_at).not_to be_nil
```
ethax-ross, line 105: Would it be better to check for the date rather than not `nil` here? so `be_within(1.minute).of(Time.zone.now)`

app/serializers/api/application_serializer.rb
cwrw: I think in the application serialiser we should stick to the created at, only in the participant serialisers we change this value?

app/serializers/api/applications_csv_serializer.rb
```diff
@@ -96,5 +96,12 @@ def updated_at(application)
         application.updated_at,
       ].compact.max.rfc3339
     end
+
+    def created_at(application)
```
cwrw, line 100: same here regarding keeping this the same, its only on participant serialisers that we need to change this?

app/serializers/api/participant_serializer.rb
```diff
@@ -71,7 +71,7 @@ class AttributesSerializer < Blueprinter::Base
               targeted_delivery_funding_eligibility: application.targeted_delivery_funding_eligibility,
               withdrawal: withdrawal(application:, lead_provider: options[:lead_provider]),
               deferral: deferral(application:, lead_provider: options[:lead_provider]),
-              created_at: application.created_at.rfc3339,
+              created_at: [application.accepted_at, application.created_at].compact.max.rfc3339,
```
cwrw, line 74: could we only use accepted at here, as this should always we populated when an application is accepted, and we should never fallback to the `created_at`?

app/services/applications/query.rb
```diff
@@ -78,7 +78,7 @@ def where_participant_ids_in(participant_ids)
     end
 
     def order_by
-      sort_order(sort:, model: Application, default: { created_at: :asc })
+      sort_order(sort:, model: Application, default: Arel.sql("COALESCE(applications.accepted_at, applications.created_at) ASC"))
```
cwrw, line 81: same here around this being in the participant query instead and this should remain the same

app/services/participants/query.rb
```diff
@@ -52,7 +52,7 @@ def where_from_participant_id_is(from_participant_id)
     end
 
     def order_by
-      sort_order(sort:, model: User, default: { created_at: :asc })
+      sort_order(sort:, model: User, default: Arel.sql("COALESCE(applications.accepted_at, users.created_at) ASC"))
```
mooktakim, line 55: I don't think we need to use `COALESCE` here as we're only looking at accepted applications.
The field `accepted_at` should be set, so we could just order by `{ accepted_at: :asc }`

mooktakim (approved):

cwrw (approved):

ethax-ross (approved):

PR #1563, opened 2024-07-30
Limit response to single errors on certain actions

ethax-ross:
[Jira-3348](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3348)

### Context

In a number of scenarios we return multiple error messages whereas in ECF we would only return a single error at a time.

We want to update NPQ reg to be more consistent with ECF; only returning single errors on occasion.

- Only return one declaration error at a time when voiding

Currently we can return multiple declaration errors when voiding a declaration.

Only return a single declaration error at a time.

- Limit multiple errors on participant actions

Currently, if there is an error on the `participant_id` attribute and the `course_identifier` attribute we would return both errors. Instead, we only want to return the `participant_id` error if there are multiple to match the ECF behaviour.

Minor update to error matcher so we can also check if an error exists (or doesn't exist) for a given attribute.

## Guidance to review

We are removing the `inclusion` validation from the participant actions as the error message is the same as the `invalid` course error which is performed in the custom `CourseForParticipant` validator. The validator also already has the logic in to  omit the error if other errors exist. The validator also ensures the application -> course matches, so we can remove that check from the `Action` class `#application` method.

### DIFF
--- a/app/services/declarations/create.rb
+++ b/app/services/declarations/create.rb
@@ -15,7 +15,7 @@ module Declarations
     validates :lead_provider, presence: true
     validates :participant_id, presence: true
     validates :participant, participant_presence: true, participant_not_withdrawn: true
-    validates :course_identifier, inclusion: { in: Course::IDENTIFIERS }, allow_blank: false, course_for_participant: true
+    validates :course_identifier, course_for_participant: true
     validates :declaration_date, declaration_date: true
     validates :declaration_date, presence: true
     validates :declaration_type, presence: true
--- a/app/services/declarations/void.rb
+++ b/app/services/declarations/void.rb
@@ -59,6 +59,8 @@ module Declarations
     end
 
     def declaration_is_paid
+      return if errors[:declaration].any?
+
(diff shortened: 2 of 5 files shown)


app/services/participants/action.rb
```diff
@@ -28,7 +28,7 @@ def application
         &.applications
         &.accepted
         &.includes(:course)
-        &.find_by(lead_provider:, course: { identifier: course_identifier })
```
cwrw, line 31: not sure about this one, testing locally and getting strange results, let's chat tomorrow

cwrw: tl;dr: this will select any application for a user, which will bring up other validations unnecessarily. For instance a participant has a withdrawn application for course 1 and another active application for course 2. I am trying to defer the second application and pass in the wrong course identifier, it selects the application for course 1 and returns an error for it being withdrawn, which we don't want

cwrw (approved):

leandroalemao (approved):

PR #1564, opened 2024-07-30
[CPDLP-3345] Remove unused request params before calling participant action services

leandroalemao:
### Context

Ticket: [CPDLP-3345](https://dfedigital.atlassian.net/browse/CPDLP-3345)

Currently if we pass in an extra field in the request json for instance reason to the resume endpoint we get a 500 error.

We should ignore unused fields in JSON requests rather than return 500 errors in the action endpoints.

[CPDLP-3345]: https://dfedigital.atlassian.net/browse/CPDLP-3345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v3/participants_controller.rb
+++ b/app/controllers/api/v3/participants_controller.rb
@@ -16,7 +16,7 @@ module API
       end
 
       def resume
-        service = ::Participants::Resume.new(participant_action_params.slice(*::Participants::Resume.attribute_names))
+        service = ::Participants::Resume.new_filtering_attributes(participant_action_params)
 
         if service.resume
           render json: to_json(service.participant)
@@ -26,7 +26,7 @@ module API
       end
 
       def defer
-        service = ::Participants::Defer.new(participant_action_params.slice(*::Participants::Defer.attribute_names))
+        service = ::Participants::Defer.new_filtering_attributes(participant_action_params)
 
         if service.defer
           render json: to_json(service.participant)
@@ -36,7 +36,7 @@ module API
       end
 
       def withdraw
-        service = ::Participants::Withdraw.new(participant_action_params.slice(*::Participants::Withdraw.attribute_names))
+        service = ::Participants::Withdraw.new_filtering_attributes(participant_action_params)
 
         if service.withdraw
           render json: to_json(service.participant)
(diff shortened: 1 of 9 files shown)


spec/requests/api/v1/participants_spec.rb
```diff
@@ -44,7 +44,9 @@ def path(id = nil)
       resume_api_v1_participant_path(ecf_id: id)
     end
 
-    it_behaves_like "an API update endpoint"
+    it_behaves_like "an API update endpoint" do
+      let(:params) { { data: { attributes: attributes.merge(reason: "test") } } }
```
ethax-ross, line 48: Not sure we need these `params` defined? they should be pulled from the `attributes` further up. Is it worth modifying the `an API update endpoint` shared context to have an explicit test that sends some random data along with the standard attributes to ensure an error isn't returned?

app/controllers/api/v3/participants_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
       end
 
       def resume
-        service = ::Participants::Resume.new(participant_action_params)
+        service = ::Participants::Resume.new(participant_action_params.slice(*::Participants::Resume.attribute_names))
```
ethax-ross, line 19: Instead of slicing the attributes for all actions/controllers, would it be cleaner to modify the core `Action` model with a convenience method that we can also easily test? something like:

```
def self.new_filtering_attributes(attributes = {})
  filtered_attributes = attributes.slice(*attribute_names)
  new(filtered_attributes)
end
```

cwrw (approved):

ethax-ross (commented):
Looks good, I think we still want a test in the shared context though to ensure these controllers are correctly filtering attributes?

spec/support/shared_examples/participant_action_support.rb
```diff
@@ -33,6 +33,16 @@
       it { is_expected.to have_error(:participant_id, :invalid_participant, "Your update cannot be made as the '#/participant_id' is not recognised. Check participant details and try again.") }
     end
   end
+
+  describe ".new_filtering_attributes" do
+    context "when extra params are sent to the action service" do
+      subject { described_class.new_filtering_attributes(lead_provider:, participant:, course_identifier:, reason: "test", schedule_identifier: "any") }
```
ethax-ross, line 39: I think this changed from `participant` to `participant_id` recently

```suggestion
      subject { described_class.new_filtering_attributes(lead_provider:, participant_id:, course_identifier:, reason: "test", schedule_identifier: "any") }
```

ethax-ross (approved):
Looks good, just need to fix a spec failure after a rebase I think

PR #1568, opened 2024-08-01
[CPDLP-3293] NPQExit - Configure logit

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3293

* Added `append_info_to_payload` to `Api::BaseController` to track metadata for logit.

### Guidance

* [Search for the review app log in the last 10 min](https://kibana-uk1.logit.io/s/b2876053-0173-44fc-983d-99567292ba31/app/data-explorer/discover#?_a=(discover:(columns:!(app.host,url.original,url.path),isDirty:!t,sort:!()),metadata:(indexPattern:'8ac115c0-aac1-11e8-88ea-0383c11b333a',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-10m,to:now))&_q=(filters:!(),query:(language:kuery,query:'*npq-registration-review-1568-*')))

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -4,33 +4,12 @@ module API
 
     include API::TokenAuthenticatable
     include ActionController::MimeResponds
+    include API::LoggerPayload
 
     rescue_from ActionController::UnpermittedParameters, with: :unpermitted_parameter_response
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ArgumentError, with: :bad_request_response
     rescue_from API::Errors::FilterValidationError, with: :filter_validation_error_response
-
-    def append_info_to_payload(payload)
-      super
-      payload[:current_user_class] = current_lead_provider&.class&.name
-      payload[:current_user_id] = current_lead_provider&.id
-      payload[:current_user_name] = current_lead_provider&.name
-
-      payload[:query_params] = request.query_parameters.to_json
-      payload[:request_headers] = request.env.slice(
-        "HTTP_HOST",
-        "HTTP_ACCEPT",
-        "HTTP_VERSION",
-        "QUERY_STRING",
-        "HTTP_USER_AGENT",
-      ).transform_values(&:to_s).to_json
-      payload[:request_body] = request.raw_post
-
-      payload[:response_headers] = response.headers.transform_values(&:to_s).to_json
-      if response.status != 200
-        payload[:response_body] = response.body
-      end
-    end
 
   private
 
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -16,12 +16,49 @@ module API
       payload[:current_user_id] = current_lead_provider&.id
       payload[:current_user_name] = current_lead_provider&.name
 
-      # payload[:request_headers] = request.env.transform_values(&:to_s).to_json
+      payload[:request_headers] = request.env.slice(
+        "CONTENT_TYPE",
+        "GATEWAY_INTERFACE",
+        "HTTP_ACCEPT",
+        "HTTP_ACCEPT_ENCODING",
+        "HTTP_AUTHORIZATION",
+        "HTTP_CONNECTION",
+        "HTTP_HOST",
+        "HTTP_USER_AGENT",
+        "HTTP_VERSION",
+        "ORIGINAL_FULLPATH",
+        "ORIGINAL_SCRIPT_NAME",
+        "PATH_INFO",
+        "QUERY_STRING",
+        "REMOTE_ADDR",
+        "REQUEST_METHOD",
+        "REQUEST_PATH",
+        "REQUEST_URI",
+        "ROUTES_9080_SCRIPT_NAME",
+        "SCRIPT_NAME",
+        "SERVER_NAME",
+        "SERVER_PORT",
+        "SERVER_PROTOCOL",
+        "SERVER_SOFTWARE",
+        "puma.request_body_wait",
+        "rack.after_reply",
+        "rack.attack.called",
+        "rack.attack.throttle_data",
+        "rack.hijack?",
+        "rack.multiprocess",
+        "rack.multithread",
+        "rack.request.query_hash",
+        "rack.request.query_string",
+        "rack.run_once",
+        "rack.url_scheme",
+        "rack.version",
+      )
+      payload[:request_body] = request.raw_post
 
-      # if response.status != 200
-      #   payload[:response_body] = response.body
-      # end
-      # payload[:response_headers] = response.headers.to_json
+      payload[:response_headers] = response.headers
+      if response.status != 200
+        payload[:response_body] = response.body
+      end
     end
 
   private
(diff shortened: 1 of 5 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -9,6 +9,57 @@ class BaseController < ActionController::API
...
+      payload[:current_user_class] = current_lead_provider&.class&.name
+      payload[:current_user_id] = current_lead_provider&.id
+      payload[:current_user_name] = current_lead_provider&.name
+
+      payload[:request_headers] = request.env.slice(
+        "CONTENT_TYPE",
+        "GATEWAY_INTERFACE",
+        "HTTP_ACCEPT",
+        "HTTP_ACCEPT_ENCODING",
+        "HTTP_AUTHORIZATION",
```
leandroalemao, line 23: Maybe we better off not including this as I saw: `"HTTP_AUTHORIZATION":"Bearer ambition-token"` in my tests on Logit?

app/controllers/api/base_controller.rb
```diff
@@ -9,6 +9,57 @@ class BaseController < ActionController::API
...
+        "rack.url_scheme",
+        "rack.version",
+      ).transform_values(&:to_s).to_json
+      payload[:request_body] = request.raw_post
+
+      payload[:response_headers] = response.headers.transform_values(&:to_s).to_json
+      if response.status != 200
+        payload[:response_body] = response.body
+      end
+    end
```
leandroalemao, line 61: Seems other projects adding `request_query_params` to the payload as well.. maybe it's a good idea:

https://github.com/DFE-Digital/apply-for-teacher-training/blob/99dc097ac0c47678865ff2aa1f14dd3e2701ad81/app/controllers/api_docs/api_docs_controller.rb#L8

app/controllers/api/base_controller.rb
```diff
@@ -9,6 +9,57 @@ class BaseController < ActionController::API
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ArgumentError, with: :bad_request_response
 
+    def append_info_to_payload(payload)
+      super
+      payload[:current_user_class] = current_lead_provider&.class&.name
+      payload[:current_user_id] = current_lead_provider&.id
+      payload[:current_user_name] = current_lead_provider&.name
+
+      payload[:request_headers] = request.env.slice(
```
leandroalemao, line 18: Why do we need this list of headers? Wouldn't be better to add all headers regardless or just remove some like `HTTP_AUTHORIZATION`?

leandroalemao (commented):
Hi @mooktakim .. I've tested and added some comments.

Noticed on Logit, for every single request I do to the review app, there're 2 entries in the logs. not sure why?

image

app/controllers/api/base_controller.rb
mooktakim: I'll remove it

app/controllers/api/base_controller.rb
mooktakim: I'll remove auth

mooktakim:
> Hi @mooktakim .. I've tested and added some comments.
> 
> Noticed on Logit, for every single request I do to the review app, there're 2 entries in the logs. not sure why?
> 
> image

I think there's two entries because of one from nginx and another from rails.

app/controllers/api/base_controller.rb
leandroalemao: Think the query_params are mixed into the `params_json` if a request_body is sent.. just wondering if make sense to have a specific field in the payload only for the `query_params`.

image

config/environments/review.rb
```diff
@@ -1,7 +1,7 @@
 require Rails.root.join("config/environments/production")
 
 Rails.application.configure do
-  config.log_level = :debug
+  config.log_level = :info
```
leandroalemao, line 4: Just noticed we have 2 lines for `config.log_level` on `config/environments/production.rb`

leandroalemao (approved):

app/controllers/api/base_controller.rb
```diff
@@ -10,6 +10,56 @@ class BaseController < ActionController::API
     rescue_from ArgumentError, with: :bad_request_response
     rescue_from API::Errors::FilterValidationError, with: :filter_validation_error_response
 
+    def append_info_to_payload(payload)
```
cwrw, line 13: is this something we can move into a concern, and could we add a test for this as well please?

app/controllers/api/base_controller.rb
```diff
@@ -10,6 +10,56 @@ class BaseController < ActionController::API
...
+        "rack.attack.called",
+        "rack.attack.throttle_data",
+        "rack.hijack?",
+        "rack.multiprocess",
+        "rack.multithread",
+        "rack.request.query_hash",
+        "rack.request.query_string",
+        "rack.run_once",
+        "rack.url_scheme",
+        "rack.version",
```
cwrw, line 53: discussed in person let's select the following:
```
{"HTTP_HOST"=>"xxx",
   "HTTP_ACCEPT"=>"application/json",
   "HTTP_VERSION"=>"HTTP/1.1",
   "QUERY_STRING"=>"page%5Bpage%5D=35&page%5Bper_page%5D=100",
   "HTTP_USER_AGENT"=>"GuzzleHttp/6.5.5 curl/7.83.1 PHP/7.4.33"},
   ```

File: app/controllers/api/base_controller.rb (hunk 2 above)
Comment (mooktakim) [reply app/controllers/api/base_controller.rb:61]: Ok i've added `query_params`

File: config/environments/review.rb (hunk 4 above)
Comment (mooktakim) [reply config/environments/review.rb:4]: Good catch @leandroalemao !
Very strange.
I have changed it to what we have for ECF app.

File: app/controllers/api/base_controller.rb (hunk 5 above)
Comment (mooktakim) [reply app/controllers/api/base_controller.rb]: Moved to concern and added test

File: config/environments/production.rb [hunk 7]
```diff
@@ -74,15 +74,10 @@
...
 
-  # Logging
-  config.log_level = :info
-
   if ENV["RAILS_LOG_TO_STDOUT"].present?
-    config.rails_semantic_logger.add_file_appender = false
-
     $stdout.sync = true
-
-    config.semantic_logger.add_appender(io: $stdout, level: Rails.application.config.log_level, formatter: :json)
```
cwrw, line 85: can we keep this, can see it's not defaulting to `Rails.application.config.log_level` unless I'm wrong?

cwrw: I can see it defaults to `:trace`

app/controllers/api/base_controller.rb
```diff
@@ -4,34 +4,13 @@ class BaseController < ActionController::API
 
     include API::TokenAuthenticatable
     include ActionController::MimeResponds
+    include API::LoggerPayload
```
cwrw, line 7: might need to use prepend

cwrw (approved):

cwrw (approved):

PR #1576, opened 2024-08-08
Parallelise test suite in CI

ethax-ross:
[Jira-3249](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3249)

### Context

We want to parallelise the test suite in CI to make running our tests faster.

- Add Knapsack gem

This gem will let us generate the time it takes to run each spec file so that the test runner can more efficiently divide the tests between workers.

- Parallelise test suite in CI

Generate the knapsack spec report and update the GitHub CI to run a matrix for the test suite.

### Guidance for review

The spec runtime has been reduced from ~17m to ~5m.


peteryates:
How much difference does it make? [I unparallelised it just a few months ago](https://github.com/DFE-Digital/npq-registration/pull/1216) - I guess all the new functionality's take its toll?

cwrw:
@peteryates looks like it's halved the time :D from 14 min aprox to 7 minutes, the tests themselves are now 5 min but all the other things add more time

Gemfile
```diff
@@ -61,6 +61,7 @@ group :development, :test do
   gem "capybara"
   gem "capybara-screenshot"
   gem "dotenv-rails"
+  gem "knapsack"
```
cwrw, line 64: 🧠

knapsack_rspec_report.json
```diff
@@ -0,0 +1,285 @@
+{
+  "spec/lib/questionnaires/senco_start_date_spec.rb": 0.1587239997461438,
```
cwrw, line 2: is this something we need to commit to the repo 🤔 as it's generated in the CI flow?

ethax-ross: Yeah we need to commit it, it needs it ahead of running the test suite so it can split it evenly by runtime per worker. It'll work when we add new specs, but just gradually become outdated/less efficient (but still more efficient than without) - you just need to manually re-generate it every now and again.

cwrw: ahh I see 👍

cwrw (approved):

ethax-ross:
> How much difference does it make? [I unparallelised it just a few months ago](https://github.com/DFE-Digital/npq-registration/pull/1216) - I guess all the new functionality take its toll?

@peteryates that parallelised it across cores but this will split it across github action workers. It cuts the runtime by about 2/3rds

ethax-ross:
Closing and re-opening as a new PR as the CI appears stuck

PR #1584, opened 2024-08-12
[CPDNPQ-2004] TRN import with missing leading zeros

TBenedyk:
https://dfedigital.atlassian.net/browse/CPDNPQ-2004

Excel strips leading zeros from TRNs. This PR adds them back if they are missing.

TRNs must be 7 digits. If the importer receives a validated TRN that is less then it will assume excel has stripped them. This PR therefore add 0's until the TRN is 7 digits.

### DIFF
--- a/app/services/importers/manual_validation.rb
+++ b/app/services/importers/manual_validation.rb
@@ -16,9 +16,10 @@ class Importers::ManualValidation
       Rails.logger.debug "no application found for #{row['application_ecf_id']}" if application.nil?
       next if application.nil?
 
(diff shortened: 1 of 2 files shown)


app/services/importers/manual_validation.rb
```diff
@@ -33,4 +34,8 @@ def check_headers
   def rows
     @rows ||= CSV.read(path_to_csv, headers: true)
   end
+
+  def add_leading_zero(trn)
```
slawosz, line 38: I would rather have a method `trn` that do `rjust` but this is fine too :)

slawosz (approved):

PR #1585, opened 2024-08-12
[CPDLP-3391] Only surface accepted applications in enrolments

leandroalemao:
### Context

Ticket: [CPDLP-3391](https://dfedigital.atlassian.net/browse/CPDLP-3391)

On the participant serialiser in npq enrolments we should only surface accepted applications rather than all applications

Amend the participant serialiser to show a list of only accepted applications for that provider.

[CPDLP-3391]: https://dfedigital.atlassian.net/browse/CPDLP-3391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/serializers/api/participant_serializer_spec.rb
+++ b/spec/serializers/api/participant_serializer_spec.rb
@@ -98,6 +98,15 @@ RSpec.describe API::ParticipantSerializer, type: :serializer do
           }.stringify_keys,
         ])
       end
+
+      context "when there're multiple application with different lead provider approval states" do
+        before { create(:application, lead_provider:, user: participant) }
+
+        it "serializes only accepted `npq_enrolments`" do
+          expect(attributes["npq_enrolments"].size).to eq(1)
+          expect(attributes["npq_enrolments"][0]["npq_application_id"]).to eq(application.ecf_id)
+        end
+      end
     end
 
     context "when serializing the `v3` view" do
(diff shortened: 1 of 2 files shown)


app/models/application.rb
```diff
@@ -30,7 +30,7 @@ class Application < ApplicationRecord
   scope :unsynced, -> { where(ecf_id: nil) }
   scope :expired_applications, -> { where(lead_provider_approval_status: "rejected").where("created_at < ?", cut_off_date_for_expired_applications) }
   scope :active_applications, -> { where.not(id: expired_applications) }
-  scope :accepted, -> { where(lead_provider_approval_status: "accepted") }
+  scope :accepted, -> { where(lead_provider_approval_status: "accepted").where.not(accepted_at: nil) }
```
cwrw, line 33: I think we should not look into `accepted_at` and ignore it as the determinate is the lead provider approval status, if we get to a state it's not correct its a data issue which hopefully we won't have in production. Let me know you're thoughts :)

spec/serializers/api/participant_serializer_spec.rb
```diff
@@ -98,6 +98,14 @@
           }.stringify_keys,
         ])
       end
+
+      context "when application is not accepted" do
+        let(:application) { create(:application, lead_provider:) }
```
cwrw, line 103: could this test instead check for the applicant having multiple applications for the same provider, but not accepted? same below

cwrw (approved):

PR #1593, opened 2024-08-13
[CPDLP-3252] All model enums should include `_suffix: true`

leandroalemao:
### Context

Ticket: [CPDLP-3252](https://dfedigital.atlassian.net/browse/CPDLP-3252)

We added a lot of enums to our models in the new NPQ world. We agreed it would be better to add suffixes to ensure we don’t mix up fields in lookups

Add suffixes to all field enums and amend any look up methods/specs.

[CPDLP-3252]: https://dfedigital.atlassian.net/browse/CPDLP-3252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -33,15 +33,15 @@ module ApplicationHelper
   end
 
   def accepted?(application)
-    application.lead_provider_approval_status&.capitalize == "Accepted"
+    application.accepted_lead_provider_approval_status?
   end
 
   def pending?(application)
-    application.lead_provider_approval_status&.capitalize == "Pending"
+    application.pending_lead_provider_approval_status?
   end
 
   def rejected?(application)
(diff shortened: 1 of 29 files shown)


peteryates:
Is this a good idea? It makes finding things _really hard_.

So hard in fact that @Lockyy [had to leave breadcrumbs](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/participant_profile.rb#L21)! 😅

peteryates:
Fair. I'm still not the biggest fan because someone who doesn't know the ins and outs of Rails might end up on a wild goose chase looking for `def withdrawn_training_status!` which won't exist.

Still, not the end of the world.

mooktakim (approved):

PR #1599, opened 2024-08-14
[CPDLP-3257] Add in DB `kind_of_nursery` enum in application

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3257

We currently set an enum for `kind_of_nursery` in the application model but don’t have it in the db

Set `kind_of_nursery` as an enum in the db.


ethax-ross (approved):
Looks good and works locally. Values match what's in the ECF database (which is also a `string` column type 😬):

```
[nil,
 "private_nursery",
 "local_authority_maintained_nursery",
 "preschool_class_as_part_of_school",
 "another_early_years_setting",
 "childminder"]
```

PR #1600, opened 2024-08-14
Add state machine to `Declaration` model

ethax-ross:
### Context

We want to add a state machine to the `Declaration` model to be consistent with `Statement` and `StatementItem`.

- Add state machine to `Declaration` model

Add state machine, defining transitions.

- Use state machine where possible

We can clean up/remove some of our methods to use the convenience methods provided by the state machine library.

### Guidance to review

I opted not to add a state machine to the `ParticipantOutcome#state` as we don't appear to transition between states with outcomes, instead creating a new outcome of a different state.

I'm not sure how much value this is actually adding; I think we need to go more 'all in' on the state machine and look to consolidate some of the constants/state logic spread throughout the various service classes.

### DIFF
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -46,6 +46,36 @@ class Declaration < ApplicationRecord
     clawed_back: "clawed_back",
   }, _suffix: true
 
+  state_machine :state, initial: :submitted do
+    event :mark_eligible do
+      transition [:submitted] => :eligible
+    end
+
+    event :mark_payable do
+      transition [:eligible] => :payable
+    end
+
+    event :mark_paid do
+      transition [:payable] => :paid
+    end
+
+    event :mark_ineligible do
+      transition %i[submitted eligible payable paid] => :ineligible
+    end
+
+    event :mark_awaiting_clawback do
+      transition %i[paid] => :awaiting_clawback
+    end
+
+    event :mark_clawed_back do
+      transition %i[awaiting_clawback] => :clawed_back
+    end
+
+    event :mark_voided do
+      transition %i[submitted eligible payable ineligible] => :voided
+    end
+  end
+
   enum declaration_type: {
     started: "started",
     "retained-1": "retained-1",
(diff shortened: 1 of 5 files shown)


docs/state_machines.md
```diff
@@ -36,3 +36,35 @@ stateDiagram-v2
...
+    eligible --> ineligible: mark_ineligible
+    payable --> ineligible: mark_ineligible
+    paid --> ineligible: mark_ineligible
+    paid --> awaiting_clawback: mark_awaiting_clawback
+    awaiting_clawback --> clawed_back: mark_clawed_back
+    submitted --> voided: mark_voided
+    eligible --> voided: mark_voided
+    payable --> voided: mark_voided
+    ineligible --> voided: mark_voided
+```
```
peteryates, line 70: Nice 👌🏽

leandroalemao (approved):
Looks good.

There's one more state in ECF app but it's not being used anywhere...happy to just not add it then.

```
event :mark_submitted do
   transition [:eligible] => :submitted
end
```

On the state machine usage throughout the various service classes, think it's gonna increase when we add/migrate finance/statements features to NPQ.

docs/state_machines.md
cwrw: 🧜‍♀️

cwrw (approved):
thanks @ethax-ross, I agree that this doesn't feel like it's adding much value, especially around the services and where the validation and business logic lives. I think it's ok to see this as more of a documentation exercise that we can expand on later, and to keep the current validations in the services for now and something we will keep in mind while we evolve the service :)

app/models/declaration.rb
```diff
@@ -46,6 +45,36 @@ class Declaration < ApplicationRecord
...
+    event :mark_payable do
+      transition [:eligible] => :payable
+    end
+
+    event :mark_paid do
+      transition [:payable] => :paid
+    end
+
+    event :mark_ineligible do
+      transition %i[submitted eligible payable paid] => :ineligible
```
cwrw, line 62: thinking about this, from ECF it looks like we only actually mark it as ineligible if  it's submitted?

cwrw (approved):

PR #1601, opened 2024-08-14
Change `Schedule#allowed_declaration_types` to `declaration_types` enum

ethax-ross:
[Jira-3353](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3353)

### Context

We have duplicate enum types for declaration type; we want to use the `declaration_type` on `Schedule` and remove the `schedule_declaration_types` enum.

- Change `Schedule#allowed_declaration_types` to `declaration_types` enum

### Guidance to review

This turned out to be non-trivial due to Postgres not supporting changing the type of an enum/array column. We have to remove the default and cast all the values as part of the migration.

As we're only running in sandbox this is probably fine, but it has the potential for data issues during this migration (the default will be removed while the casting takes place, so could introduce bad data). We can write a more robust/complex migration to mitigate this, but its probably not worth it for sandbox and the chances are it will run without issue.


leandroalemao (approved):
Tested locally.. looking good. Tks 👍

mooktakim (approved):

cwrw (approved):

PR #1620, opened 2024-08-19
[CPDLP-3434] Fix flakey tests linked to rack attack

leandroalemao:
### Context

Ticket: [CPDLP-3434](https://dfedigital.atlassian.net/browse/CPDLP-3434)

Intermittently we get 429 failures in one of the parallel builds in Rspec in NPQ reg see https://github.com/DFE-Digital/npq-registration/actions/runs/10453888479/job/28945276236. After it’s rebuilt it passes ok

- Fix this issue to avoid flakiness and delays in the build
- Rack attack was being turned on in a test and depending upon whether it runs before the `spec/requests` specs, we then were getting rack attack blocking requests specs afterwards.
- Add missing `spec/support/rack_attack.rb` to turn rack attack off by default in tests and cause we still wanna optionally turn it on in [spec/support/shared_examples/rate_limiting_support.rb](https://github.com/DFE-Digital/npq-registration/blob/main/spec/support/shared_examples/rate_limiting_support.rb#L3) in order to test specifics like [spec/requests/rate_limiting_spec.rb](https://github.com/DFE-Digital/npq-registration/blob/main/spec/requests/rate_limiting_spec.rb ).

[CPDLP-3434]: https://dfedigital.atlassian.net/browse/CPDLP-3434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/support/helpers/admin_login.rb
+++ b/spec/support/helpers/admin_login.rb
@@ -9,8 +9,6 @@ module Helpers
     end
 
     def sign_in_as(admin_account)
-      Rack::Attack.enabled = false
-
       visit("/admin")
       expect(page).to have_current_path(sign_in_path)
 
@@ -24,8 +22,6 @@ module Helpers
       page.click_button "Sign in"
 
       expect(page).to have_current_path("/admin")
-
-      Rack::Attack.enabled = true
     end
   end
 end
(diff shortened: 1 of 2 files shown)


cwrw (approved):
🙌

PR #1621, opened 2024-08-20
CPDNPQ-2028 updated RTTA flow

TBenedyk:
https://dfedigital.atlassian.net/browse/CPDNPQ-2028

This update changes the flow for the "Return to teaching adviser" question. Users are only shown it if they answer "Other" for the "Work setting" question.

It also adds a new response for "Work setting" of "Another setting". This answer replaces the old "Other" response and uses the same logic. The new "Other" response means something else. This means that we need to migrate the data for previous "Other" responses to "Another setting".

It has been confirmed that this data is not transmitted externally so no external migrations are required.


db/migrate/20240823120904_update_work_setting_other_answers.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateWorkSettingOtherAnswers < ActiveRecord::Migration[7.1]
+  def up
+    Application.where(work_setting: "other").update(work_setting: "another_setting")
+  end
+
+  def down
+    Application.where(work_setting: "another_setting").update(work_setting: "other")
+  end
+end
```
TBenedyk, line 9: Only around ~1500 applications in prod will be updated so no major concerns with this taking too long.

slawosz (approved):

PR #1624, opened 2024-08-21
[CPDLP-3442] Add new table and model for outcomes api request

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3442

Following [CPDLP-3396](https://dfedigital.atlassian.net/browse/CPDLP-3396) we came up with a plan to implement sending outcomes to TRA in NPQ reg app.

- Add new table and model for participant outcomes api request;
- Check indexes/FK we need to also create (per ECF app);

[CPDLP-3396]: https://dfedigital.atlassian.net/browse/CPDLP-3396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/participant_outcome_api_request.rb
+++ b/app/models/participant_outcome_api_request.rb
@@ -1,10 +1,5 @@
 class ParticipantOutcomeAPIRequest < ApplicationRecord
   belongs_to :participant_outcome
 
-  validates :ecf_id,
-            presence: { message: "Enter an ECF ID" },
-            uniqueness: {
-              case_sensitive: false,
-              message: "ECF ID must be unique",
-            }
+  validates :ecf_id, presence: true, uniqueness: { case_sensitive: false }
 end
--- a/app/models/participant_outcome_api_request.rb
+++ b/app/models/participant_outcome_api_request.rb
@@ -1,0 +1,10 @@
+class ParticipantOutcomeAPIRequest < ApplicationRecord
+  belongs_to :participant_outcome
+
+  validates :ecf_id,
+            presence: { message: "Enter an ECF ID" },
+            uniqueness: {
+              case_sensitive: false,
+              message: "ECF ID must be unique",
+            }
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_08_14_120004) do
+ActiveRecord::Schema[7.1].define(version: 2024_08_20_153944) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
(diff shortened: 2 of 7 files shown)


db/schema.rb
```diff
@@ -288,6 +288,19 @@
     t.index ["user_id"], name: "index_participant_id_changes_on_user_id"
   end
 
+  create_table "participant_outcome_api_requests", force: :cascade do |t|
```
cwrw, line 291: could we add an `ecf_id` here as well?

db/migrate/20240821124341_add_foreign_key_to_participant_outcome_api_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddForeignKeyToParticipantOutcomeAPIRequests < ActiveRecord::Migration[7.1]
+  def change
+    add_foreign_key :participant_outcome_api_requests, :participant_outcomes, primary_key: :ecf_id
```
cwrw, line 3: Im wondering if we need to link them via the `ecf_id` as opposed to the normal integer ids on the models 🤔  we should have ecf ids on both but link them via the normal ids maybe?

mooktakim: @leandroalemao agree with abeer here, i don't think we should use `ecf_id` for primary key, its inconsistent with the rest of the app. 
We should see `ecf_id` as a separation leftover and maybe in the future will get removed.

leandroalemao: yeah agree tks guys!! Addressed in [6626487](https://github.com/DFE-Digital/npq-registration/pull/1624/commits/6626487038291e71b4dd5ce9460c6c559f662cb5) 👍 😄

app/models/participant_outcome_api_request.rb
```diff
@@ -0,0 +1,10 @@
+class ParticipantOutcomeAPIRequest < ApplicationRecord
+  belongs_to :participant_outcome
+
+  validates :ecf_id,
+            presence: { message: "Enter an ECF ID" },
```
ethax-ross, line 5: Can we move these to the translations file

ethax-ross (approved):
Looks good, just a minor comment on the validation message

app/models/participant_outcome_api_request.rb
leandroalemao: Hi @ethax-ross 👍 tks.. ended up removing the custom message as don't think it's needed in this case.

ethax-ross (approved):

mooktakim (approved):

PR #1627, opened 2024-08-22
Remove hard-coded coverage paths from sonar-project.properties

ethax-ross:
These change on each run and are specified in the action when we call the sonar-scanner CLI. The path to the combined coverage report was also wrong as I didn't realise simplecov puts it in the default `coverage` directory instead of the directory where the `collate` is ran.


peteryates (approved):
Looks good

leandroalemao (approved):

PR #1642, opened 2024-08-27
Migrate `ParticipantIdChange` records from ECF

ethax-ross:
[Jira-3437](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3437)

### Context

We want to migrate participant ID changes from ECF to NPQ reg. We have to do this in a separate migrator (to the users migrator) as we need to migrate the users first (the `ParticipantIdChange` model does not store the `ecf_id`, instead using the NPQ reg ids).

- Add ParticipantIdChange migrator

### DIFF
--- a/app/services/migration/migrators/participant_id_change.rb
+++ b/app/services/migration/migrators/participant_id_change.rb
@@ -1,7 +1,7 @@
 module Migration::Migrators
   class ParticipantIdChange < Base
     class << self
-      def model_count
+      def record_count
         ecf_participant_id_changes.count
       end
 
@@ -29,7 +29,9 @@ module Migration::Migrators
           user_id: user_id_by_ecf_id[ecf_participant_id_change.user_id],
           from_participant_id: user_id_by_ecf_id[ecf_participant_id_change.from_participant_id],
           to_participant_id: user_id_by_ecf_id[ecf_participant_id_change.to_participant_id],
-        )
+        ).tap do |participant_id_change|
+          participant_id_change.update!(created_at: ecf_participant_id_change.created_at)
+        end
       end
     end
   end
--- a/app/services/migration/migrators/participant_id_change.rb
+++ b/app/services/migration/migrators/participant_id_change.rb
@@ -1,0 +1,36 @@
+module Migration::Migrators
+  class ParticipantIdChange < Base
+    class << self
+      def model_count
+        ecf_participant_id_changes.count
+      end
+
+      def model
+        :participant_id_change
+      end
+
+      def dependencies
+        %i[user]
+      end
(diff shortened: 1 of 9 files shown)


app/services/migration/migrators/participant_id_change.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      def dependencies
+        %i[user]
+      end
+
+      def ecf_participant_id_changes
+        Migration::Ecf::ParticipantIdChange
+          .joins(:user)
+          .where(user: { id: User.ecf_users.pluck(:id) })
```
cwrw, line 19: do we need to filter this 🤔 I think maybe it's worth importing all of them incase we are hiding errors?

ethax-ross: I figured it would be best to filter them here as we're only importing a subset of the users so presumed we'd only want the corresponding participant id changes? otherwise we will try and bring across id changes for users we didn't import and it'll fail as a foreign key won't exist

cwrw: ah yes, we don't want all ECF users 👍

cwrw (approved):

app/services/migration/migrators/participant_id_change.rb
```diff
@@ -0,0 +1,36 @@
...
+    def call
+      migrate(self.class.ecf_participant_id_changes) do |ecf_participant_id_change|
+        ecf_ids = ecf_participant_id_change.attributes.slice("user_id", "from_participant_id", "to_participant_id").values
+        user_id_by_ecf_id = ::User.where(ecf_id: ecf_ids).select(:id, :ecf_id).index_by(&:ecf_id).transform_values(&:id)
+
+        ::ParticipantIdChange.find_or_create_by!(
+          user_id: user_id_by_ecf_id[ecf_participant_id_change.user_id],
+          from_participant_id: user_id_by_ecf_id[ecf_participant_id_change.from_participant_id],
+          to_participant_id: user_id_by_ecf_id[ecf_participant_id_change.to_participant_id],
+        )
```
leandroalemao, line 32: @ethax-ross Should we migrate `created_at` field as well as it's being used in the serializer?

ethax-ross: Yeah good catch I'll add that 👍

leandroalemao (approved):
@ethax-ross Looking good tks

PR #1648, opened 2024-08-29
Migrate `Schedules` from ECF

ethax-ross:
### Context

We want to migrate NPQ schedules from ECF into NPQ reg.

- Add `Schedule` migrator

Add the `Schedule` migrator to bring across all NPQ schedules from ECF. As part of this change we also now create `CourseGroup` during the `Course` migration (we can't do this in its own migrator as it doesn't fir the pattern of bringing across records from ECF).

A `Schedule` must match on cohort, name, identifier and course group (derived from the ECF schedule type). We set the applies from/to based on the milestone dates. All milestone dates for ECF schedules must be the same or a failure is recorded.

### DIFF
--- a/app/services/migration/migrators/course.rb
+++ b/app/services/migration/migrators/course.rb
@@ -1,5 +1,12 @@
 module Migration::Migrators
   class Course < Base
+    COURSE_GROUP_NAMES = %w[
+      leadership
+      specialist
+      support
+      ehco
+    ].freeze
+
     class << self
       def record_count
         ecf_npq_courses.count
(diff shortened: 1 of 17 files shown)


app/services/migration/migrators/course.rb
```diff
@@ -15,13 +22,31 @@ def ecf_npq_courses
...
+    def find_course_group!(course, name)
+      course_definition = ::Courses::DEFINITIONS.find { |d| d[:name] == name }
+
+      if course_definition.nil?
+        course.errors.add(:base, "A course group could not be found for the course #{name}")
+        raise ActiveRecord::RecordInvalid, course
+      end
+
+      ::CourseGroup.find_by(name: course_definition[:course_group_name])
+    end
```
mooktakim, line 50: Its better to use `ecf_npq_course.identifier` instead of `name` for matching.

app/services/migration/migrators/schedule.rb
```diff
@@ -0,0 +1,74 @@
...
+          cohort_id: cohorts_by_start_year[ecf_schedule.cohort.start_year].id,
+          name: ecf_schedule.name,
+          identifier: ecf_schedule.schedule_identifier,
+          course_group:,
+        ).tap do |schedule|
+          ecf_milestone = ecf_schedule.milestones.first
+          schedule.update!(
+            applies_from: ecf_milestone.start_date,
+            applies_to: ecf_milestone.payment_date,
+          )
```
mooktakim, line 42: We need to set `schedule.allowed_declaration_types`.
```ruby
schedule.update!(
  applies_from: ecf_milestone.start_date,
  applies_to: ecf_milestone.payment_date,
  allowed_declaration_types: ecf_schedule.milestones.pluck(:declaration_type),
)
```

app/services/migration/migrators/schedule.rb
```diff
@@ -0,0 +1,74 @@
...
+        course_group = course_groups_by_identifier(ecf_schedule.type)
+
+        unless course_group
+          ecf_schedule.errors.add(:base, "Course group not found for schedule")
+          raise ActiveRecord::RecordInvalid, ecf_schedule
+        end
+
+        ::Schedule.find_or_initialize_by(
+          cohort_id: cohorts_by_start_year[ecf_schedule.cohort.start_year].id,
+          name: ecf_schedule.name,
```
mooktakim, line 34: `name` should move down to the `.tap`, we want to find/initialize with `identifier`, `cohort` and `cohort_group`

mooktakim (changes requested):
couple of changes, but otherwise pretty good

mooktakim (approved):

PR #1649, opened 2024-08-30
Migrate `Declarations` from ECF

ethax-ross:
[Jira-3436](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3436)

### Context

We want to bring across declarations from ECF into NPQ reg; matching on `ecf_id` and taking the ECF values as the source of truth.

- Migrate Declarations from ECF

Migrate `Declaration` records from ECF. We run a second migrator to set the `superseded_by_id` on the migrated declarations, as we need to be able to look up their NPQ reg ids for this.

We match declarations on `ecf_id` and update the attributes with the values in ECF. We may need to optimise the application lookup depending on runtime.


app/services/migration/migrators/declaration.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def dependencies
+        %i[cohort application lead_provider course user]
+      end
+    end
+
+    def call
+      migrate(self.class.ecf_declarations) do |ecf_declaration|
+        declaration = ::Declaration.find_or_initialize_by(ecf_id: ecf_declaration.id)
+        latest_declaration_state = ecf_declaration.declaration_states.order(created_at: :desc).first
```
cwrw, line 24: Usually we surface the reason for ineligible states in the serialiser:
```
declaration.declaration_states.detect(&:ineligible?)&.state_reason
```
and we only set this reason in the record declaration if it's ineligible:
```
participant_declaration.make_ineligible!(reason: :duplicate)
```
Maybe we check to only grab the ineligible state reasons similar to the first query and setting it on the declaration?

app/services/migration/migrators/declaration.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        declaration.update!(
+          created_at: ecf_declaration.created_at,
```
cwrw, line 27: :100

spec/factories/migration/ecf/declaration_states.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :ecf_migration_declaration_state, class: "Migration::Ecf::DeclarationState" do
+    state_reason { Declaration.state_reasons.keys.sample }
```
cwrw, line 5: could we change this to set the value on ineligible?

spec/services/migration/migrators/declaration_superseded_by_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+
+    def create_npq_resource(ecf_resource)
+      lead_provider = create(:lead_provider, ecf_id: ecf_resource.cpd_lead_provider.npq_lead_provider.id)
+      cohort = create(:cohort, start_year: ecf_resource.cohort.start_year)
+      course = create(:course, identifier: ecf_resource.course_identifier)
+      user = create(:user, ecf_id: ecf_resource.user.id)
+      application = create(:application, :accepted, course:, user:)
+      create(:declaration, ecf_id: ecf_resource.id, cohort:, lead_provider:, application:)
```
cwrw, line 18: could we make this ineligible so it matches superseded declarations incase in ECF?

ethax-ross: not sure I follow you here? there's actually a bunch of setup that's not needed in this migrator I've now removed (as the migrator simply matches on the `ecf_id` and updates the `superseeded_by` it doesn't care about cohort/user/application)

cwrw: ah sorry meant that a superseded declaration is always ineligible, if we show it in the test it helps indicate why those are ineligible thats all :)

cwrw (approved):
looks good past the spec comment :)

PR #1650, opened 2024-08-30
Add `ParticipantOutcome` migrator

ethax-ross:
[Jira-3438](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3438)

### Context

We want to bring across `ParticipantOutcome` records from ECF to NPQ reg.

- Add ParticipantOutcome migrator

Migrate `ParticipantOutcome` records from ECF to NPQ reg.


app/models/migration/ecf/participant_outcome/npq.rb
```diff
@@ -0,0 +1,7 @@
+module Migration::Ecf
+  class ParticipantOutcome::Npq < BaseRecord
```
cwrw, line 2: we should probably bring across the participant outcome api requests as well, will add a ticket

leandroalemao (approved):
Looking good tks @ethax-ross

cwrw (approved):

PR #1651, opened 2024-08-30
[CPDLP-3448] Create NPQ Contract models

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3448

* Created `ContractTemplate` model
* Created `Contract` model
* Created seed script


app/models/contract.rb
```diff
@@ -0,0 +1,7 @@
+class Contract < ApplicationRecord
+  belongs_to :statement
+  belongs_to :course
+  belongs_to :contract_template
+
+  validates :course_id, uniqueness: { scope: :statement_id, message: "Can only have one contract for statement and course" }
```
ethax-ross, line 6: Can we stick this in the translations file? ditto for other validation messages

db/seeds/data/contracts/2021.csv
```diff
@@ -0,0 +1,69 @@
+provider_name,cohort_year,course_identifier,recruitment_target,per_participant,service_fee_installments,special_course,monthly_service_fee,funding_cap
```
ethax-ross, line 1: Do we need to use CSVs for the seed data? is it a case of the values are specific/constant for each lead provider/cohort?

app/models/contract.rb
peteryates: The problem with using locales is it's much more difficult to spot when one's missing and we end up with `"is invalid"` errors. We're just going to avoid locales altogether in ECF2.

Harder here though as there's already a mix of approaches.

ethax-ross: True, but there should be tests ensuring the messages are present/correct; having them in the model just adds clutter and IMO they're easier to digest/keep DRY in the locales. But that's just my opinion 🙂

mooktakim: Moved to the translation file 👍

db/seeds/data/contracts/2021.csv
mooktakim: There are differences between cohort and courses. Using the existing seed csv from ecf will make sure we don't have unusual scenarios.

cwrw: hmmm I was hoping to avoid csvs as much as we can in NPQ, especially with the different cohorts. We were able to get away from this with the factories and seed data for statements as well as the statement populaters by making it more generic, is this something we can't replicate here? happy to have a chat

db/migrate/20240830095100_create_contract_templates.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.integer :service_fee_percentage, null: false, default: 40
+      t.decimal :per_participant, null: false
+      t.integer :number_of_payment_periods
+      t.integer :output_payment_percentage, null: false, default: 60
+      t.decimal :monthly_service_fee, default: 0.0
+      t.decimal :targeted_delivery_funding_per_participant, default: 100.0
+      t.boolean :special_course, null: false, default: false
+
+      # NPQContract.id from ecf app
+      t.uuid :ecf_id, index: true
```
cwrw, line 15: could we add the `, default: "gen_random_uuid()", null: false, index: { unique: true }` here?

mooktakim: @cwrw 
I was thinking as we don't show this ID anywhere, I'm not sure if we should generate it, that way we only have ID's for ecf migrated contracts.

mooktakim: Added generate to `ecf_id`

db/seeds/base/add_contracts.rb
```diff
@@ -0,0 +1,21 @@
+LeadProvider.find_each do |lead_provider|
+  Cohort.find_each do |cohort|
+    Course.find_each do |course|
+      if cohort.start_year < 2023 && (course.senco? || course.npqlpm?)
```
cwrw, line 4: A bit of a nit, maths is a special cours eand starts 2023, but isn't special in 2024. Senco starts 2024 and isn't a special course :)

mooktakim: Updated 👍

leandroalemao (approved):
Looks good 👍 tks

cwrw (approved):
thanks @mooktakim 🙌

ethax-ross (approved):
Looks good 👍

PR #1655, opened 2024-09-02
Add `StatementLineItem` to `Statement` migrator

ethax-ross:
[Jira-3440](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3440)

### Context

As part of the `Statement` migrator we also want to bring across `StatementLineItem` records from ECF. We can match these on the `declaration_id` and `state` for each statement.

- Migrate statement line items from ECF

### DIFF
--- a/app/models/migration/ecf/finance/statement.rb
+++ b/app/models/migration/ecf/finance/statement.rb
@@ -5,6 +5,7 @@ module Migration::Ecf::Finance
     belongs_to :cohort
     belongs_to :cpd_lead_provider
     has_one :npq_lead_provider, through: :cpd_lead_provider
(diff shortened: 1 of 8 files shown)


cwrw (commented):
looks good, I wonder if we should have a separate migrator vs in the same statement migrator, mostly because if there are errors I expect them to be separate to the statement 🤔 happy to discuss

cwrw (approved):

mooktakim (approved):

PR #1661, opened 2024-09-03
Fix sonar cloud in CI

ethax-ross:
As of v4.4 of `actions/upload-artifact@v4` it no longer uploads hidden files; this means the `.resultset.json` coverage report is omitted. Add `include-hidden-files: true` to ensure hidden files are picked up.

https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#hidden-files


cwrw (approved):
⭐ !

PR #1662, opened 2024-09-03
Add `ApplicationState` migrator

ethax-ross:
[Jira-3489](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3489)

### Context

We want to bring across the `ParticipantProfileState` records from ECF to NPQ reg, as `ApplicationState` records. The `participant_profile_id` will match up to the `ecf_id` on the `Application` in NPQ reg.

- Add `ApplicationState` migrator

### DIFF
--- a/app/services/migration/migrators/application_state.rb
+++ b/app/services/migration/migrators/application_state.rb
@@ -1,0 +1,38 @@
+module Migration::Migrators
+  class ApplicationState < Base
+    class << self
+      def record_count
+        ecf_participant_profile_states.count
+      end
+
+      def model
+        :application_state
+      end
+
+      def dependencies
+        %i[application lead_provider]
+      end
+
+      def ecf_participant_profile_states
+        Migration::Ecf::ParticipantProfileState
+          .joins(participant_profile: :teacher_profile)
+          .includes(:cpd_lead_provider)
+          .where(teacher_profile: { user_id: User.ecf_users.pluck(:id) })
+      end
+    end
+
+    def call
+      migrate(self.class.ecf_participant_profile_states) do |ecf_participant_profile_state|
+        application_state = ::ApplicationState.find_or_initialize_by(ecf_id: ecf_participant_profile_state.id)
+
+        ecf_lead_provider_id = ecf_participant_profile_state.cpd_lead_provider&.npq_lead_provider&.id
+        application_state.lead_provider = ::LeadProvider.find_by!(ecf_id: ecf_lead_provider_id) if ecf_lead_provider_id
+
+        ecf_application_id = ecf_participant_profile_state.participant_profile_id
+        application_state.application_id = ::Application.select(:id).find_by!(ecf_id: ecf_application_id).id
+
+        application_state.update!(ecf_participant_profile_state.attributes.slice(%w[state reason created_at updated_at]))
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/services/migration/migrators/application_state.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      def dependencies
+        %i[application]
+      end
+
+      def ecf_participant_profile_states
+        Migration::Ecf::ParticipantProfileState
+          .joins(participant_profile: :teacher_profile)
+          .includes(:cpd_lead_provider)
+          .where(teacher_profile: { user_id: User.ecf_users.pluck(:id) })
```
cwrw, line 20: should the ecf users be a scope on the ECF User model, feels a bit odd to read into another migrator to get that query 🤔 happy to chat through

app/services/migration/migrators/application_state.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def call
+      migrate(self.class.ecf_participant_profile_states) do |ecf_participant_profile_state|
+        ecf_application_id = ecf_participant_profile_state.participant_profile_id
+        application_id = ::Application.select(:id).find_by!(ecf_id: ecf_application_id).id
+        application_state = ::ApplicationState.find_or_initialize_by(application_id:)
+
+        ecf_lead_provider_id = ecf_participant_profile_state.cpd_lead_provider&.npq_lead_provider&.id
+        application_state.lead_provider = ::LeadProvider.find_by!(ecf_id: ecf_lead_provider_id) if ecf_lead_provider_id
```
cwrw, line 31: should we also depend on lead provider migrator 🤔 as applications doesn't right now in it's current form due to the diffing out of attributes

app/services/migration/migrators/application_state.rb
```diff
@@ -0,0 +1,37 @@
...
+          .includes(:cpd_lead_provider)
+          .where(teacher_profile: { user_id: User.ecf_users.pluck(:id) })
+      end
+    end
+
+    def call
+      migrate(self.class.ecf_participant_profile_states) do |ecf_participant_profile_state|
+        ecf_application_id = ecf_participant_profile_state.participant_profile_id
+        application_id = ::Application.select(:id).find_by!(ecf_id: ecf_application_id).id
+        application_state = ::ApplicationState.find_or_initialize_by(application_id:)
```
cwrw, line 28: so I believe an application can have multiple states, so we might need to use the ecf id of the profile state to find or initialise by instead

app/services/migration/migrators/application_state.rb
ethax-ross: I almost did this when I implemented it originally, but I reasoned that since the migrators depend on each other anyway it makes sense. Happy to change it to a scope on the `User` if you feel differently though

leandroalemao (approved):
Looks good tks @ethax-ross

app/services/migration/migrators/application_state.rb
cwrw: since this is short lived code, its ok to keep as is for now

db/schema.rb
```diff
@@ -135,6 +137,34 @@
     t.index ["start_year"], name: "index_cohorts_on_start_year", unique: true
   end
 
+  create_table "contract_templates", force: :cascade do |t|
```
cwrw, line 140: not sure those should be committed here?

cwrw (approved):

PR #1663, opened 2024-09-03
Temporarily toggle api on in sandbox for testing

leandroalemao:
### Context

Temporarily toggle api on in sandbox for testing.

### DIFF
--- a/config/environments/sandbox.rb
+++ b/config/environments/sandbox.rb
@@ -1,14 +1,14 @@
 require Rails.root.join("config/environments/production")
 
 Rails.application.configure do
-  config.log_level = :info
+  config.log_level = :debug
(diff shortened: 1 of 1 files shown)


config/environments/sandbox.rb
```diff
@@ -1,14 +1,14 @@
...
-  config.log_level = :info
+  config.log_level = :debug
   config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/healthcheck") } } }
 
   # Enable/disable aspects of the NPQ separation
   config.npq_separation = {
-    admin_portal_enabled: false,
-    api_enabled: false,
+    admin_portal_enabled: true,
+    api_enabled: true,
```
cwrw, line 10: I thought we wanted to test on separation instead as it has all the setup? not keen on turning this on on sandbox as its connected to the ECF sandbox 🤔

PR #1664, opened 2024-09-03
Clear migration failures prior to running

ethax-ross:
[Jira-3485](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3485)

### Context

We are finding that failures are duplicated in the reports, so the same ID for an object that has failed will appear multiple times.

- Clear migration failures prior to running

As we store failures in Redis, when we clear the NPQ reg database and re-run a migration after a failed migration the failure manager will add the failures to those from the last run (as the `DataMigration` models/ids get reset so are the same as the previous run).

To avoid this, we can purge the failures from Redis for a `DataMigration` as part of the migration preparation.

- Fix rounding issues with data migrations

We are rounding down in the `MigrationHelper`, however we were still rounding up in the `DataMigration` model; this ensures both round down correctly and improves the tests.

### Guidance to review

The `FailureManager` is becoming a not so great mix of instance/class methods; it could do with a tidy up but its probably not worth the effort given this code is going to be relatively short-lived.

### DIFF
--- a/app/models/migration/data_migration.rb
+++ b/app/models/migration/data_migration.rb
@@ -16,13 +16,13 @@ module Migration
     def percentage_migrated_successfully
       return 0 unless processed_count&.positive?
 
-      ((processed_count - failure_count) / processed_count.to_f * 100).round
+      ((processed_count - failure_count) / processed_count.to_f * 100).floor
     end
 
     def percentage_migrated
       return 0 unless total_count&.positive?
 
-      (processed_count / total_count.to_f * 100).round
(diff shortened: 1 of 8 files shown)


cwrw (approved):
looks good 🙌 tomorrow I'll try clearing everything from migration and testing again, tried now but seems to be down 😅

PR #1672, opened 2024-09-04
Optimise migrators

ethax-ross:
[Jira-3482](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3482)

### Context

We want to improve the performance of the migrators so that we can perform a full migration more quickly.

- Optimise migrators

Various optimisations to the migrators; mainly pre-loading lookups into memory in bulk once instead of querying multiple times individually.

### DIFF
--- a/app/services/migration/migrators/base.rb
+++ b/app/services/migration/migrators/base.rb
@@ -48,7 +48,7 @@ module Migration::Migrators
       end
 
       def records_per_worker
-        1_000
+        10_000
       end
     end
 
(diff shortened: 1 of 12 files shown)


leandroalemao (approved):
Looking good.. tested locally here! much difference well done @ethax-ross tks

app/services/migration/migrators/base.rb
```diff
@@ -83,8 +83,48 @@ def data_migration
...
+      @application_ids_by_ecf_id ||= ::Application.pluck(:ecf_id, :id).to_h
+    end
+
+    def lead_provider_ids_by_ecf_id
+      @lead_provider_ids_by_ecf_id ||= ::LeadProvider.pluck(:ecf_id, :id).to_h
+    end
+
+    def cohort_ids_by_start_year
+      @cohort_ids_by_start_year ||= ::Cohort.pluck(:start_year, :id).to_h
+    end
```
mooktakim, line 126: Should we use redis cache or memcache for this? that way all the workers will shared the same cache.
Also, I wonder if this is using up all the RAM, redis would resolve that.

ethax-ross: I actually tried this as part of the investigation but our Redis instance is quite small and it ended up being a bottleneck. I think there was also contention around which worker populates the cache (as you get ~20 trying to do it at the same time). I have the changes stashed if we want to revisit it/bump our Redis instance though.

mooktakim (approved):

cwrw (approved):
thanks @ethax-ross excellent work. We'll do another iteration soon and revisit some ideas

PR #1673, opened 2024-09-04
Only check verified TRNs in User migrator

ethax-ross:
[Jira-3487](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3487)

### Context

We only want to flag an issue if an ECF user is found with multiple, verified TRNs that are different on their NPQ applications.

- Only check verified TRNs in User migrator

### DIFF
--- a/app/services/migration/migrators/user.rb
+++ b/app/services/migration/migrators/user.rb
@@ -30,11 +30,11 @@ module Migration::Migrators
       migrate(self.class.ecf_users) do |ecf_user|
         user = ::User.find_or_initialize_by(ecf_id: ecf_user.id)
 
-        unique_trns = ecf_user.npq_applications.pluck(:teacher_reference_number).compact.uniq
-        validate_multiple_trns!(unique_trns, user)
+        trns = unique_validated_trns(ecf_user)
+        validate_multiple_trns!(trns, user)
 
         user.update!(
-          trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : unique_trns.last,
+          trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : trns.last,
           full_name: ecf_user.full_name || user.full_name,
           email: ecf_user.email || user.email,
           uid: ecf_user.get_an_identity_id || user.uid,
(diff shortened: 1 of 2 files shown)


app/services/migration/migrators/user.rb
```diff
@@ -44,8 +44,16 @@ def call
 
   private
 
-    def validate_multiple_trns!(unique_trns, user)
-      return unless unique_trns.size > 1
+    def unique_validated_trns(ecf_user)
+      ecf_user.npq_applications
```
cwrw, line 48: could we include npq applications in the query, so there are no n+1s?

app/services/migration/migrators/user.rb
```diff
@@ -44,8 +44,16 @@ def call
...
+        .where(teacher_reference_number_verified: true)
```
cwrw, line 49: maybe we use select to avoid an extra query here 🤔

app/services/migration/migrators/user.rb
ethax-ross: I don't think that would make any difference here as - as confusing and unintuitive as it is - to be consistent with ECF behaviour `npq_applications` runs a separate query:

https://github.com/DFE-Digital/npq-registration/blob/main/app/models/migration/ecf/user.rb#L8-L10

app/services/migration/migrators/user.rb
ethax-ross: As we `pluck` the query results I don't think `select` would add any benefit?

app/services/migration/migrators/user.rb
cwrw: ah gotcha 👍

app/services/migration/migrators/user.rb
cwrw: so initially I wanted to avoid a query altogether and do it in memory and add a select loop but since it has to be a query we can't get around it I guess

cwrw (approved):

mooktakim (approved):

PR #1674, opened 2024-09-04
[CPDLP-3450] NPQ Contracts migration from ECF to NPQ

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3450

* Created `Migration::Ecf::NpqContract` model
* Created `Migration::Migrators::Contract` migrator
* Created tests and factories.

### DIFF
--- a/app/services/migration/migrators/contract.rb
+++ b/app/services/migration/migrators/contract.rb
@@ -1,0 +1,67 @@
+module Migration::Migrators
+  class Contract < Base
+    class << self
+      def record_count
+        ecf_contracts.count
+      end
+
+      def model
+        :contract
+      end
+
+      def ecf_contracts
+        Migration::Ecf::NpqContract.includes(:cohort, npq_lead_provider: [:cpd_lead_provider])
+      end
+
+      def dependencies
+        %i[cohort lead_provider course statement]
+      end
+    end
+
+    def call
+      migrate(self.class.ecf_contracts) do |ecf_contract|
+        course = courses_by_identifier[ecf_contract.course_identifier]
+
+        ecf_statements(ecf_contract).find_each do |ecf_statement|
+          statement = ::Statement.find_by!(ecf_id: ecf_statement.id)
+
+          contract_template = ::ContractTemplate.find_or_create_by!(ecf_id: ecf_contract.id)
+          contract_template.update!(
+            created_at: ecf_contract.created_at,
+            updated_at: ecf_contract.updated_at,
+
+            service_fee_percentage: ecf_contract.service_fee_percentage,
+            output_payment_percentage: ecf_contract.output_payment_percentage,
+            per_participant: ecf_contract.per_participant,
+            number_of_payment_periods: ecf_contract.number_of_payment_periods,
+            recruitment_target: ecf_contract.recruitment_target,
+            service_fee_installments: ecf_contract.service_fee_installments,
+            targeted_delivery_funding_per_participant: ecf_contract.targeted_delivery_funding_per_participant,
+            monthly_service_fee: ecf_contract.monthly_service_fee,
+            special_course: ecf_contract.special_course,
+          )
+
+          contract = ::Contract.find_or_initialize_by(
+            statement:,
+            course:,
+          )
+          contract.update!(contract_template:)
+        end
+      end
+    end
+
+    def ecf_statements(ecf_contract)
+      Migration::Ecf::Finance::Statement.where(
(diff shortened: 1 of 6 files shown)


app/services/migration/migrators/contract.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    def call
+      migrate(self.class.ecf_contracts) do |ecf_contract|
+        course = courses_by_identifier[ecf_contract.course_identifier]
+
+        ecf_statements(ecf_contract).find_each do |ecf_statement|
+          statement = ::Statement.find_by!(ecf_id: ecf_statement.id)
+
+          contract_template = ::ContractTemplate.find_or_create_by!(ecf_id: ecf_contract.id)
```
cwrw, line 28:
```suggestion
          contract_template = ::ContractTemplate. find_or_initialize_by(ecf_id: ecf_contract.id)
```

spec/services/migration/migrators/contract_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      create(
+        :ecf_migration_npq_contract,
+        npq_lead_provider:,
+        cohort:,
+        version: "0.0.1",
+      )
+    end
+
+    describe "#call" do
+      it "sets the created Contract and ContractTemplate attributes correctly" do
```
cwrw, line 71: might be worth using the shared context here for migrators

app/services/migration/migrators/contract.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    def call
+      migrate(self.class.ecf_contracts) do |ecf_contract|
+        course = courses_by_identifier[ecf_contract.course_identifier]
+
+        ecf_statements(ecf_contract).find_each do |ecf_statement|
+          statement = ::Statement.find_by!(ecf_id: ecf_statement.id)
+
+          contract_template = ::ContractTemplate.find_or_initialize_by(ecf_id: ecf_contract.id)
```
ethax-ross, line 28: As the attributes have the same names could do:


```
    SHARED_ATTRIBUTES = %w[
      created_at
      updated_at
      service_fee_percentage
      output_payment_percentage
      per_participant
      number_of_payment_periods
      recruitment_target
      service_fee_installments
      targeted_delivery_funding_per_participant
      monthly_service_fee
      special_course
    ].freeze

contract_template.update!(ecf_contract.attributes.slice(SHARED_ATTRIBUTES))
```

Or if there are only a few we don't want to migrate across, something like:

```
contract_template.update!(ecf_contract.attributes.excluding(:id))
```

spec/services/migration/migrators/contract_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      )
+      create(
+        :contract,
+        statement:,
+        course:,
+        contract_template:,
+      )
+    end
+
+    def setup_failure_state
```
ethax-ross, line 51: Can we add a comment around why this should be a failure?

spec/services/migration/migrators/contract_spec.rb
ethax-ross: I think it already is?

spec/services/migration/migrators/contract_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        expect(contracts.count).to eq(1)
+
+        contract = statement.contracts.first
+        expect(contract.course.identifier).to eq(ecf_resource1.course_identifier)
+
+        contract_template = contract.contract_template
+        expect(contract_template.ecf_id).to eq(ecf_resource1.id)
+
+        attrs = ecf_resource1.attributes.slice(
+          :service_fee_percentage,
```
ethax-ross, line 95: I think the ECF NPQContract factory and the NPQ ContractTemplate factory both have the same values for each of these anyway, which means the test will pass regardless of if they get set in the migrator 😬 can we change one of the factories to be slightly different so its testing the attribute assignment?

ethax-ross (approved):
Looks good, just one thing I noticed about the factories happen to have the same attribute values in ECF/NPQ; if we make them different the test will be more robust, but other than that I think this is good to go 👍

spec/services/migration/migrators/contract_spec.rb
mooktakim: Makes sense, updating now

ethax-ross (approved):

leandroalemao (commented):
Missed that one but well done @mooktakim tks

PR #1677, opened 2024-09-04
[CPDLP-3488] Update NPQ participant serialiser to show validated trn

leandroalemao:
### Context

Ticket: [CPDLP-3488](https://dfedigital.atlassian.net/browse/CPDLP-3488)

In ECF we surface the TRN in the NPQ participant serialiser from the teacher profile. The teacher profile is only updated if the TRN is validated.

In NPQ update the trn to only read from the teacher_reference_number if it is validated on the application.

[CPDLP-3488]: https://dfedigital.atlassian.net/browse/CPDLP-3488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/api/participant_serializer.rb
+++ b/app/serializers/api/participant_serializer.rb
@@ -25,7 +25,9 @@ module API
           end
         end
 
-        field(:trn, name: :teacher_reference_number)
+        field(:teacher_reference_number) do |object, _options|
+          object.trn if object.trn_verified
+        end
         field(:updated_at) do |object, _options|
           updated_at(object)
         end
@@ -34,7 +36,9 @@ module API
       view :v2 do
         field(:email)
         field(:full_name)
-        field(:trn, name: :teacher_reference_number)
+        field(:teacher_reference_number) do |object, _options|
+          object.trn if object.trn_verified
+        end
         field(:updated_at) do |object, _options|
           updated_at(object)
         end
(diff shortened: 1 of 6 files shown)


spec/serializers/api/participant_serializer_spec.rb
```diff
@@ -55,8 +55,20 @@
...
+        context "when trn is verified" do
+          before { participant.update!(trn_verified: true) }
+
+          it "serializes the `teacher_reference_number`" do
+            expect(attributes["teacher_reference_number"]).to eq(participant.trn)
+          end
+        end
+
+        context "when trn is not verified" do
+          it "serializes the `teacher_reference_number`" do
```
ethax-ross, line 68: `it "returns nil"`?

app/serializers/api/participant_serializer.rb
```diff
@@ -25,7 +25,9 @@ class AttributesSerializer < Blueprinter::Base
           end
         end
 
-        field(:trn, name: :teacher_reference_number)
+        field(:teacher_reference_number) do |object, _options|
```
ethax-ross, line 28: It can be for a future ticket, but we should DRY up this serializer/spec so we're not repeating the same fields for each view, similar to how its done in the declaration serializer https://github.com/DFE-Digital/npq-registration/blob/6aaca99983a36898948db3d4ed264e7f78aa051a/app/serializers/api/declaration_serializer.rb

ethax-ross (approved):

app/serializers/api/participant_serializer.rb
cwrw: will add to my notes 👍

mooktakim (approved):

PR #1683, opened 2024-09-06
Remove `lead_provider_approval_status` from application migrator

ethax-ross:
[Jira-3491](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3491)

### Context

We will be taking the ECF version of this attribute, so we don't want to treat it as a failure if they are different.

- Remove `lead_provider_approval_status` from application migrator


ethax-ross:
Ditching as we're going to update the values of NPQ reg with those in ECF so this will be redundant.

PR #1685, opened 2024-09-06
[CPDLP-3486] Fix User migrator email and other data

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3486

* Updated User migrator to raise error:
  * when there are multiple emails
  * when the email does not match existing users email


app/services/migration/migrators/user.rb
```diff
@@ -58,5 +62,20 @@ def validate_multiple_trns!(trns, user)
...
+
+    def validate_multiple_emails!(unique_emails, user)
+      return unless unique_emails.size > 1
+
+      user.errors.add(:base, "There are multiple different emails from user identities in NPQ applications")
+      raise ActiveRecord::RecordInvalid, user
+    end
+
+    def validate_existing_email_match!(ecf_email, user)
+      return if user.email.blank?
```
ethax-ross, line 74: I don't think we should need this as NPQ users have to have emails? 🤔

app/services/migration/migrators/user.rb
```diff
@@ -33,10 +33,14 @@ def call
         trns = unique_validated_trns(ecf_user)
         validate_multiple_trns!(trns, user)
 
+        unique_emails = ecf_user.npq_applications.map { |app| app.participant_identity.email.to_s.downcase.strip }.compact.uniq
```
ethax-ross, line 36: Worth extracting this to a separate method to be consistent with `unique_validated_trns`?

app/services/migration/migrators/user.rb
mooktakim: This is when a new User is initialized that exists in ECF but doesn't in NPQ with the `ecf_id`, that should not validate.

app/services/migration/migrators/user.rb
```diff
@@ -52,11 +56,30 @@ def unique_validated_trns(ecf_user)
         .uniq
     end
 
+    def unique_emails(ecf_user)
+      ecf_user.npq_applications.map { |app| app.participant_identity.email.to_s.downcase.strip }.compact.uniq
```
cwrw, line 60: could we include participant identities from participant profiles as well, just incase a participant profile has a different identity to an NPQ application?

app/services/migration/migrators/user.rb
```diff
@@ -33,10 +33,14 @@ def call
...
+        emails = unique_emails(ecf_user)
+        validate_multiple_emails!(emails, user)
+        validate_existing_email_match!(emails.first, user)
+
         user.update!(
           trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : trns.last,
           full_name: ecf_user.full_name || user.full_name,
-          email: ecf_user.email || user.email,
+          email: emails.first || user.email,
           uid: ecf_user.get_an_identity_id || user.uid,
```
cwrw, line 44: thinking about DoB/nino/active alert, can we pull those back from ECF, although they don't get updated by admin I wonder if someone went into the console and updated, I think worth overwriting from ECF anyway?

cwrw (commented):
thanks Mook, a few questions. After we get the failures for emails, I'll check with Jake around the impact of updating this in NPQ

app/services/migration/migrators/user.rb
cwrw: could we also pull in `trn_verified` as `teacher_reference_number_verified` from ECF, as that could be updated on the ECF side as well

ethax-ross (approved):
Looks good, has it been tested on the migration env?

cwrw (approved):
Thanks @mooktakim , I asked Jake around email overwriting but we'll followup with the failures and reach out to the dev team as he wasn't sure, so one in later tickets. +1 for testing on migration, I can deploy this once the current run finishes

app/services/migration/migrators/user.rb
```diff
@@ -52,11 +62,36 @@ def unique_validated_trns(ecf_user)
...
     end
 
+    def unique_emails(ecf_user)
+      emails = ecf_user.npq_profiles.map { |pp| pp.participant_identity.email }
+      emails += ecf_user.npq_applications.map { |app| app.participant_identity.email }
+      emails.map { |email| email.to_s.downcase.strip }.compact.uniq
+    end
+
+    def most_recent_updated_npq_application(ecf_user)
+      ecf_user.npq_applications.order(updated_at: :desc).first
```
cwrw, line 72: having a further think about this, maybe it's worth also looking for the user -> profile -> npq application if the npq application is empty here as opposed to what I suggested earlier regarding adding the safe navigation operator, it feels safer 🤔

app/services/migration/migrators/user.rb
```diff
@@ -33,11 +33,21 @@ def call
...
         user.update!(
           trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : trns.last,
           full_name: ecf_user.full_name || user.full_name,
-          email: ecf_user.email || user.email,
+          email: emails.first || user.email,
           uid: ecf_user.get_an_identity_id || user.uid,
+          date_of_birth: npq_application&.date_of_birth || user.date_of_birth,
+          national_insurance_number: npq_application&.nino || user.national_insurance_number,
+          active_alert: npq_application&.active_alert,
+          trn_verified: npq_application&.teacher_reference_number_verified,
```
cwrw, line 50: in the report we are getting `  PG::NotNullViolation: ERROR:  null value in column "trn_verified" of relation "users" violates not-null constraint`, could we fallback to the user value for both the alert and verified values incase?

cwrw (approved):

PR #1693, opened 2024-09-10
Migrate ECF applications to NPQ reg

ethax-ross:
[Jira-3497](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3497)

### Context

We want to bring across the data associated with NPQ applications in ECF to NPQ reg. Previously we were performing an integrity check on the attributes, but after investigation we determined all attributes should be updated from ECF to NPQ reg for applications.

The school, course, lead provider and user should not be different and we have added an integrity check to ensure this.

- Migrate ECF applications to NPQ reg
 
### Guidance to review

The attributes that are identical in each:

```
created_at
updated_at
headteacher_status
eligible_for_funding
funding_choice
works_in_school
employer_name
employment_role
targeted_support_funding_eligibility
teacher_catchment_iso_country_code
works_in_nursery
works_in_childcare
kind_of_nursery
targeted_delivery_funding_eligibility
funding_eligiblity_status_code
employment_type
lead_mentor
primary_establishment
number_of_pupils
tsf_primary_eligibility
tsf_primary_plus_eligibility
lead_provider_approval_status
teacher_catchment
teacher_catchment_country
notes
funded_place
```

The attributes that needed to be remapped:

| ECF    | NPQ |
| -------- | ------- |
| itt_provider  | itt_provider - legal_name   |
| private_childcare_provider_urn | private_childcare_provider - provider_urn    |
| cohort - start_year | cohort - start_year    |
| school_ukprn | ukprn - provider_urn    |
| profile - training_status | private_childcare_provider - provider_urn    |
| profile - schedule_id | schedule_id    |

The relationships that we check are consistent without overwritting:

| ECF    | NPQ |
| -------- | ------- |
| school_urn  | school - urn   |
| npq_course_id | course - ecf_id    |
| npq_lead_provider - id | lead_provider - ecf_id    |
| profile - user_id | user - ecf_id    |

The vast majority of failures are around applications where we can't find the school in NPQ reg, oddly, since there are no school failures I'm not sure why the applications fail:

Screenshot 2024-09-11 at 15 17 41

### DIFF
--- a/app/services/migration/migrators/application.rb
+++ b/app/services/migration/migrators/application.rb
@@ -1,13 +1,14 @@
 module Migration::Migrators
   class Application < Base
-    ATTRIBUTES_TO_COMPARE = %w[
+    ATTRIBUTES = %w[
       headteacher_status
       eligible_for_funding
       funding_choice
-      teacher_catchment
       works_in_school
       employer_name
       employment_role
+      targeted_support_funding_eligibility
+      teacher_catchment_iso_country_code
       works_in_nursery
       works_in_childcare
       kind_of_nursery
@@ -16,9 +17,14 @@ module Migration::Migrators
       employment_type
       lead_mentor
       primary_establishment
+      number_of_pupils
       tsf_primary_eligibility
       tsf_primary_plus_eligibility
       lead_provider_approval_status
+      teacher_catchment
+      teacher_catchment_country
+      notes
+      funded_place
     ].freeze
 
     class << self
@@ -30,8 +36,14 @@ module Migration::Migrators
         :application
       end
 
+      def dependencies
+        %i[cohort lead_provider schedule course user school]
+      end
+
       def ecf_npq_applications
-        Migration::Ecf::NpqApplication.joins(:participant_identity)
+        Migration::Ecf::NpqApplication
+          .joins(:participant_identity)
+          .includes(:cohort, :user, profile: :schedule)
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/services/migration/migrators/application.rb
```diff
@@ -1,13 +1,16 @@
 module Migration::Migrators
   class Application < Base
-    ATTRIBUTES_TO_COMPARE = %w[
+    ATTRIBUTES = %w[
+      created_at
```
cwrw, line 4: I think in terms of application probably not safe bringing along the timestamps as it is an existing model in NPQ 🤔

app/services/migration/migrators/application.rb
```diff
@@ -30,41 +38,86 @@ def model
...
+          hash[cohort_id][identifier] ||= {}
+          hash[cohort_id][identifier][course_group] = id
+        end
+      end
     end
 
-    def compare_attributes_values!(ecf_npq_application, application)
-      return unless ATTRIBUTES_TO_COMPARE.any? { |attribute| ecf_npq_application[attribute] != application[attribute] }
+    def ensure_relationships_are_consistent!(ecf_npq_application, application)
+      if application.school_id != find_school_id!(urn: ecf_npq_application.school_urn)
```
cwrw, line 96: there are many school failures in the report, I think we can avoid some of them by ignoring `nil` from the lookup. Also worth revisiting why others are failing as schools are there 🤔

app/services/migration/migrators/application.rb
ethax-ross: Is it worth keeping the earliest `created_at` and latest `updated_at`?

cwrw: So Application will almost always have an earlier timestamp as it is created before being created in ECF, in terms of updated at, it will all probably be set to the day after migrators, so not too worried about it, as we're asking providers to sync everything after migration

app/services/migration/migrators/application.rb
```diff
@@ -30,41 +36,86 @@ def model
...
       migrate(self.class.ecf_npq_applications) do |ecf_npq_application|
-        application = applications_by_ecf_id[ecf_npq_application.id]
-        raise ActiveRecord::RecordNotFound, "Application not found" unless application
+        application = ::Application.find_by!(ecf_id: ecf_npq_application.id)
+
+        ensure_relationships_are_consistent!(ecf_npq_application, application)
+
+        ecf_schedule = ecf_npq_application.profile&.schedule
+        if ecf_schedule
+          schedule_cohort_id = find_cohort_id!(start_year: ecf_schedule.cohort.start_year)
```
cwrw, line 60: hmmm I think it would have been easier for us to have `ecf_id` on schedule to avoid all this, will add a ticket

cwrw (approved):
thanks for this! 🌟

leandroalemao (approved):

PR #1696, opened 2024-09-11
Reconfigure Docker Compose for development

rwrrll:
### Context

No ticket – based on my first-time setup experience as a new team member. I can keep this as a local patch if there's no appetite to merge!

Re-configures Docker Compose for development so that on any machine with Docker installed you can type `docker compose up -d` and start working, no other dependencies/setup needed.

- By default, starts Postgres and Rails containers, and mounts the project directory as a volume so that you can work from your local filesystem
- Also provides an `ops` container with Azure and Kubernetes tools available

See changes to docs for more info.

#### Changes outside the scope of Docker Compose: 

To get the Rails container working properly:

1. To make browser-driven tests work, the Rails-default `webdrivers` gem is swapped for `selenium-webdriver` specifically — as far as I can see, Capybara only uses Selenium anyway, and all the tests pass, so hopefully not a biggie.

2. `bin/dev` now binds to 0.0.0.0 rather than 127.0.0.1, to accept requests from outside the container. (It looks the default command in Dockerfile already specifies the same.)

#### Changes for Docker Compose users:

If you already use Compose (excluding VS Code devcontainers) then this PR would affect you. But I'm not sure anyone does?

Apart from developers, I searched the project for "compose" and didn't see anything except a Make task that isn't referenced anywhere else (aside: can/should we delete this?), so it looks like Compose also isn't used for ops either? 

In theory even if you currently use Docker Compose, with this commit checked out it should be a case of running `docker compose up -d` and carrying on, but it'd be great to know of any problems.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,3 @@
-web: env RUBY_DEBUG_OPEN=true bin/rails server
+web: env RUBY_DEBUG_OPEN=true bin/rails server -p 3000 -b 0.0.0.0
 js: yarn build --watch
 css: yarn build:css --watch
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,9 +1,6 @@
 require "knapsack"
 Knapsack::Adapters::RSpecAdapter.bind
 
-require "webdrivers"
-Webdrivers.install_dir = [Webdrivers.install_dir, ENV["TEST_ENV_NUMBER"]].compact.join(File::SEPARATOR)
-
 require "simplecov"
 require "simplecov_json_formatter"
 SimpleCov.start "rails"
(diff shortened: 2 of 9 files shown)


Procfile.dev
peteryates, line 1: Worth kicking off workers here too?

peteryates (commented):
Looks good to me! Left a comment re workers but happy to see this.

spec/spec_helper.rb
```diff
@@ -1,9 +1,6 @@
 require "knapsack"
 Knapsack::Adapters::RSpecAdapter.bind
 
-require "webdrivers"
```
cwrw, line 4: I think it should probably be ok, but worth double checking everything works with `bundle exec rake parallel:spec` as expected

cwrw: So just tested running the parallel spec on both branches: on main it passes fine, on you're branch I had to explicitly install `chromedriver` and after that fully passes (which you have in docker compose) so all good

cwrw (approved):
🙌

PR #1697, opened 2024-09-12
Fix bugs in Contract migrator

ethax-ross:
[Jira-3563](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3563)

### Context

The `Contract` migrator has a number of failures caused by a bug in the migration code; we are not correctly copying attributes from the ECF model to the NPQ reg model.

- Optimise lookups in Contract migrator

To be consistent with other migrators and to hopefully speed things up we preload the lookups (course/statement).

- Fix attribute slicing in migrators

In the `Contract` and `ApplicationState` migrators we slice attributes using an array; this results in an empty hash and the tests were passing as we made the same slicing error. We need to use the splat operator here to correctly slice the attributes. In the `Contract` migrator we were also using symbols in the tests instead of strings, which is why the test was also incorrectly passing (slicing by an attribute that doesn't exist omits it).

### Guidance for review

Screenshot 2024-09-12 at 11 13 22

### DIFF
--- a/app/services/migration/migrators/contract.rb
+++ b/app/services/migration/migrators/contract.rb
@@ -34,17 +34,17 @@ module Migration::Migrators
 
     def call
       migrate(self.class.ecf_contracts) do |ecf_contract|
-        course = courses_by_identifier[ecf_contract.course_identifier]
+        course_id = find_course_id!(identifier: ecf_contract.course_identifier)
 
         ecf_statements(ecf_contract).find_each do |ecf_statement|
-          statement = ::Statement.find_by!(ecf_id: ecf_statement.id)
+          statement_id = find_statement_id!(ecf_id: ecf_statement.id)
 
           contract_template = ::ContractTemplate.find_or_initialize_by(ecf_id: ecf_contract.id)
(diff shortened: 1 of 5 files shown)


app/services/migration/migrators/contract.rb
```diff
@@ -28,25 +28,25 @@ def ecf_contracts
...
         ecf_statements(ecf_contract).find_each do |ecf_statement|
-          statement = ::Statement.find_by!(ecf_id: ecf_statement.id)
+          statement_id = find_statement_id!(ecf_id: ecf_statement.id)
 
           contract_template = ::ContractTemplate.find_or_initialize_by(ecf_id: ecf_contract.id)
-          contract_template.update!(ecf_contract.attributes.slice(SHARED_ATTRIBUTES))
+          contract_template.update!(ecf_contract.attributes.slice(*SHARED_ATTRIBUTES))
 
-          contract = ::Contract.find_or_initialize_by(
-            statement:,
```
cwrw, line 46: I think we should keep this the same as it was as we should always have one unique entry for statement/course

PR #1699, opened 2024-09-12
Make `training_status` `active` on accepting an application

ethax-ross:
[Jira-3564](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3564)

### Context

Currently, the default training status of an application is `active` even if they have not been accepted. This is confusing; instead, we want to allow `training_status` to be `nil` (by default) and update it to be `active` on accepting an application.

- Make `training_status` `active` on accepting an application


public/api/docs/v3/swagger.yaml
```diff
@@ -1093,6 +1093,7 @@ components:
         training_status:
           description: Return only records that have this training status
           type: string
+          nullable: true
```
cwrw, line 1096: I don't think as far as the API is concerned this is a nullable field, it will always be populated in the participant profile response so shouldn't be nullable there

spec/factories/applications.rb
```diff
@@ -17,6 +17,7 @@
     itt_provider
     funding_choice { Application.funding_choices.keys.sample }
     lead_mentor { Faker::Boolean.boolean }
+    training_status { :active }
```
cwrw, line 20: maybe we add this only on accepted 🤔

cwrw (approved):

spec/services/applications/accept_spec.rb
```diff
@@ -358,6 +360,7 @@
 
           expect(service.accept).to be_truthy
           expect(service.application.lead_provider_approval_status).to eql("accepted")
+          expect(service.application.training_status).to eql("active")
```
leandroalemao, line 363: Could be as well 

`expect(service.application).to be_active_training_status`

leandroalemao (approved):
All seems good tks @ethax-ross

PR #1705, opened 2024-09-13
[CPDLP-3484] User migrator find or initialise logic to deal with orphaned ecf user, but return error for all the other scenarios

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3484

- Added new `find_or_initialize_user(ecf_user)` method:
  - use `ecf_user.id` and `ecf_user.get_an_identity_id` to find users on NPQ
  - if users don't exist, initialize new user and return
  - if user exists with only `ecf_user.id`, return user
  - if both id's return users they are the same, return user
  - if both id's return users, but they are different, raise error
  - if only `ecf_user.get_an_identity_id` return a user, look into ECF with the `ecf_id`
    - if not found, update NPQ user's `ecf_id`
    - if ECF user exists, check if its an orphaned user (transferred etc)
      - if yes, update NPQ user's `ecf_id`
      - if no, raise an error

### DIFF
--- a/app/services/migration/migrators/user.rb
+++ b/app/services/migration/migrators/user.rb
@@ -28,7 +28,7 @@ module Migration::Migrators
 
     def call
       migrate(self.class.ecf_users) do |ecf_user|
-        user = ::User.find_or_initialize_by(ecf_id: ecf_user.id)
+        user = find_or_initialize_user(ecf_user)
 
         trns = unique_validated_trns(ecf_user)
         validate_multiple_trns!(trns, user)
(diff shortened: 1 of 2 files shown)


app/services/migration/migrators/user.rb
```diff
@@ -54,6 +54,72 @@ def call
...
+          # Orphaned ecf_user, we set `user.ecf_id` to `ecf_user.id`, return user
+          user_with_gai_id.ecf_id = ecf_user.id
+          return user_with_gai_id
+        end
+
+        ecf_user.errors.add(:base, "User found with ecf_user.get_an_identity_id, but its user.ecf_id linked to another ecf_user that is not an orphan")
+        raise ActiveRecord::RecordInvalid, ecf_user
+      end
+
+      ecf_user.errors.add(:base, "End of find_or_initialize_user logic, something weird happened")
```
ethax-ross, line 119: Can we cover this with a test?

spec/services/migration/migrators/user_spec.rb
```diff
@@ -79,6 +93,120 @@ def setup_failure_state
...
+              expect(orphaned_ecf_user.npq_applications).to be_empty
+
+              instance.call
+              expect(user.reload.ecf_id).to eq(ecf_user.id)
+              expect(user.full_name).to eq("New Name")
+            end
+          end
+
+          context "when linked ecf_user is not an orphan" do
+            let(:failure_manager) do
```
ethax-ross, line 194: We mock this already in the shared context, so you can test failures similar to this:

https://github.com/DFE-Digital/npq-registration/blob/main/spec/services/migration/migrators/application_state_spec.rb#L36

app/services/migration/migrators/user.rb
```diff
@@ -54,6 +54,72 @@ def call
 
   private
 
+    def find_or_initialize_user(ecf_user)
+      # Find Users with `ecf_user.id`
+      users_with_ecf_id = ::User.where(ecf_id: ecf_user.id)
```
ethax-ross, line 59: Instead of loading the whole user model into memory here and in the queries below can we just pluck the `id` attributes and compare on that? has this been ran on the migration env yet?

app/services/migration/migrators/user.rb
```diff
@@ -54,6 +54,72 @@ def call
 
   private
 
+    def find_or_initialize_user(ecf_user)
```
ethax-ross, line 57: This is a bit of a monster function; can we extract several methods to make it easier to follow?

app/services/migration/migrators/user.rb
mooktakim: I don't think we can actually get to this point, added this just incase.

spec/services/migration/migrators/user_spec.rb
mooktakim: We need to check for multiple failure messages. The mock doesn't work for that.

app/services/migration/migrators/user.rb
mooktakim: We'll need to return the User if found, no point running the query multiple times.

app/services/migration/migrators/user.rb
mooktakim: I'm not sure, i think the readability is important here as we are dealing with various scenarios, just running through step by step.

cwrw: I think I agree around breaking this up, but I think we're probably going to move this out into the deduper/service anyway in your current ticket Mook, so maybe we break up there rather than here?

app/services/migration/migrators/user.rb
cwrw: yea it didn't get to it in my testing for the general cases, but I did get to it when I broke things in my pr. I think worth adding a test regardless incase you never know :)

cwrw (commented):
looks good, thanks @mooktakim ! Could we deploy to the migration env and check the new run before we merge/approve togeother, happy to chat with @ethax-ross on the method size as well to agree how to go forward

spec/services/migration/migrators/user_spec.rb
ethax-ross: I'm sure it does 😅 

```
expect(failure_manager).to have_received(:record_failure).once.with(ecf_user, "Validation failed: User found with ecf_user.get_an_identity_id, but its user.ecf_id linked to another ecf_user that is not an orphan")
expect(failure_manager).to have_received(:record_failure).once.with(non_orphaned_ecf_user, "Validation failed: Participant identity email from ECF does not match existing user email in NPQ")
```

app/services/migration/migrators/user.rb
ethax-ross: I think it would be much easier to digest/maintain/extend if it was refactored into multiple smaller, well named functions

app/services/migration/migrators/user.rb
mooktakim: We shouldn't get to this point. I've removed the raise to avoid confusion.

spec/services/migration/migrators/user_spec.rb
mooktakim: Yes thats what i needed lol Updated code now.

cwrw (approved):
thanks @mooktakim! we've agreed that we need to break up the method, and will happen in a followup pr for the dedupe service

app/services/migration/migrators/user.rb
cwrw: agreed, this will be addressed shortly hopefully in the followup pr :)

ethax-ross (approved):

leandroalemao (approved):
Looks good tks @mooktakim

PR #1708, opened 2024-09-16
Add `ecf_id` to remaning migration tables

ethax-ross:
[Jira-3569](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3569)

### Context

We want to add the `ecf_id` to all the tables (apart from schools) where we're bringing across data from ECF. This will enable us to simplify the migration logic and speed up some of the lookups.

- Add ecf_id to remaining migration tables

Add the `ecf_id` to the remaining migration tables, apart from `schools` as these already exist in NPQ reg and are only verified/not actually migrated as part of the migration process.

The `ecf_id` fields are nullable since we don't expose them to lead providers and will not need to generate them going forward.

-  Use ecf_id in all migrators

We now have `ecf_id` on cohorts, contracts, participant_id_changes, schedules and statement_items, so we want to match on this attribute going forward as part of the  migration process.

- Optimise cohort lookups with ecf_id

Use `ecf_id` for cohort lookups to avoid including the cohort in the underlying queries.

- Clean up application migrator to use schedule ecf_id

As we now have `ecf_id` on schedule we can perform a simpler lookup.

### Guidance for review

| Main | This branch |
| -- | -- |
| Screenshot 2024-09-17 at 13 20 10 |  Screenshot 2024-09-17 at 10 57 45 |

### DIFF
--- a/app/services/migration/migrators/contract.rb
+++ b/app/services/migration/migrators/contract.rb
@@ -42,12 +42,9 @@ module Migration::Migrators
           contract_template = ::ContractTemplate.find_or_initialize_by(ecf_id: ecf_contract.id)
           contract_template.update!(ecf_contract.attributes.slice(*SHARED_ATTRIBUTES))
 
-          contract = ::Contract.find_or_initialize_by(
-            statement_id:,
-            course_id:,
-          )
+          contract = ::Contract.find_or_initialize_by(ecf_id: ecf_contract.id)
 
-          contract.update!(contract_template:)
+          contract.update!(contract_template:, statement_id:, course_id:)
         end
       end
     end
--- a/app/services/migration/migrators/contract.rb
+++ b/app/services/migration/migrators/contract.rb
@@ -42,9 +42,12 @@ module Migration::Migrators
           contract_template = ::ContractTemplate.find_or_initialize_by(ecf_id: ecf_contract.id)
           contract_template.update!(ecf_contract.attributes.slice(*SHARED_ATTRIBUTES))
 
-          contract = ::Contract.find_or_initialize_by(ecf_id: ecf_contract.id)
+          contract = ::Contract.find_or_initialize_by(
+            statement_id:,
+            course_id:,
+          )
 
-          contract.update!(contract_template:, statement_id:, course_id:)
+          contract.update!(contract_template:)
         end
       end
     end
--- a/db/migrate/20240916124923_add_ecf_id_to_remaining_tables.rb
+++ b/db/migrate/20240916124923_add_ecf_id_to_remaining_tables.rb
@@ -1,0 +1,15 @@
+class AddEcfIdToRemainingTables < ActiveRecord::Migration[7.1]
+  def change
+    add_column :cohorts, :ecf_id, :uuid, null: true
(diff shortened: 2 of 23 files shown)


app/services/migration/migrators/contract.rb
```diff
@@ -42,12 +42,9 @@ def call
           contract_template = ::ContractTemplate.find_or_initialize_by(ecf_id: ecf_contract.id)
           contract_template.update!(ecf_contract.attributes.slice(*SHARED_ATTRIBUTES))
 
-          contract = ::Contract.find_or_initialize_by(
-            statement_id:,
-            course_id:,
-          )
+          contract = ::Contract.find_or_initialize_by(ecf_id: ecf_contract.id)
```
cwrw, line 45: not sure we need this, we are using the `ecf_id` in the contract template which maps to the NPQ contract id in ecf, the join table i.e. contract table shouldn't have an `ecf_id` 🤔 as it doesn't map back, I think we can remove/ignore it?

cwrw (commented):
looking good just the comment. I also noticed extra failures on declaration/application state after this change, just want to confirm  if they are related might be worth confirming they aren't :)

db/migrate/20240916124923_add_ecf_id_to_remaining_tables.rb
```diff
@@ -0,0 +1,15 @@
+class AddEcfIdToRemainingTables < ActiveRecord::Migration[7.1]
+  def change
+    add_column :cohorts, :ecf_id, :uuid, null: true
```
leandroalemao, line 3: Do we need to also add `default: "gen_random_uuid()"` for new records going forward or will this be used only for the migration work?

leandroalemao (commented):
Hi @ethax-ross .. changes are looking good.. only one question please.. tks

db/migrate/20240916124923_add_ecf_id_to_remaining_tables.rb
ethax-ross: As we don't expose these models via the API we don't need to give them an `ecf_id` going forward; if they have the `ecf_id` set it indicates they were created as part of the migration (those with `nil` `ecf_id`s will indicate they were added after)

app/services/migration/migrators/contract.rb
ethax-ross: yeah that sounds sensible I'll remove it 👍

db/migrate/20240916124923_add_ecf_id_to_remaining_tables.rb
leandroalemao: Well yeah make sense..  those models I don't think we'll expose `ecf_id` in the future either.. tks @ethax-ross 👍

leandroalemao (approved):

cwrw (approved):

ethax-ross:
I think I hadn't refreshed the ECF database - I re-ran it today and declarations/application state aligns with main now, although there are a couple of other differences now but I think its just down to changing data:

Screenshot 2024-09-19 at 15 58 50

PR #1714, opened 2024-09-17
Add disabled flag to `IttProvider` and `PrivateChildcareProvider`

ethax-ross:
[Jira-3573](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3573)

### Context

We want to include `IttProvider` and `PrivateChildCareProvider` instances from ECF that are not currently in NPQ reg. In order to bring them across as part of the migration process and not change the current behaviour in NPQ reg, we will flag any new providers as disabled as part of the migration.

- Add disabled_at to provider models and exclude by default

A `default_scope` has been added to ensure that existing behaviour will continue to only show the providers currently in NPQ reg (and not disabled).

- Migrate IttProvider and PrivateChildcareProvider

We want to bring across any IttProvider or PrivateChildcareProvider records that exist in ECF (as part of an NPQApplication) but are not present in NPQ reg. When we bring them across we want to flag them as disabled so that they don't appear in other areas of the NPQ reg application.

Providers are matched case insensitively. We can also fallback on the `operating_name` of the `IttProvider` if set.

- Move disable logic to concern

To avoid duplicating the logic, add `Disableable` concern we can pull into `IttProvider` and `PrivateChildcareProvider`.

### Guidance for review

TODO: run on migration / add screenshot


cwrw (approved):

mooktakim (approved):

db/schema.rb
```diff
@@ -282,6 +282,7 @@
     t.boolean "approved"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.datetime "disabled_at"
```
leandroalemao, line 285: Maybe we could add an index to `disabled_at` as well as it'll be in all queries in the end?

ethax-ross: Its a good shout, but I'm not sure its worth it given the sizes of these tables; the overhead of the index probably outweighs the benefit 🤷‍♂️

leandroalemao: Yeah agree.. just seen.. the tables size are near to "zero".. index won't make any difference. Tks

leandroalemao (approved):

PR #1719, opened 2024-09-17
[CPDLP-3576] Data cleanup - Update validation for emails in User migrator

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3576

* Updated `Migration::Migrators::User` to use the most recent NPQ Application participant identity's email for npq user.


spec/services/migration/migrators/user_spec.rb
```diff
@@ -37,21 +37,22 @@ def setup_failure_state
...
-        create(:ecf_migration_npq_application, teacher_reference_number: ecf_user.npq_applications.first.teacher_reference_number, participant_identity:)
-
-        instance.call
-        expect(failure_manager).to have_received(:record_failure).with(ecf_user, /There are multiple different emails from user identities in NPQ applications/)
-      end
-
-      it "records a failure when existing email in npq does not match ecf email" do
-        ecf_user = create(:ecf_migration_user, :npq, email: "email-1@example.com")
-        create(:user, ecf_id: ecf_user.id, email: "email-2@example.com")
+      it "sets User email from recent updated npq application participant identity" do
```
ethax-ross, line 40: It looks like we have two very similar tests in here:

```
sets User email from recent updated npq application participant identity
sets User attributes from recent updated npq application
```

Is it worth combining them?

app/services/migration/migrators/user.rb
```diff
@@ -33,16 +33,13 @@ def call
         trns = unique_validated_trns(ecf_user)
         validate_multiple_trns!(trns, user)
 
-        emails = unique_emails(ecf_user)
-        validate_multiple_emails!(emails, user)
-        validate_existing_email_match!(emails.first, user)
-
         npq_application = most_recent_updated_npq_application(ecf_user)
```
cwrw, line 36: I wonder if this should be by created at rather than updated at, as we might have an older application updated today, but a newer one created which has the latest data 🤔 what do you think?

mooktakim: Ok, i've updated to `created_at`

spec/services/migration/migrators/user_spec.rb
mooktakim: Combined tests 👍

app/services/migration/migrators/user.rb
cwrw: thanks could we update the method name as well :)

spec/services/migration/migrators/user_spec.rb
ethax-ross: It looks like both are still there? I think we can ditch `sets User attributes from recent updated npq application` now as your other test covers it? although it might be worth having a separate test that makes sure we set the email to the `user.email` if there's no application -> participant identity -> email

app/services/migration/migrators/user.rb
mooktakim: Updated

spec/services/migration/migrators/user_spec.rb
mooktakim: I don't think its a real scenario. Participant Identity is required for NPQ Application and email is required for Participant Identity. We won't run the migration on the user if it doesn't have applications.

ethax-ross: Is it worth ditching the `|| user.email` fallback in that case if its redundant?

cwrw (approved):
will run today on migration to confirm 🙌

spec/services/migration/migrators/user_spec.rb
mooktakim: removed

PR #1727, opened 2024-09-18
[CPDLP-3575] Update user from/to participant id changes table

leandroalemao:
### Context

Ticket: [CPDLP-3575](https://dfedigital.atlassian.net/browse/CPDLP-3575)

After reviewing errors in migrators, we found out there're some missing users in NPQ as they are orphans. We’ve agreed to change the table to UUIDs and only populate those values without bringing across the users.

Update user from/to in NPQ for Participant id changes table.

[CPDLP-3575]: https://dfedigital.atlassian.net/browse/CPDLP-3575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/participant_id_change.rb
+++ b/app/models/participant_id_change.rb
@@ -2,8 +2,8 @@
 
 class ParticipantIdChange < ApplicationRecord
   belongs_to :user
-  belongs_to :from_participant, class_name: "User"
-  belongs_to :to_participant, class_name: "User"
+  belongs_to :from_participant, class_name: "User", primary_key: "ecf_id"
+  belongs_to :to_participant, class_name: "User", primary_key: "ecf_id"
 
   validates :user, :from_participant, :to_participant, presence: true
 end
--- a/app/models/participant_id_change.rb
+++ b/app/models/participant_id_change.rb
@@ -2,8 +2,6 @@
 
 class ParticipantIdChange < ApplicationRecord
   belongs_to :user
-  belongs_to :from_participant, class_name: "User", primary_key: "ecf_id"
-  belongs_to :to_participant, class_name: "User", primary_key: "ecf_id"
 
-  validates :user, :from_participant, :to_participant, presence: true
+  validates :user, :from_participant_id, :to_participant_id, presence: true
 end
--- a/db/migrate/20240917142926_remove_foreign_keys_from_participant_id_changes.rb
+++ b/db/migrate/20240917142926_remove_foreign_keys_from_participant_id_changes.rb
@@ -1,0 +1,8 @@
+class RemoveForeignKeysFromParticipantIdChanges < ActiveRecord::Migration[7.1]
+  def up
+    remove_foreign_key :participant_id_changes, column: :from_participant_id
+    remove_foreign_key :participant_id_changes, column: :to_participant_id
+  end
+
+  def down; end
+end
--- a/db/migrate/20240918145749_change_from_and_to_participant_ids_to_uuid_in_participant_id_changes.rb
+++ b/db/migrate/20240918145749_change_from_and_to_participant_ids_to_uuid_in_participant_id_changes.rb
@@ -1,8 +1,55 @@
 class ChangeFromAndToParticipantIdsToUuidInParticipantIdChanges < ActiveRecord::Migration[7.1]
   def up
-    change_column :participant_id_changes, :from_participant_id, :uuid
-    change_column :participant_id_changes, :to_participant_id, :uuid
+    add_column :participant_id_changes, :from_participant_uuid, :uuid
+    add_column :participant_id_changes, :to_participant_uuid, :uuid
+
+    execute <<-SQL
+      UPDATE participant_id_changes SET from_participant_uuid = users.ecf_id
+      FROM users WHERE participant_id_changes.from_participant_id = users.id;
+    SQL
+
+    execute <<-SQL
+      UPDATE participant_id_changes SET to_participant_uuid = users.ecf_id
+      FROM users WHERE participant_id_changes.to_participant_id = users.id;
+    SQL
+
+    remove_column :participant_id_changes, :from_participant_id
+    remove_column :participant_id_changes, :to_participant_id
+
+    rename_column :participant_id_changes, :from_participant_uuid, :from_participant_id
+    rename_column :participant_id_changes, :to_participant_uuid, :to_participant_id
+
+    change_column_null :participant_id_changes, :from_participant_id, false
+    change_column_null :participant_id_changes, :to_participant_id, false
+
+    add_index :participant_id_changes, :from_participant_id
+    add_index :participant_id_changes, :to_participant_id
   end
 
-  def down; end
+  def down
+    add_column :participant_id_changes, :from_participant_bigid, :bigint
+    add_column :participant_id_changes, :to_participant_bigid, :bigint
+
+    execute <<-SQL
+      UPDATE participant_id_changes SET from_participant_bigid = users.id
+      FROM users WHERE participant_id_changes.from_participant_id = users.ecf_id;
+    SQL
+
+    execute <<-SQL
+      UPDATE participant_id_changes SET to_participant_bigid = users.id
+      FROM users WHERE participant_id_changes.to_participant_id = users.ecf_id;
+    SQL
+
+    remove_column :participant_id_changes, :from_participant_id
+    remove_column :participant_id_changes, :to_participant_id
+
+    rename_column :participant_id_changes, :from_participant_bigid, :from_participant_id
+    rename_column :participant_id_changes, :to_participant_bigid, :to_participant_id
+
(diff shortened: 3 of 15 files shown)


db/schema.rb
```diff
@@ -147,10 +147,10 @@
     t.decimal "monthly_service_fee", default: "0.0"
     t.decimal "targeted_delivery_funding_per_participant", default: "100.0"
     t.boolean "special_course", default: false, null: false
-    t.uuid "ecf_id"
+    t.uuid "ecf_id", default: -> { "gen_random_uuid()" }, null: false
```
cwrw, line 150: not sure about those migrations and committing them here 🤔

app/models/participant_id_change.rb
```diff
@@ -2,8 +2,8 @@
 
 class ParticipantIdChange < ApplicationRecord
   belongs_to :user
-  belongs_to :from_participant, class_name: "User"
-  belongs_to :to_participant, class_name: "User"
+  belongs_to :from_participant, class_name: "User", primary_key: "ecf_id"
```
cwrw, line 5: should those be `optional: true`?

cwrw: I guess the question is if I add an id that doesn't exist in the user table will I have an issue here?

spec/services/migration/migrators/participant_id_change_spec.rb
```diff
@@ -28,8 +28,8 @@ def setup_failure_state
         participant_id_change = ParticipantIdChange.first
         expect(participant_id_change).to have_attributes({
           user_id: User.find_by(ecf_id: ecf_resource1.user_id).id,
-          from_participant_id: User.find_by(ecf_id: ecf_resource1.from_participant_id).id,
```
cwrw, line 31: could we add a test that checks if the from participant user doesn't exist in npq nothing fails?

app/models/participant_id_change.rb
```diff
@@ -2,8 +2,8 @@
 
 class ParticipantIdChange < ApplicationRecord
   belongs_to :user
-  belongs_to :from_participant, class_name: "User"
-  belongs_to :to_participant, class_name: "User"
+  belongs_to :from_participant, class_name: "User", primary_key: "ecf_id", optional: true
+  belongs_to :to_participant, class_name: "User", primary_key: "ecf_id", optional: true
```
mooktakim, line 6: I don't think we should have `belongs_to` for `from_participant` and `to_participant` as we shouldn't ever use them to find users.

Instead add validation presence for `from_participant_id` and `to_participant_id`

cwrw: that's fair actually 👍

db/migrate/20240917142926_remove_foreign_keys_from_participant_id_changes.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveForeignKeysFromParticipantIdChanges < ActiveRecord::Migration[7.1]
+  def up
+    remove_foreign_key :participant_id_changes, column: :from_participant_id
```
ethax-ross, line 3: Is it worth renaming these to include `ecf_id` like we do for the other ECF ids?

spec/services/migration/migrators/participant_id_change_spec.rb
```diff
@@ -28,11 +28,23 @@ def setup_failure_state
...
+          to_participant_id: ecf_resource1.to_participant_id,
           created_at: ecf_resource1.created_at,
         })
       end
+
+      context "when corresponding from/to users do not exist" do
+        def create_npq_resource(ecf_resource)
+          user = create(:user, ecf_id: ecf_resource.user.id)
+
+          create(:participant_id_change, user:, from_participant_id: SecureRandom.uuid, to_participant_id: SecureRandom.uuid)
```
ethax-ross, line 41: Should we not be creating a participant ID change in ECF where the users don't exist in NPQ reg here?

db/migrate/20240917142926_remove_foreign_keys_from_participant_id_changes.rb
cwrw: good q, I think worth discussing with the team, right now we're adding `ecf_id`s everywhere, after separation some of those will be generated automatically on creation as they are required by the API, some of them not. Eventually we might remove the unnecessary ones after separation from the db and rename the remaining `ecf_id`s to something else. For this one we could follow the same pattern and rename to include `ecf_id` for now 🤔

mooktakim: `from_user_ecf_id` and `to_user_ecf_id`?

cwrw: maybe `from_participant_ecf_id` and `to_participant_ecf_id` instead?

cwrw: discussing in person with @leandroalemao: I think we should probably stick to the same name as they match what we have in the API for now, and we will probably rename the `ecf_id` after separation so doesn't feel like we need to change them here

cwrw (approved):
thanks @leandroalemao, let's run on the migration env before merging

spec/services/migration/migrators/participant_id_change_spec.rb
```diff
@@ -28,11 +28,21 @@ def setup_failure_state
...
           user_id: User.find_by(ecf_id: ecf_resource1.user_id).id,
-          from_participant_id: User.find_by(ecf_id: ecf_resource1.from_participant_id).id,
-          to_participant_id: User.find_by(ecf_id: ecf_resource1.to_participant_id).id,
+          from_participant_id: ecf_resource1.from_participant_id,
+          to_participant_id: ecf_resource1.to_participant_id,
           created_at: ecf_resource1.created_at,
         })
       end
+
+      context "when corresponding from/to users do not exist" do
```
ethax-ross, line 37: This test reads a bit confusingly to me:

> when corresponding from/to users do not exist it creates a new record in NPQ when a matching participant_id_change does not exist

What about something like:

```
context "when a matching participant_id_change and the to/from participants do not exist in NPQ reg" do
  def create_npq_resource(ecf_resource)
    create(:user, ecf_id: ecf_resource.user.id)
  end

  it "creates a new participant_id_change record in NPQ reg" do
    expect { instance.call }.to change(ParticipantIdChange, :count).by(2)
  end
end
```

db/migrate/20240918145749_change_from_and_to_participant_ids_to_uuid_in_participant_id_changes.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeFromAndToParticipantIdsToUuidInParticipantIdChanges < ActiveRecord::Migration[7.1]
+  def up
+    change_column :participant_id_changes, :from_participant_id, :uuid
```
cwrw, line 3: discussed in person: there is a failure on deploy:
```
ActiveRecord::StatementInvalid: PG::DatatypeMismatch: ERROR:  column "from_participant_id" cannot be cast automatically to type uuid
```
we need to revisit this migration :)

spec/factories/participant_id_changes.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :participant_id_change do
     user
-    association :from_participant, factory: :user
-    association :to_participant, factory: :user
+    from_participant_id { SecureRandom.uuid }
+    to_participant_id { SecureRandom.uuid }
```
mooktakim, line 7: minor fix, i think it should be:
```ruby
to_participant_id { user.ecf_id }
```
As we're changing to current user (not all records, but if makes sense to keep this for the factory)

db/migrate/20240918145749_change_from_and_to_participant_ids_to_uuid_in_participant_id_changes.rb
mooktakim: Should we assume this table doesn't already exist and we add data with migration only?
Otherwise, we'll have to have a step to migrate existing data to `ecf_id`.

Locally we'll have to reset db.

leandroalemao: Yeah ideally we should need to, if we had existings records in the table.

- Add a new uuid field
- Run a update to populate the new uuid
- Drop the old bigint ID field
- Rename the new uuid field to the old name

As we don't have existing records in the DB, we should be fine with this "quickfix".

mooktakim (approved):

cwrw (approved):
thanks @leandroalemao ran on migration and all looks good 🙌

ethax-ross (approved):

PR #1729, opened 2024-09-19
Only raise on multiple application TRNs if no teacher profile TRN

ethax-ross:
[Jira-3594](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3594)

### Context

We only want to raise an error for multiple verified `NPQApplication` TRNs if there is not a `teacher_profile.trn` on the ECF user. We will favour the profile TRN over the `NPQApplication` TRNs and fallback on the NPQ user TRN in case there is not a verified TRN in ECF (and the NPQ user already has one).

- Only raise on multiple application TRNs if no teacher profile TRN

### Guidance for review

Screenshot 2024-09-20 at 15 47 40

### DIFF
--- a/app/services/migration/migrators/user.rb
+++ b/app/services/migration/migrators/user.rb
@@ -30,21 +30,22 @@ module Migration::Migrators
       migrate(self.class.ecf_users) do |ecf_user|
         user = find_or_initialize_user(ecf_user)
 
-        trns = unique_validated_trns(ecf_user)
-        validate_multiple_trns!(trns, user)
+        application_trns = validated_application_trns(ecf_user)
+        raise_if_multiple_application_trns_and_no_teacher_profile_trn!(application_trns, ecf_user)
+        ecf_trn = ecf_user.teacher_profile&.trn || application_trns.last
 
         npq_application = most_recent_created_npq_application(ecf_user)
         email = npq_application&.participant_identity&.email
 
         user.update!(
-          trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : trns.last,
+          trn: ecf_trn || user.trn,
           full_name: ecf_user.full_name || user.full_name,
           email:,
           uid: ecf_user.get_an_identity_id || user.uid,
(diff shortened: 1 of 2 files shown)


app/services/migration/migrators/user.rb
```diff
@@ -30,21 +30,22 @@ def call
       migrate(self.class.ecf_users) do |ecf_user|
         user = find_or_initialize_user(ecf_user)
 
-        trns = unique_validated_trns(ecf_user)
-        validate_multiple_trns!(trns, user)
+        application_trns = validated_appication_trns(ecf_user)
```
leandroalemao, line 33: * `validated_application_trns(ecf_user)`

leandroalemao (commented):
Have you got a chance to run the migrators after these changes? So we can see an improvement in errors reports? Tks

app/services/migration/migrators/user.rb
```diff
@@ -30,21 +30,22 @@ def call
...
-          trn: ecf_user.teacher_profile&.trn ? ecf_user.teacher_profile.trn : trns.last,
+          trn:,
           full_name: ecf_user.full_name || user.full_name,
           email:,
           uid: ecf_user.get_an_identity_id || user.uid,
           date_of_birth: npq_application.date_of_birth || user.date_of_birth,
           national_insurance_number: npq_application.nino || user.national_insurance_number,
           active_alert: npq_application.active_alert || user.active_alert,
-          trn_verified: npq_application.teacher_reference_number_verified || user.trn_verified,
+          trn_verified: user&.trn_verified || trn.present?,
```
cwrw, line 48: Not sure about this, I think if there is a teacher profile with a trn we are sure that the trn is verified, if there is 1 trn we can read it from the verified on NPQ application from ECF, after that we can fallback to the one on the user. However as you implemented if there is no trn then this value should be false 👍

leandroalemao (approved):
Looks good tks @ethax-ross

cwrw (approved):

PR #1736, opened 2024-09-23
[CPDLP-3577] data cleanup add dedupe service in user migrator

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3577

* `Migration::UserMigrator` service created to run through various scenarios to find, initialize or merge users.


app/services/migration/migrators/user.rb
```diff
@@ -28,7 +28,7 @@ def ecf_users
 
     def call
       migrate(self.class.ecf_users) do |ecf_user|
-        user = find_or_initialize_user(ecf_user)
+        user = ::Migration::UserMigrator.new(ecf_user).find_initialize_or_merge_user
```
ethax-ross, line 31: Is there a different name we can use? We already have `Migrator::User` which is confusing alongside `Migration::UserMigrator`. What about moving this and `Migration::UserMerger` to namespace them under `Migrator::User` as `Migration::User::Merger` and `Migration::User::Creator` or similar?

spec/services/migration/user_merger_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    context "when from_user has applications" do
+      let!(:application) { create(:application, user: from_user) }
+
+      it "moves applications over to to_user" do
+        expect(application.user).to eq(from_user)
+
+        subject.merge!
+
+        expect(application.reload.user).to eq(to_user)
+        expect(from_user.uid).to be_nil
```
ethax-ross, line 19: Worth calling this out in its own test?

app/services/migration/user_migrator.rb
```diff
@@ -0,0 +1,117 @@
...
+        end
+
+        # Scenario 2 on JIRA (when running on ecf_user with 123)
+        raise_error("User found with ecf_user.get_an_identity_id, but its user.ecf_id linked to another ecf_user that is not an orphan")
+      end
+    end
+
+  private
+
+    def ecf_user_id_user
```
ethax-ross, line 52: I think the method naming for retrieving NPQ users from ECF details are a bit confusing; I expected this one to return an ECF user at first glance. What about following ActiveRecord with something like `user_by_ecf_user_id` and `user_by_ecf_user_get_an_identity_id`?

app/services/migration/user_migrator.rb
```diff
@@ -0,0 +1,117 @@
+module Migration
+  class UserMigrator
+    attr_reader :ecf_user
+
+    def initialize(ecf_user)
+      @ecf_user = ecf_user
+    end
+
+    def find_initialize_or_merge_user
```
ethax-ross, line 9: I think this could just be `find_or_initialize`; the merging that goes on is an implementation detail and having it in the method name makes it unclear if it will return anything if merging

app/services/migration/user_migrator.rb
```diff
@@ -0,0 +1,117 @@
...
+    def raise_error(msg)
+      ecf_user.errors.add(:base, msg)
+      raise ActiveRecord::RecordInvalid, ecf_user
+    end
+
+    def user_does_not_exist?
+      ecf_user_id_user.nil? && ecf_user_gai_id_user.nil?
+    end
+
+    def validate_multiple_users_with_same_ecf_id
```
ethax-ross, line 80: `def raise_on_multiple_ecf_users_with_same_ecf_id!`

app/services/migration/user_migrator.rb
```diff
@@ -0,0 +1,117 @@
+module Migration
+  class UserMigrator
+    attr_reader :ecf_user
+
+    def initialize(ecf_user)
+      @ecf_user = ecf_user
+    end
+
+    def find_initialize_or_merge_user
+      if user_does_not_exist?
```
ethax-ross, line 10: Can we make these methods clear on which source we're looking at, so `ecf_user_does_not_exist?` and `both_ecf_ids_return_same_user?` etc

app/services/migration/migrators/user.rb
mooktakim: Renamed to `Migration::Users::Creator` and `Migration::Users::Merger`

app/services/migration/user_migrator.rb
mooktakim: renamed👍

app/services/migration/user_migrator.rb
mooktakim: Yes, i started off with that naming but I didn't like it ending with `_id`. I'll rename them.

app/services/migration/users/creator.rb
```diff
@@ -0,0 +1,119 @@
...
+      def initialize(ecf_user)
+        @ecf_user = ecf_user
+      end
+
+      def find_or_initialize
+        if user_does_not_exist_on_npq?
+          return ::User.new(ecf_id: ecf_user.id)
+        end
+
+        raise_on_multiple_ecf_users_with_same_ecf_id!
```
ethax-ross, line 15: Should this go right at the top? otherwise we would miss those where the user doesn't exist in ECF potentially? 🤔

ethax-ross (commented):
I think this is looking good; has it been ran on the migration env?

app/services/migration/users/creator.rb
cwrw, line 15: should this fail first 🤔 as we did previously?

mooktakim: I can move it, but we won't need this once `ecf_id` uniqueness validation is added.

app/services/migration/users/merger.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      def initialize(from_user:, to_user:)
+        @from_user = from_user
+        @to_user = to_user
+      end
+
+      def merge!
+        ApplicationRecord.transaction do
+          from_user.applications.update!(user: to_user)
+          from_user.participant_id_changes.update!(user: to_user) # not sure?
```
cwrw, line 14: I think this migrator happens after users so will be ok and not needed here

mooktakim: @cwrw 
I wanted this service to be "complete", maybe we'll use it in other places.

leandroalemao (commented):
Logic seems fine with others comments.. have you tried to run this in migration env to see the results? Tks

cwrw (approved):
thanks @mooktakim 🙌

ethax-ross (approved):

PR #1737, opened 2024-09-23
[CPDLP-3493] NPQ Internal API - Add feature flag instead of application config

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3493

- Added feature: `Feature.ecf_api_disabled?` with flipper


app/services/feature.rb
```diff
@@ -51,5 +54,9 @@ def disable_registration!
     def enable_registration!
       Flipper.disable(REGISTRATION_DISABLED)
     end
+
+    def ecf_api_disabled?
```
leandroalemao, line 58: Hi @mooktakim Would be nice to add a spec for this method?

leandroalemao (commented):
Also would be nice to add a method to enable the feature flag? ie: `disable_ecf_api!`

ethax-ross (commented):
Looks good past Leandro's comment

cwrw:
Looks good to me too, @leandroalemao regarding toggling I was wondering if we do need toggling methods as those are given to us by `Flipper` so need to add a wrapper method?

mooktakim:
> Looks good to me too, @leandroalemao regarding toggling I was wondering if we do need toggling methods as those are given to us by `Flipper` so need to add a wrapper method?

I believe flipper has an admin UI to enable/disable features, unlikely to need the toggle:
`/admin/feature_flags/features`

spec/services/feature_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe Feature do
```
ethax-ross, line 3: There's already a spec for this; can we consolidate them? https://github.com/DFE-Digital/npq-registration/blob/main/spec/lib/services/feature_spec.rb

ethax-ross: might be worth improving the general coverage of this service while you're in there; looks like only the `registration_closed?` method is tested at the moment

mooktakim: 🤦 i will merge it into the correct location.

cwrw (approved):

ethax-ross (approved):

PR #1744, opened 2024-09-26
[CPDLP-3494] Disable Internal APIs with feature flag

leandroalemao:
### Context

Ticket: [CPDLP-3494](https://dfedigital.atlassian.net/browse/CPDLP-3494)

We currently have the internal APIs disabled by a rails configuration but that was mostly for the separation environments and we didn’t retain previous logic like emails being sent out.

Disable Internal APIs using a feature flag instead.


app/services/handle_submission_for_store.rb
```diff
@@ -157,8 +157,8 @@ def headteacher_status
     end
   end
 
-  def enqueue_job
-    ApplicationSubmissionJob.perform_later(user:, email_template:)
+  def enqueue_job(application)
+    Feature.ecf_api_disabled? ? SendApplicationSubmissionEmailJob.perform_later(application:, email_template:) : ApplicationSubmissionJob.perform_later(user:, email_template:)
```
mooktakim, line 161: Minor point, can we break this into multi-line:
```ruby
if Feature.ecf_api_disabled?
  SendApplicationSubmissionEmailJob.perform_later(application:, email_template:) 
else
  ApplicationSubmissionJob.perform_later(user:, email_template:)
end
```

mooktakim (approved):
otherwise all good

app/jobs/application_submission_job.rb
```diff
@@ -4,6 +4,8 @@ class ApplicationSubmissionJob < ApplicationJob
   queue_as :default
 
   def perform(user:, email_template:)
+    return if Feature.ecf_api_disabled?
```
ethax-ross, line 7: Is it worth queueing your new `ApplicationSubmissionMailer` from here to avoid the duplication? or pulling it out entirely and changing the caller to something like:

```
def enqueue_jobs(application)
  ApplicationSubmissionJob.perform_later(user:, email_template:) unless if Feature.ecf_api_disabled?
  SendApplicationSubmissionEmailJob.perform_later(application:, email_template:)
end
```

spec/jobs/send_application_submission_email_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:course) { create(:course, :leading_teaching) }
+  let(:application) { create(:application, course:, raw_application_data: { "funding_amount" => "123" }) }
+
+  subject(:job) { described_class.new(application:, email_template: "b8b53310-fa6f-4587-972a-f3f3c6e0892e") }
+
+  describe "#perform" do
+    context "when ecf_api_disabled flag is toggled off" do
+      before { Flipper.disable(Feature::ECF_API_DISABLED) }
+
+      it "returns nil" do
```
ethax-ross, line 13: Would it be better to test that `application_submitted_mail` isn't called here? the return value might be `nil` regardless 🤔

spec/lib/services/ecf/npq_profile_updater_spec.rb
```diff
@@ -117,16 +102,62 @@
...
+    it "calls ecf to update the eligible_for_funding attribute" do
+      subject.call
+      expect(update_ecf_stub).to have_been_requested
+    end
+
+    context "when ecf_api_disabled flag is toggled on" do
+      before { Flipper.enable(Feature::ECF_API_DISABLED) }
+
+      it "returns nil" do
+        expect(subject.call).to be_nil
```
ethax-ross, line 130: Would it be more reliable to negate the test above here?:

```
subject.call
expect(update_ecf_stub).not_to have_been_requested
```

app/jobs/application_submission_job.rb
```diff
@@ -4,6 +4,8 @@ class ApplicationSubmissionJob < ApplicationJob
...
+
     if user.synced_to_ecf?
       update_ecf_user_details(user:)
     else
```
cwrw, line 11: not implemented in this pr but it's interesting that we call all applications on a user to send emails/create applications in ECF for, wonder if this is the cause of the duplication we see in ECF, or if there is a reason we are doing this for all applications rather than the new one created

app/models/user.rb
```diff
@@ -114,7 +114,7 @@ def get_an_identity_user
   end
 
   def ecf_user
-    return if ecf_id.blank? || Rails.application.config.npq_separation[:ecf_api_disabled]
+    return if ecf_id.blank? || Feature.ecf_api_disabled?
```
cwrw, line 117: I think we'll also need to disable the following as they call this method to avoid errors:
```
GetAnIdentityIdSyncJob
EcfUserUpdater
```

app/services/ecf/npq_profile_updater.rb
```diff
@@ -18,7 +18,7 @@ def call
     end
 
     def tsf_data_field_update
-      return if Rails.application.config.npq_separation[:ecf_api_disabled]
+      return if Feature.ecf_api_disabled?
```
cwrw, line 21: maybe we can add something similar to the call as well, even if it's called from a mass updater/rake task just incase?

cwrw:
Past Ross's comments in the specs looks good to me, thanks @leandroalemao :)

cwrw (approved):

ethax-ross (approved):

mooktakim (approved):

PR #1745, opened 2024-09-26
[CPDNPQ-2078] Admin > Participants > Participant page

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2078
(n.b. discussion here around which attributes are in scope for this screen versus applications#show)

### Screenshot

Register for a national professional qualification · 1 20pm · 09-26

### DIFF
--- a/spec/features/npq_separation/admin/users_spec.rb
+++ b/spec/features/npq_separation/admin/users_spec.rb
@@ -71,7 +71,7 @@ RSpec.feature "User administration", type: :feature do
     end
 
     scenario "shows a summary of each user application" do
-      applications = 2.times.map { create(:application, user:) }
+      applications = %i[headship senco].map { create(:application, user:, course: create(:course, _1)) }
       visit npq_separation_admin_user_path(user)
 
       expect(page).to have_css("h1", text: "Applications")


app/views/npq_separation/admin/users/_application.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      row.with_value(text: application.school.urn)
+    end
+
+    sl.with_row do |row|
+      row.with_key(text: "School UKPRN")
+      row.with_value(text: application.school.ukprn)
+    end
+
+    sl.with_row do |row|
+      row.with_key(text: "Funded Place?")
```
cwrw, line 37: maybe this has been agreed, but should funded place have a question mark in the key? Also we have three values for funded place `nil`, `true` and `false`, funded place only is set on applications from cohort 2024 onwards on approved applications. Otherwise the value is `nil` as in previous cohorts it's not applicable, maybe we should distinguish those in the method below? If this has been already agreed then all good

cwrw (commented):
one comment otherwise all good :)

app/views/npq_separation/admin/users/_application.html.erb
cwrw: preference is to have that covered if that's ok... eyeing sonar cloud percentages :D

cwrw (approved):
just the last thing about the test otherwise looking good

cwrw (approved):
🙌

PR #1752, opened 2024-09-30
Run migration on a nightly schedule

ethax-ross:
[Jira-3512](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3512)

### Context

We want to refresh the migration DB every night so that we can run the migration on a schedule.

- Refresh migration DB nightly

Refresh migration DB at midnight.

- Run migration nightly

We want to run the data migration on a nightly basis to monitor how it
performs/the results.

Run migration at 12:30am every morning; this is 30 minutes after the migration DBs are refreshed.

### DIFF
--- a/app/jobs/crons/migration_job.rb
+++ b/app/jobs/crons/migration_job.rb
@@ -1,0 +1,17 @@
+class Crons::MigrationJob < CronJob
+  queue_as :high_priority
+
+  include Sentry::Cron::MonitorCheckIns
+
+  # run at 12:30 AM every day
+  self.cron_expression = "30 0 * * *"
+
+  sentry_monitor_check_ins slug: "migration"
+
+  def perform
+    return unless Rails.application.config.npq_separation[:migration_enabled]
+
+    migrator = Migration::Coordinator.new
+    migrator.migrate!
+  end
+end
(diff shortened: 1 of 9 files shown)


app/jobs/crons/migration_job.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  include Sentry::Cron::MonitorCheckIns
+
+  # run at 12:30 AM every day
+  self.cron_expression = "30 0 * * *"
+
+  sentry_monitor_check_ins slug: "migration"
+
+  def perform
+    return unless Rails.application.config.npq_separation[:migration_enabled]
```
cwrw, line 12: does this mean we'll need to manually go in and check migrations vs github actions showing us if they passed/fail instead?

cwrw: also can we disable migration in staging/review otherwise we will start getting weird failures as it's enabled there?

leandroalemao: Also maybe better off using a feature flag instead so we can disable without having to 
make changes to config/re-deploy?

ethax-ross: I've updated to disable in non-migration/dev environments; I don't mind if we want to switch to a feature flag instead of config but suggest we pick that up separately 👍

leandroalemao (approved):

cwrw (approved):

PR #1761, opened 2024-10-01
[CPDLP-3509] Ensure ECF ID generation on all new records in NPQ

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3509

We would like to ensure that after separation, any new records created for the API in the db have a UUID which will be surfaced correctly.

- Ensure ecf id is populated for new applications and users after separation
- Ensure ecf id is required for applications and users after separation
- Fix migrations (db recreation was failing due a default scope added to IttProvider
- Remove default from contract templates db table
- Add missing index to application

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -36,22 +36,13 @@ class Application < ApplicationRecord
 
   validate :schedule_cohort_matches
   # TODO: add constraints into the DB after separation
-  validates :ecf_id,
-            presence: { message: "Enter an ECF ID" },
-            uniqueness: {
-              case_sensitive: false,
-              message: "ECF ID must be unique",
-            }, if: -> { Feature.ecf_api_disabled? }
+  validates :ecf_id, presence: true, if: -> { Feature.ecf_api_disabled? }
 
   after_commit :touch_user_if_changed
 
   # TODO: remove this and add default: "gen_random_uuid()" in the DB after separation
-  before_validation(on: :create) do
-    self.ecf_id = if Feature.ecf_api_disabled? & !ecf_id
-                    SecureRandom.uuid
-                  else
-                    ecf_id
-                  end
+  before_validation do
+    self.ecf_id ||= SecureRandom.uuid if Feature.ecf_api_disabled? && ecf_id.blank?
   end
 
   enum kind_of_nursery: {
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -35,8 +35,24 @@ class Application < ApplicationRecord
   scope :with_targeted_delivery_funding_eligibility, -> { where(targeted_delivery_funding_eligibility: true) }
 
   validate :schedule_cohort_matches
+  # TODO: add constraints into the DB after separation
+  validates :ecf_id,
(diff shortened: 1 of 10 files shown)


app/models/application.rb
```diff
@@ -35,9 +35,25 @@ class Application < ApplicationRecord
...
+            presence: { message: "Enter an ECF ID" },
+            uniqueness: {
+              case_sensitive: false,
+              message: "ECF ID must be unique",
+            }, if: -> { Feature.ecf_api_disabled? }
 
   after_commit :touch_user_if_changed
 
+  # TODO: remove this and add default: "gen_random_uuid()" in the DB after separation
+  before_validation(on: :create) do
```
ethax-ross, line 49: Could we not do this in a migration to set the default UUID as we do in some of the other tables?

```
add_column :statements, :ecf_id, :uuid, default: "gen_random_uuid()", null: false
```

This way if the UUID isn't set by ECF it will be auto-generated? it might be worth adding a couple of tests to ensure if we preset a uuid it doesn't change on creation if already populated but is set if `nil`?

leandroalemao: Well yeah this is the idea, that's why I added the TODO message to be done after separation.. but we can't do it now as there's logic behind it in `ApplicationSubmissionJob` where it fetches new applications still without ecf_id to be synced with ECF.. same for user really.

mooktakim: I suppose the key thing here is that it generates `ecf_id` only when the feature flag is on.

mooktakim: Should it only be on `create`?
I'm thinking of scenario when Application that's pending in NPQ, they haven't been accepted yet, so they won't be `ecf_id`. They later get accepted (after migration) and won't ever get `ecf_id` generated.

Maybe set it to all `before_validation` and when application is accepted?

leandroalemao: Tks @mooktakim. I've updated the callback to all `before_validation` which would make no harm and will cover when an application is accepted as well. So we'll have an `ecf_id` set if none after the feature flag is on whenever an application is created or updated.

app/models/application.rb
```diff
@@ -35,9 +35,16 @@ class Application < ApplicationRecord
   scope :with_targeted_delivery_funding_eligibility, -> { where(targeted_delivery_funding_eligibility: true) }
 
   validate :schedule_cohort_matches
+  # TODO: add constraints into the DB after separation
+  validates :ecf_id, presence: true, uniqueness: { case_sensitive: false }, if: -> { Feature.ecf_api_disabled? }
```
cwrw, line 39: can we remove the uniqueness index for now from both users/applications, as they are not and need to be added by the NPQ team when the data is cleaned up?

cwrw (commented):
thanks @leandroalemao looks good, as discussed we will be adding a ticket to populate the empty ecf ids in the migration for both applications/users. I think we need to remove the unique index on both as this will to be fixed by the NPQ team before we can add those. Thanks!

cwrw (approved):

mooktakim (approved):

PR #1774, opened 2024-10-02
[CPDLP-3516] Amend applications migrator to set the user id to match the one in ecf

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3516

In the data migrators we currently raise an error if the NPQ user on application is different than the ECF user on application in the application migrator. However we have cases where transfers happened that have not been reflected in NPQ.

- amend the application migrator to set the user id to match the one in ecf
- remove the failure on application if the users are different between ecf and npq


cwrw (approved):
Looks great, before merging do you mind running this on the migration env just to see the failures go, after that feel free to merge :)

PR #1775, opened 2024-10-03
Fix migration report GitHub action

ethax-ross:
[Jira-3512](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3512)

### Context

I couldn't test this action until it was merged into main, so there were inevitably a _few_ issues 🙂 

Contains various fixes and minor tweaks:

- Pass `production` environment instead of `migration` (`az login` fails with `migration` - something probably off with the GitHub secret here)
- Install `kubectl` and run commands directly instead of via `make` (the `make` commands output a load of other info that messes up the comparisons/output we want)
- Add step for Slack webhook secret
- Add duration to post and format failures 

### Guidance to review

Screenshots of the Slack messages below - I can't seem to change the Slack user it is posted as; we could ask the infra team but I don't think its worth the effort given this is short-lived.

| Success | Failure |
| -- | -- |
| Screenshot 2024-10-03 at 12 28 00 | Screenshot 2024-10-03 at 12 26 30 |

### DIFF
--- a/.github/workflows/report-migration-results.yml
+++ b/.github/workflows/report-migration-results.yml
@@ -4,11 +4,11 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: GitHub environment to report from
+        description: GitHub environment to run from
         type: choice
-        default: migration
+        default: production
         options:
-          - migration
+          - production
         required: true
   schedule:
     - cron: "0 8 * * *" # Run at 8am, daily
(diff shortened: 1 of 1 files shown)


cwrw (approved):
🙌 nice one!

PR #1776, opened 2024-10-03
[CPDLP-3607] Add logic to recalculate `participant_outcome_state` after separation

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3607

NPQ currently through an internal API to ECF gets outcomes for users to surface on the accounts pages. We would like to retain the same logic when internal APIs are turned off

- Currently `participant_outcome_state` is being updated via a scheduled job which will be disabled after separation.
- We have two services that creates outcomes, so do an application `participant_outcome_state` needs to be updated:
  - When a new declaration is created and then an outcome is created.
  - When a declaration is voided.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -188,6 +188,12 @@ class Application < ApplicationRecord
     eligible_for_dfe_funding?(with_funded_place: true)
   end
 
+  def participant_outcome_state
+    return read_attribute(:participant_outcome_state) unless Feature.ecf_api_disabled?
+
+    declarations.completed.billable_or_voidable.latest_first.first&.participant_outcomes&.latest&.state
+  end
+
 private
 
   def funding_eligibility(with_funded_place:)
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -188,8 +188,8 @@ class Application < ApplicationRecord
     eligible_for_dfe_funding?(with_funded_place: true)
   end
 
-  def participant_outcome_state
-    return read_attribute(:participant_outcome_state) unless Feature.ecf_api_disabled?
+  def latest_participant_outcome_state
+    return participant_outcome_state unless Feature.ecf_api_disabled?
 
     declarations.completed.billable_or_voidable.latest_first.first&.participant_outcomes&.latest&.state
   end
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -494,4 +494,32 @@ RSpec.describe Application do
       end
     end
   end
+
+  describe "#participant_outcome_state" do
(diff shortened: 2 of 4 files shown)


app/services/participant_outcomes/create.rb
```diff
@@ -35,7 +35,9 @@ def create_outcome
         @created_outcome = if outcome_already_exists?
                              latest_existing_outcome
                            else
-                             build_outcome.tap(&:save!)
+                             outcome = build_outcome.tap(&:save!)
+                             outcome.application.update!(participant_outcome_state: state)
```
cwrw, line 39: instead of doing this here, I was thinking we add a feature flag where we read from `participant_outcome_state` and instead read it from the declaration outcomes instead? this way we will avoid having to backfill any data, happy to discuss this one

app/services/participant_outcomes/void.rb
```diff
@@ -10,11 +10,15 @@ class Void
...
-        completion_date: declaration_date,
-        state: "voided",
-      )
+      ApplicationRecord.transaction do
+        outcome = ParticipantOutcome.create!(
+          declaration:,
+          completion_date: declaration_date,
+          state: "voided",
+        )
+        outcome.application.update!(participant_outcome_state: "voided")
```
cwrw, line 19: not sure we handle voided in the `participant_outcome_state` and should probably nullify, but let's discuss general approach

app/models/application.rb
```diff
@@ -181,6 +181,12 @@ def fundable?
     eligible_for_dfe_funding?(with_funded_place: true)
   end
 
+  def participant_outcome_state
+    return read_attribute(:participant_outcome_state) unless Feature.ecf_api_disabled?
+
+    declarations.completed.billable_or_voidable.latest_first.first&.participant_outcomes&.latest&.state
```
cwrw, line 187: `billable_or_changeable` rather than voidable, as that includes ineligible?

spec/models/application_spec.rb
```diff
@@ -460,4 +460,32 @@ def subject.transient_previously_funded
...
+
+    context "when ecf_api_disabled flag is toggled off" do
+      before { Flipper.disable(Feature::ECF_API_DISABLED) }
+
+      it "returns the attribute value" do
+        expect(subject).to eq("anything")
+      end
+    end
+
+    context "when ecf_api_disabled flag is toggled on" do
```
cwrw, line 483: Could we add more testing to cover possibly voided completed declarations which should be ignored, as well as have other declarations past completed? Also if there are no completed declarations as well shouldn't return anything?

cwrw (approved):
🙌

PR #1777, opened 2024-10-03
[CPDLP-3608] Fix bug with duplicate records in NPQ

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3608

* `Participants::Query` now by default sorts by `created_at`
  * Adding `applications.accepted_at` to sorting forces query to return multiple `User` records as it has to join `Application`
* Added `distinct` to the query
* Added test to replicate the bug

### DIFF
--- a/app/services/participants/query.rb
+++ b/app/services/participants/query.rb
@@ -60,11 +60,12 @@ module Participants
     end
 
     def order_by
-      sort_order(sort:, model: User, default: { accepted_at: :asc })
+      sort_order(sort:, model: User, default: { created_at: :asc })
     end
 
     def all_participants
       User
+        .distinct
         .joins(:applications).merge(Application.accepted)
         .includes(
           :participant_id_changes,
--- a/spec/services/participants/query_spec.rb
+++ b/spec/services/participants/query_spec.rb
@@ -232,12 +232,6 @@ RSpec.describe Participants::Query do
 
       subject(:participants) { query.participants }
 
-      before do
-        participant1.applications.first.update!(accepted_at: 1.day.ago)
-        participant2.applications.first.update!(accepted_at: 10.days.ago)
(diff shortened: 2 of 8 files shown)


app/services/participants/query.rb
```diff
@@ -60,11 +60,12 @@ def where_from_participant_id_is(from_participant_id)
     end
 
     def order_by
-      sort_order(sort:, model: User, default: { accepted_at: :asc })
+      sort_order(sort:, model: User, default: { created_at: :asc })
```
leandroalemao, line 63: On the sorting order.. this change rolls back the one made here? https://dfedigital.atlassian.net/browse/CPDLP-3349

cwrw (commented):
@mooktakim just to confirm other queries are not returning duplicates? could we return distinct on all the others please?

spec/services/participants/query_spec.rb
```diff
@@ -222,8 +238,8 @@
         participant3.applications.first.update!(accepted_at: 2.days.ago)
       end
 
-      it "orders applications by accepted_at in ascending order" do
-        expect(query.participants).to eq([participant2, participant3, participant1])
+      it "by default orders by created_at in ascending order" do
```
cwrw, line 241: could we update the before/setup to update the created at instead as the setup is now confusing?

app/services/participants/query.rb
cwrw: Having a rethink about this, past Mook's change which was blocked by this, I think we should be ordering on the user created at, as this is a participant/user endpoint rather than application endpoint. I think revisiting that ticket we should probably have kept it by user

cwrw (approved):

leandroalemao (approved):
Looking good tks @mooktakim

PR #1781, opened 2024-10-04
[CPDNPQ-2076] Admin Management

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2076

Implement admin management currently at `/admin/admins` as new version at `/npq-separation/admin/admins`

### DIFF
--- a/spec/components/npq_separation/navigation_structures/admin_navigation_structure_spec.rb
+++ b/spec/components/npq_separation/navigation_structures/admin_navigation_structure_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe NpqSeparation::NavigationStructures::AdminNavigationStructure, ty
       "Finance" => "/npq-separation/admin/finance/statements",
       "Schools" => "/npq-separation/admin/schools",
       "Lead providers" => "/npq-separation/admin/lead-providers",
-      "Settings" => "#",
+      "Admins" => "/npq-separation/admin/admins",
     }.each_with_index do |(name, href), i|
       it "#{name} with href #{href} is at position #{i + 1}" do
         expect(subject[i].name).to eql(name)


config/routes.rb
```diff
@@ -244,7 +244,7 @@
         end
 
         resources :lead_providers, only: %i[index show], path: "lead-providers"
-        resources :admins, only: %i[index]
+        resources :admins, except: :edit
```
peteryates, line 247: I think referring to 'admins' inside the admin section could possibly be a bit confusing. As these people will be DfE staff, maybe `/admin/staff`?

```suggestion
        resources :admins, except: :edit, path: `staff`
```

peteryates:
Looking good so far. I left a suggestion on the naming as having an 'admins' section inside the admin section feels like it could be a bit confusing. Perhaps referring to DfE people as 'staff' would make things easier? (there's a fair chance that could just make everything _more_ confusing - worth having a chat with the team).

db/migrate/20241001141442_add_unique_index_to_admins.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToAdmins < ActiveRecord::Migration[7.1]
+  def change
+    add_index :admins, :email, unique: true
```
cwrw, line 3: are we sure in all envs when we merge this down this migration won't fail? staging/separation/sandbox/production? might be worth checking

app/controllers/npq_separation/admin/admins_controller.rb
```diff
@@ -1,3 +1,61 @@
...
+  def index
+    @admins = Admin.all
+  end
+
+  def new
+    @admin = Admin.new
+  end
+
+  def create
+    @admin = Admin.new(admin_params)
```
cwrw, line 13: `find_or_initialize_by` email?

app/controllers/npq_separation/admin/admins_controller.rb
```diff
@@ -1,3 +1,61 @@
...
+      flash[:error] = t(".failure", email: @admin.email)
+    end
+
+    redirect_to action: :index
+  end
+
+  def destroy
+    @admin = Admin.find(params[:id])
+
+    if @admin.super_admin?
```
cwrw, line 38: there is logic in the current interface to avoid deleting current admin `current_admin`, should we retain that?

cwrw (commented):
Thanks @rwrrll , I'm just curious why we opted to rebuild this page in separation when we aren't hiding the old one that exists now with similar functionality? is there a reason for reimplementing this now vs post separation?

peteryates:
@cwrw aren't we deleting the old admin section and replacing it with the new one at the point of separation?

cwrw:
> @cwrw aren't we deleting the old admin section and replacing it with the new one at the point of separation?

In an ideal world yes, but unfortunately due to time constraints we are only adding missing functionality to the admin which includes the finance dashboards and all the other ECF tools. After separation the admin dashboard can be fully cleaned up and moved to the new world.

Appreciate this can be confusing but we have discussed this with the NPQ team and they are happy with the approach for now.

PR #1790, opened 2024-10-08
Change accepted scope to a check in the select body

cwrw:
### Context

The current accepted scope adds an n+1 to the API response for every participant as it generates a query. Instead add the check in the select body to avoid n+1s

Change accepted scope to a check in the select body

failures:

### DIFF
--- a/app/serializers/api/participant_serializer.rb
+++ b/app/serializers/api/participant_serializer.rb
@@ -104,7 +104,7 @@ module API
         def applications(object, options)
           return Application.none unless options[:lead_provider]
 
-          object.applications.accepted.select { |application| application.lead_provider_id == options[:lead_provider].id }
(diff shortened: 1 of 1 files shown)


app/serializers/api/participant_serializer.rb
```diff
@@ -104,7 +104,7 @@ class << self
         def applications(object, options)
           return Application.none unless options[:lead_provider]
 
-          object.applications.accepted.select { |application| application.lead_provider_id == options[:lead_provider].id }
+          object.applications.select { |application| application.lead_provider_approval_status == "accepted" && application.lead_provider_id == options[:lead_provider].id }
```
mooktakim, line 107: minor, can we change to `application.accepted_lead_provider_approval_status?`

mooktakim (approved):
👍

PR #1802, opened 2024-10-11
[CPDLP-3519] Add contract for course validation to change schedule and declarations create services

leandroalemao:
### Context

Ticket: [CPDLP-3519](https://dfedigital.atlassian.net/browse/CPDLP-3519)

We want to add current contract logic in record declaration and change schedules.

Add contract for course validation to change schedule and declarations create services.

[CPDLP-3519]: https://dfedigital.atlassian.net/browse/CPDLP-3519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/participants/change_schedule.rb
+++ b/app/services/participants/change_schedule.rb
@@ -8,7 +8,7 @@ module Participants
 
     validates :schedule_identifier, presence: true
     validate :validate_new_schedule_found
-    # validates :cohort, npq_contract_for_cohort_and_course: true # TODO we don't have NPQ Contract yet
+    validates :cohort, contract_for_cohort_and_course: true
     validate :validate_not_withdrawn
     validate :validate_new_schedule_valid_with_existing_declarations
     validate :validate_changing_to_different_schedule
@@ -28,14 +28,14 @@ module Participants
       true
     end
 
-  private
-
-    def new_cohort
-      @new_cohort ||= cohort ? Cohort.find_by(start_year: cohort) : fallback_cohort
+    def cohort
+      @cohort ||= super ? Cohort.find_by(start_year: super) : fallback_cohort
     end
 
+  private
+
     def new_schedule
-      @new_schedule ||= Schedule.find_by(identifier: schedule_identifier, cohort: new_cohort)
+      @new_schedule ||= Schedule.find_by(identifier: schedule_identifier, cohort:)
     end
 
     def fallback_cohort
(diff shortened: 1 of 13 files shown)


app/services/participants/change_schedule.rb
```diff
@@ -8,7 +8,7 @@ class ChangeSchedule < Action
 
     validates :schedule_identifier, presence: true
     validate :validate_new_schedule_found
-    # validates :cohort, npq_contract_for_cohort_and_course: true # TODO we don't have NPQ Contract yet
+    validates :cohort, contract_for_cohort_and_course: true
```
cwrw, line 11: having a think about this, do we really need it here, in ECF it's for funding cap, here we get the cap from cohort so maybe we don't need to check the contract after all?

spec/services/participants/change_schedule_spec.rb
```diff
@@ -56,14 +58,9 @@
...
-        before do
-          create(:schedule, :npq_leadership_autumn, cohort:)
-        end
+        let(:new_schedule) { create(:schedule, :npq_leadership_spring, cohort:) }
 
-        it "is invalid and returns an error message" do
-          expect(subject.change_schedule).to be_truthy
-          expect(subject).to have_error(:schedule_identifier, :already_on_the_profile, "Selected schedule is already on the profile")
-        end
+        it { is_expected.to have_error(:schedule_identifier, :already_on_the_profile, "Selected schedule is already on the profile") }
```
cwrw, line 63: any reason we changed this test? as it was testing attempting the change twice? we could set up a new contract for that course?

cwrw (commented):
looks good, just a minor question :)

spec/services/participants/change_schedule_spec.rb
leandroalemao: This spec was failing after the changes but I realised it was kind confusing as well with the first expect to be truthy and the second expecting an error:

`expect(subject.change_schedule).to be_truthy`
`expect(subject).to have_error(:schedule_identifier, :already_on_the_profile, "Selected schedule is already on the profile")`

Make the changes to keep it in line with the test above as well.

cwrw (approved):

mooktakim (approved):

PR #1803, opened 2024-10-11
[CPDNPQ-2140] Show outcomes in separated admin/applications#show

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2140

Post-separation "We can view the outcomes recorded, if it was sent to TRA and status if we get a 200 response or not"

### DIFF
--- a/spec/components/npq_separation/admin/outcomes_table_component_spec.rb
+++ b/spec/components/npq_separation/admin/outcomes_table_component_spec.rb
@@ -26,8 +26,8 @@ RSpec.describe NpqSeparation::Admin::OutcomesTableComponent, type: :component do
     context "when the outcome has not been sent to TRA" do
       let(:outcomes) { [create(:participant_outcome)] }
 
-      it "renders 'No'" do
-        expect(cell_text).to eq("No")
+      it "renders 'N/A'" do
+        expect(cell_text).to eq("N/A")
       end
     end
 


app/components/npq_separation/admin/outcomes_table_component.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      def rows
+        outcomes.sort_by(&:created_at).reverse.map do |outcome|
+          [
+            outcome.state.humanize,
+            outcome.completion_date.to_fs(:govuk_short),
+            outcome.created_at.to_date.to_fs(:govuk),
+            outcome.sent_to_qualified_teachers_api_at.try(:to_fs, :govuk_short).presence || "No",
+            helpers.boolean_red_green_nil_tag(outcome.qualified_teachers_api_request_successful),
```
cwrw, line 37: I think we have some missing logic vs what is in ECF regarding when we should show this value, let's catchup tomorrow on this. Also worth mentioning that usually outcomes are associated with a declaration so maybe we should show them after their respective declaration? Anyway let's chat tomorrow :)

app/models/participant_outcome.rb
```diff
@@ -59,6 +59,10 @@ def has_passed?
     passed_state?
   end
 
+  def latest_for_declaration?
+    declaration.participant_outcomes.order(created_at: :desc).first == self
```
cwrw, line 63:
```suggestion
    declaration.participant_outcomes.latest == self
```

app/components/npq_separation/admin/outcomes_table_component.rb
```diff
@@ -0,0 +1,47 @@
...
+        ]
+      end
+
+      def rows
+        outcomes.sort_by(&:created_at).reverse.map do |outcome|
+          [
+            outcome.state.humanize,
+            outcome.completion_date.to_fs(:govuk_short),
+            outcome.created_at.to_date.to_fs(:govuk),
+            outcome.sent_to_qualified_teachers_api_at.try(:to_fs, :govuk_short).presence || "No",
```
cwrw, line 32:
```suggestion
            outcome.sent_to_qualified_teachers_api_at.try(:to_fs, :govuk_short).presence || "N/A",
```

app/components/npq_separation/admin/outcomes_table_component.rb
```diff
@@ -0,0 +1,47 @@
...
+            recorded_by_tra_api(outcome),
+          ]
+        end
+      end
+
+      def recorded_by_tra_api(outcome)
+        if outcome.qualified_teachers_api_request_successful.nil?
+          outcome.latest_for_declaration? ? "Pending" : "N/A"
```
cwrw, line 40: I don't mind but worth noting ignoring the colour here but setting it in the success tags 🤔

cwrw (approved):
looks good, just some minor comments

PR #1806, opened 2024-10-14
[CPDNPQ-2148] Add declarations to Application details screen

jebw:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2148

Admin users need to be able to view declaration details for the Participants Application

* Show training status and course identifier in the application details part of the screen
* Added 1 card per declaration to the participant applications details screen
* Changed the controller to eager load an applications declarations and the declarations associated lead provider

failures:


app/views/npq_separation/admin/applications/show.html.erb
```diff
@@ -197,3 +206,52 @@
...
+<hr class="govuk-section-break govuk-section-break--l">
+
+<h2 class="govuk-heading-m govuk-!-margin-top-9">Declarations</h2>
+
+<% @declarations.each do |declaration| %>
+  <%= govuk_summary_card(title: declaration.declaration_type.humanize) do |summary_card| %>
+    <%= summary_card.with_summary_list do |sl| %>
+      <% sl.with_row do |row| %>
+        <% row.with_key(text: "Declaration ID" ) %>
+        <% row.with_value(text: declaration.id) %>
```
cwrw, line 219: can we also surface the ecf id here as well?

app/views/npq_separation/admin/applications/show.html.erb
```diff
@@ -197,3 +206,52 @@
...
+      <% end %>
+
+      <% sl.with_row do |row| %>
+        <% row.with_key(text: "Declaration date") %>
+        <% row.with_value(text: declaration.declaration_date.to_fs(:govuk_short)) %>
+      <% end %>
+
+      <% sl.with_row do |row| %>
+        <% row.with_key(text: "Lead provider") %>
+        <% row.with_value(text: declaration.lead_provider.name) %>
```
cwrw, line 234: could we also surface cohort as well for declarations, because a provider can submit a declaration for a cohort, then decide to change that cohort and void the declaration, so would be good to surface the separate cohort on the declaration as the application might be different

cwrw (commented):
looks good, just some minor comments

cwrw (approved):

PR #1809, opened 2024-10-15
Schedule cron jobs on db rake tasks

ethax-ross:
Currently, if we add a new cron job we need to remember to manually schedule it the first time or it won't ever run. To avoid this, the gem recommends hooking into the `db:migrate` and `db:schema:load` rake tasks to also run the cron scheduling (it will ignore jobs already scheduled and ensure new jobs are scheduled).

Set to run only on hosted environments. I manually ran the rake task in the migration environment to ensure it works and to get the migration cron job scheduled.

### DIFF
--- a/lib/tasks/jobs.rake
+++ b/lib/tasks/jobs.rake
@@ -6,3 +6,12 @@ namespace :db do
     CronJob.subclasses.each(&:schedule)
   end
 end
+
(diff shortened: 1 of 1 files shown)


ethax-ross:
Closing as we want to play it safe for now/avoid this causing jobs to be queued in production as we're not that familiar with the NPQ reg scheduled jobs.

PR #1810, opened 2024-10-15
Add token generation service for parity check

ethax-ross:
[Jira-3614](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3614)

### Context

We are creating a service to check parity between the API responses of ECF and NPQ reg, once the data has been migrated. As part of this we will need to call the endpoints as lead providers, so we will need to be able to generate tokens for each lead provider in both ECF and NPQ reg.

- Add ECF models for ApiToken generation

Bring across the models from ECF that we need to be able to generate API tokens for ECF in NPQ registration.

I haven't improved the test coverage here, just a quick lift and shift from ECF.

- Add token generation service for ECF/NPQ

As part of the parity check we will need to generate API tokens for lead providers in ECF and NPQ reg on the fly to be able to make requests to the API as these providers.

Add service to generate tokens for each lead provider.

### DIFF
--- a/app/models/migration/ecf/api_token.rb
+++ b/app/models/migration/ecf/api_token.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "abstract_interface"
+
+module Migration::Ecf
+  class APIToken < BaseRecord
+    has_paper_trail ignore: %i[updated_at last_used_at]
+    # This is meant to be an abstract class
+    # Since it is a base class for a STI, we can't actually make it abstract (not backed by a table)
+    include AbstractInterface
+    implement_instance_method :owner
+
+    def self.create_with_random_token!(**options)
+      unhashed_token, hashed_token = Devise.token_generator.generate(APIToken, :hashed_token)
+      create!(hashed_token:, **options)
+      unhashed_token
+    end
+
+    def self.find_by_unhashed_token(unhashed_token)
+      hashed_token = Devise.token_generator.digest(APIToken, :hashed_token, unhashed_token)
+      find_by(hashed_token:)
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


leandroalemao (commented):
@ethax-ross Looks good 👍 but couldn't really test locally here as ECF models are [marked as read only](https://github.com/DFE-Digital/npq-registration/blob/main/app/models/migration/ecf/base_record.rb#L9).. so am getting an exception `ActiveRecord::ReadOnlyRecord` when running `Migration::ParityCheck::TokenProvider.new.generate!` 

Also Is it possible to add something to limit these things to run only in non-production envs? Also could you please have a look at test coverage on sonarcloud? 90.7% on the new code
Tksss

mooktakim (approved):

ethax-ross:
> @ethax-ross Looks good 👍 but couldn't really test locally here as ECF models are [marked as read only](https://github.com/DFE-Digital/npq-registration/blob/main/app/models/migration/ecf/base_record.rb#L9).. so am getting an exception `ActiveRecord::ReadOnlyRecord` when running `Migration::ParityCheck::TokenProvider.new.generate!`
> 
> Also Is it possible to add something to limit these things to run only in non-production envs? Also could you please have a look at test coverage on sonarcloud? 90.7% on the new code Tksss

Good catch! I've updated to be writable so it works in non-test environments and added a flag to the npq_separation config to raise if its not in the migration environment.

On the code coverage I put a note in my commit that I've just lifted & shifted the ECF code here and not improved the coverage; I didn't think it was worth the effort as its short-lived code that has been in ECF prod for ages. Happy to be challenged on that though if you feel differently?

leandroalemao (approved):
Looking nice now tks @ethax-ross 👍 😄

PR #1816, opened 2024-10-16
[CPDLP-3520] Disable declaration validation on schedule in migration only

leandroalemao:
### Context

Ticket: [CPDLP-3520](https://dfedigital.atlassian.net/browse/CPDLP-3520)

We have found old declarations that have invalid declaration dates in relation to the schedule start date. Those are legacy declarations in cohort 2021/2022 and will need to be taken away with contract managers, but we need to unblock migration so we should bring those across.

- Change the declaration migrator/model validation to allow pulling in declarations that have invalid schedules only. 
- Only allowed on data migration only and nowhere else.

[CPDLP-3520]: https://dfedigital.atlassian.net/browse/CPDLP-3520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/migration/migrators/declaration.rb
+++ b/app/services/migration/migrators/declaration.rb
@@ -22,7 +22,7 @@ module Migration::Migrators
       end
     end
 
-    def call
+    def call(skip_declaration_date_within_schedule_validation: true)
       migrate(self.class.ecf_declarations) do |ecf_declaration|
         declaration = ::Declaration.find_or_initialize_by(ecf_id: ecf_declaration.id)
         latest_declaration_state = ecf_declaration.declaration_states.detect(&:ineligible?)
@@ -37,6 +37,7 @@ module Migration::Migrators
           cohort_id: find_cohort_id!(ecf_id: ecf_declaration.cohort_id),
           lead_provider_id: find_lead_provider_id!(ecf_id: ecf_declaration.cpd_lead_provider.npq_lead_provider.id),
           application_id: find_application_id!(course_identifier: ecf_declaration.course_identifier, ecf_user_id: ecf_declaration.user_id),
+          skip_declaration_date_within_schedule_validation:,
         )
       end
     end
--- a/app/services/migration/migrators/declaration.rb
+++ b/app/services/migration/migrators/declaration.rb
@@ -22,7 +22,7 @@ module Migration::Migrators
       end
     end
 
-    def call(skip_declaration_date_within_schedule_validation: true)
+    def call
       migrate(self.class.ecf_declarations) do |ecf_declaration|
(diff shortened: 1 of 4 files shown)


app/services/migration/migrators/declaration.rb
```diff
@@ -37,6 +37,7 @@ def call
           cohort_id: find_cohort_id!(ecf_id: ecf_declaration.cohort_id),
           lead_provider_id: find_lead_provider_id!(ecf_id: ecf_declaration.cpd_lead_provider.npq_lead_provider.id),
           application_id: find_application_id!(course_identifier: ecf_declaration.course_identifier, ecf_user_id: ecf_declaration.user_id),
+          skip_declaration_date_within_schedule_validation:,
```
mooktakim, line 40: Instead of adding `attr_accessor` on `Declaration` model, as this is a "special case" I feel it might be better to only make change here on the migration by doing:
```ruby
        declaration.attributes = {
          created_at: ecf_declaration.created_at,
          updated_at: ecf_declaration.updated_at,
          declaration_type: ecf_declaration.declaration_type,
          declaration_date: ecf_declaration.declaration_date,
          state: ecf_declaration.state,
          state_reason: latest_declaration_state&.state_reason,
          cohort_id: find_cohort_id!(ecf_id: ecf_declaration.cohort_id),
          lead_provider_id: find_lead_provider_id!(ecf_id: ecf_declaration.cpd_lead_provider.npq_lead_provider.id),
          application_id: find_application_id!(course_identifier: ecf_declaration.course_identifier, ecf_user_id: ecf_declaration.user_id),
        }
        declaration.save!(validate: false)
```

What do you think?

ethax-ross: passing `validate: false` on the save is dangerous as it will skip _all_ validation; we still want failures if the declaration is invalid according to any other rules

ethax-ross (commented):
Looks good, just a minor question/suggestion. It's a shame we can't use validation contexts for this - there's not a clean way to exclude a validation rule with a context that I can see (we would have to update all other places which would be riskier).

app/services/migration/migrators/declaration.rb
cwrw: agree around wanting to keep the other validations and just ignoring this one

cwrw (commented):
looks good, thanks Leandro 👍

ethax-ross (approved):

cwrw (approved):

mooktakim (approved):

PR #1819, opened 2024-10-16
[CPDLP-3521] Split contract migration into two migrators

leandroalemao:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3521

We're seeing `ERROR:  cached plan must not change result type` when migrating some contracts over.

It's a concurrency issue, where a contract template is being created then a linked contract is being initialised multiple times as we have multiple workers trying to create the same contract. We're splitting contract migrator into two new migrators (contract template and contract). Also limiting the workers # for the contract migrator.

image

### DIFF
--- a/app/services/migration/migrators/contract.rb
+++ b/app/services/migration/migrators/contract.rb
@@ -2,20 +2,6 @@ module Migration::Migrators
   class Contract < Base
     INFRA_WORKER_COUNT = 1
 
-    SHARED_ATTRIBUTES = %w[
-      created_at
-      updated_at
-      service_fee_percentage
-      output_payment_percentage
-      per_participant
-      number_of_payment_periods
-      recruitment_target
-      service_fee_installments
-      targeted_delivery_funding_per_participant
-      monthly_service_fee
-      special_course
-    ].freeze
-
     class << self
       def record_count
         ecf_contracts.count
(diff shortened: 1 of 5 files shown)


cwrw (commented):
~I think we shouldn't be picking up the same records through multiple workers because of this~: https://github.com/DFE-Digital/npq-registration/blob/main/app/services/migration/migrators/base.rb#L60
~However the interesting thing is here:~ https://github.com/DFE-Digital/npq-registration/blob/main/app/services/migration/migrators/base.rb#L54 ~this might be adding the extra records you mentioned as it's doing a ceiling on both, maybe we can discuss this. Let's chat tomorrow with~ @ethax-ross. ~Nice catch, thanks~ @leandroalemao. 

Nvm I just reran and it raised the same failure with 1 worker, let's discuss tomorrow

app/services/migration/migrators/contract.rb
```diff
@@ -30,6 +32,18 @@ def ecf_contracts
       def dependencies
         %i[course statement]
       end
+
+      def number_of_workers
```
ethax-ross, line 36: Could we just wrap the contract creation in a transaction to avoid the race conditions? alternatively I think instead of overriding the constant you could just do something like this I think?:

```
def number_of_workers
  1
end

def records_per_worker
  record_count
end
```

leandroalemao: Hi @ethax-ross tried that yesterday.. but it breaks some shared specs..

leandroalemao:
Closing this for now. More info on https://dfedigital.atlassian.net/browse/CPDLP-3521.

PR #1822, opened 2024-10-16
[CPDNQT-2152] Resend outcome to api

jebw:
### Context

Ticket: [CPDNPQ-2152](https://dfedigital.atlassian.net/browse/CPDNPQ-2152)

We want to allow admin users to reschedule delivery of outcomes to the qualified teachers API

* Added logic for whether to allow rescheduling to the ParticipantOutcome model
* Added a method to `nil` out the relevant ParticipantOutcome attributes, which will trigger rescheduling
* Extended the ParticipantOutcomes component table to add a resend link
* Shortened some of the other columns in ParticipantOutcomes table to make space for the `Resend` link
* Added feature specs covering the above
* Refactored the existing method to determine whether an Outcome is the most recent to work with relationships instead of doing a query directly, this allows it to benefit from eager loaded data. The number of outcomes per declaration should be low so this shouldn't result in significant data being loaded for the `applications#show` screen.
* Extended the OutcomesTableComponent to include caption_text described the latest outcome
* Swapped the OutcomesTableComponent to store strings in translations yaml file

### Screenshots

image

[CPDNPQ-2144]: https://dfedigital.atlassian.net/browse/CPDNPQ-2144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CPDNPQ-2152]: https://dfedigital.atlassian.net/browse/CPDNPQ-2152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/npq_separation/admin/outcomes_table_component.rb
+++ b/app/components/npq_separation/admin/outcomes_table_component.rb
@@ -52,7 +52,9 @@ module NpqSeparation
       def tra_api_resend_link(outcome)
         return "" unless outcome.allow_resending_to_qualified_teachers_api?
 
-        govuk_link_to("Resend", resend_npq_separation_admin_participant_outcome_path(outcome))
+        govuk_button_to("Resend",
+                        resend_npq_separation_admin_participant_outcome_path(outcome),
+                        class: "govuk-!-margin-bottom-0")
       end
     end
   end
--- a/app/components/npq_separation/admin/outcomes_table_component.rb
+++ b/app/components/npq_separation/admin/outcomes_table_component.rb
@@ -4,11 +4,13 @@ module NpqSeparation
       attr_reader :outcomes
 
       def initialize(outcomes)
-        @outcomes = outcomes
+        @outcomes = outcomes.sort_by(&:created_at).reverse
       end
 
       def call
-        render GovukComponent::TableComponent.new(head:, rows:)
+        render GovukComponent::TableComponent.new(head:, rows:) do |table|
+          table.with_caption(size: "s", text: caption_text, classes: "govuk-heading-s")
+        end
       end
 
     private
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -234,6 +234,10 @@
         resources :courses, only: %i[index show]
         resources :users, only: %i[index show]
 
+        resources :participant_outcomes, only: %i[] do
+          member { get :resend }
```
jebw, line 238: @cwrw I've followed ECF and made this a `GET` but it feels like this should be a `POST` since its modifying data? Thoughts?

spec/models/participant_outcome_spec.rb
```diff
@@ -238,4 +238,71 @@
...
+      it { is_expected.to be true }
+    end
+
+    context "with incomplete delivery" do
+      let(:outcome) { create(:participant_outcome, :sent_to_qualified_teachers_api) }
+
+      it { is_expected.to be false }
+    end
+
+    context "with undelivered" do
```
cwrw, line 302: can we add a test that it resets the fields to `nil`?

spec/models/participant_outcome_spec.rb
```diff
@@ -238,4 +238,71 @@
...
+  describe "#resend_to_qualified_teachers_api!" do
+    subject { outcome.resend_to_qualified_teachers_api! }
+
+    context "with successfully delivery" do
+      let(:outcome) { create(:participant_outcome, :successfully_sent_to_qualified_teachers_api) }
+
+      it { is_expected.to be false }
+    end
+
+    context "with failed delivery" do
```
cwrw, line 290: can we add a test that it resets the fields to `nil`?

config/routes.rb
cwrw: yea `POST` sounds better tbh 👍

app/components/npq_separation/admin/outcomes_table_component.rb
```diff
@@ -19,7 +19,8 @@ def head
           "Completion date",
           "Submitted by provider",
           "Sent to TRA API",
-          "Recorded by TRA API",
+          "Recorded by API",
+          "",
```
cwrw, line 23: there's some `caption_text` logic in ECF, not sure if we want to bring across, don't think it's a must but just incase that was missed

cwrw (commented):
looks good just some minor comments

cwrw (approved):
🙌

PR #1832, opened 2024-10-18
[CPDLP-3632] Ensure school ukprn matches school urn in application migrator

leandroalemao:
### Context

Ticket: [CPDLP-3632](https://dfedigital.atlassian.net/browse/CPDLP-3632)

We are not checking if the school urn and school ukprn come from the same school in the migration.

Ensure that any value we pull in from ECF for school urn and school ukprn match the same school and should be aligned, otherwise raise a failure.

[CPDLP-3632]: https://dfedigital.atlassian.net/browse/CPDLP-3632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/migration/migrators/application.rb
+++ b/app/services/migration/migrators/application.rb
@@ -72,8 +72,10 @@ module Migration::Migrators
         if ecf_npq_application.school_urn.present?
           application.school_id = find_school_id!(urn: ecf_npq_application.school_urn)
 
-          if application.school.ukprn.to_s == ecf_npq_application.school_ukprn.to_s
+          if ecf_npq_application.school_ukprn && ecf_npq_application.school_ukprn.to_s == application.school.ukprn.to_s
             application.ukprn = ecf_npq_application.school_ukprn
+          elsif ecf_npq_application.school_ukprn.blank?
+            application.ukprn = application.school.ukprn
           else
             ecf_npq_application.errors.add(:base, "School UKPRN does not match")
             raise ActiveRecord::RecordInvalid, ecf_npq_application
--- a/app/services/migration/migrators/application.rb
+++ b/app/services/migration/migrators/application.rb
@@ -71,6 +71,13 @@ module Migration::Migrators
 
         if ecf_npq_application.school_urn.present?
           application.school_id = find_school_id!(urn: ecf_npq_application.school_urn)
+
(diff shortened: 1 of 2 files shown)


app/services/migration/migrators/application.rb
```diff
@@ -71,6 +71,13 @@ def call
 
         if ecf_npq_application.school_urn.present?
           application.school_id = find_school_id!(urn: ecf_npq_application.school_urn)
+
+          if application.school.ukprn.to_s == ecf_npq_application.school_ukprn.to_s
+            application.ukprn = ecf_npq_application.school_ukprn
```
cwrw, line 76: since we're setting this here can we delete this line https://github.com/DFE-Digital/npq-registration/pull/1832/files#diff-a44bc9ad55c1f2d7565d5a1a513776fb08f5180b4934350ee83fba59bad1d52cR94

cwrw (approved):
one minor comments, let's run this after in the migrator :)

app/services/migration/migrators/application.rb
```diff
@@ -71,6 +71,13 @@ def call
 
         if ecf_npq_application.school_urn.present?
           application.school_id = find_school_id!(urn: ecf_npq_application.school_urn)
+
+          if application.school.ukprn.to_s == ecf_npq_application.school_ukprn.to_s
```
ethax-ross, line 75: Would this check not make more sense in the schools migrator? otherwise we'll end up with multiple errors for the same school UKPRN mismatch (one for every application at the school). Whereas if we did it in the schools migrator it would fail once for the school it was wrong for? something like:

```
migrate(self.class.ecf_schools) do |ecf_school|
  school = find_school_id!(urn: ecf_school.urn, name: ecf_school.name.downcase)
  raise if school.ukprn != ecf_school.ukprn
end
```

mooktakim: This is testing if `ecf_npq_application.school_urn` and `ecf_npq_application.school_ukprn` both point to the same school. Therefore the application record has a mismatch.

mooktakim (approved):

PR #1836, opened 2024-10-21
[CPDLP-3648] Data cleanup - Find user by email

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3648

* `Migration::Users::Creator` can now deal with email address that already exists in NPQ


app/services/migration/migrators/user.rb
```diff
@@ -32,17 +32,18 @@ def records_per_worker
 
     def call
       migrate(self.class.ecf_users) do |ecf_user|
-        user = ::Migration::Users::Creator.new(ecf_user).find_or_initialize
+        npq_application = most_recent_created_npq_application(ecf_user)
+        email = npq_application&.participant_identity&.email
+        email = email.downcase if email.present?
```
cwrw, line 37: curious when there would be no email 🤔 maybe we need to fallback to profiles if there are no applications incase to avoid a case where the email is `nil`?

cwrw (commented):
looks good past the one minor comment 🙌

app/services/migration/migrators/user.rb
mooktakim: No email means there's no npq_application, which wouldn't be included in the migration.
But either way, the `ecf_id` and `gai_id` would resolve to a user.

app/services/migration/migrators/user.rb
```diff
@@ -32,17 +32,18 @@ def records_per_worker
 
     def call
       migrate(self.class.ecf_users) do |ecf_user|
-        user = ::Migration::Users::Creator.new(ecf_user).find_or_initialize
+        npq_application = most_recent_created_npq_application(ecf_user)
+        email = npq_application&.participant_identity&.email
```
ethax-ross, line 36: Could just add to the chain here

```
email = npq_application&.participant_identity&.email&.downcase
```

app/services/migration/users/creator.rb
```diff
@@ -1,17 +1,54 @@
...
 
       def find_or_initialize
+        primary_user = find_primary_user
+
+        # User doesn't exist at all
+        if primary_user.nil? && user_by_ecf_user_email.nil?
+          return ::User.new(ecf_id: ecf_user.id)
+        end
+
+        # Primary user found and `ecf_user_email` if blank or same as primary user
```
ethax-ross, line 19: `is blank or`

ethax-ross (commented):
Looks good, just couple minor suggestions

app/services/migration/migrators/user.rb
cwrw: the thing is we have cases where the user is only connected the profile not the application so this would be blank, but that is just bad data we should resolve

cwrw (approved):

ethax-ross (approved):

leandroalemao (approved):
Looking good tks @mooktakim 👍 😄

PR #1840, opened 2024-10-22
Support pagination and filtering in parity check service

ethax-ross:
[Jira-3618](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3618)

### Context

We want to support pagination and filtering in the parity check service, so that we can compare the responses across multiple pages and with different query parameters.

- Add page attribute to ResponseComparison

We want to support pagination when calling certain endpoints as party of the parity check. To make it more managable, we're going to store the results of each page as a separate `ResponseComparison` instance.

Add `page` attribute to `ResponseComparison` model.

- Extract Client from ParityCheck service

To avoid overcomplicating the `ParityCheck` service, the request/response logic has been extracted to a separate `Client` model. This model will be expanded to support pagination and any other request/response behaviour.

- Support pagination

Update the `Client` to support paginating if the `options[:paginate]` us specified.

It will paginate until there are no pages left for either NPQ or ECF (determined by the resulting `data` count being less than the `per_page`).

- Update parity check UI to support pagination

Display comparison results grouped by description on the completion page (so comparisons with the same method/path and different page will be grouped together).

On the completion page we show one entry for paginated comparisons (if there is one difference across any page then its marked as different).

On the detail page we only show the different pages in the result, along with a summary of the average performance/response times.

The per page value is intentionally set as a low value for now so that we can see the pagination work on statements.

### Guidance for review

I've made the per-page low for now to demonstrate the UI, but we can increase it when we get to the endpoints with more records.

Best reviewed by-commit.

| Complete | Complete (expanded) | Multi page response | Multi page response (expanded) | Single response |
| -- | -- | -- | -- | -- |
| screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-2024-10-22-10_22_08 | screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-2024-10-22-10_22_33 | screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-response-comparisons-426-2024-10-22-10_22_57 |  screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-response-comparisons-426-2024-10-22-10_24_10 | screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-response-comparisons-431-2024-10-22-10_23_37 |

### DIFF
--- a/app/controllers/npq_separation/migration/parity_checks_controller.rb
+++ b/app/controllers/npq_separation/migration/parity_checks_controller.rb
@@ -16,7 +16,7 @@ class NpqSeparation::Migration::ParityChecksController < SuperAdminController
   end
 
   def response_comparison
-    @comparison = Migration::ParityCheck::ResponseComparison.with_differences.find(params[:id])
-    @response_body_diff = @comparison.response_body_diff
+    @comparison = Migration::ParityCheck::ResponseComparison.find(params[:id])
+    @matching_comparisons = Migration::ParityCheck::ResponseComparison.matching(@comparison)
   end
 end
--- a/app/helpers/migration_helper.rb
+++ b/app/helpers/migration_helper.rb
@@ -79,27 +79,44 @@ module MigrationHelper
     end
   end
 
-  def response_comparison_performance(comparison)
-    performance = (comparison.ecf_response_time_ms.to_f / comparison.npq_response_time_ms).round(1).to_s.chomp(".0")
+  def response_comparison_performance(comparisons)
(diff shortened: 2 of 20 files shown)


mooktakim (approved):

leandroalemao (approved):
Looks very good 👍 😄

cwrw (approved):
great work 🙌

PR #1847, opened 2024-10-23
Add remaining GET endpoints to parity check

ethax-ross:
[Jira-3617](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3617)
[Jira-3619](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3619)

### Context

We want to expand the parity check to include all other GET endpoints so that we can compare the results. This also involves supporting individual GET resource endpoints, which require a different `id` in the path for each provider.

- Optimise the ResponseComparison deletion on re-run

Deleting the existing parity check results can be slow as they can be very large records in the database. We don't care about callbacks or validation here, so we can optimise the deletion with `delete_all`.

- Support checking an individual resource

We need to be able to request an individual resource as part of the parity check. The difficulty with this is that we need a valid ID for each lead provider. Instead of specifying these manually, I've opted to allow dynamic substitution by evaluating code from the YAML file.

We can now specify an ID attribute in the path with `:id` and a corresponding option in the YAML file that, when evaluated (in the context of the `Client`) should return a valid id for the lead provider.

- Add remaining GET endpoints to YAML file

Add all the listing and individual GET resource endpoints to the YAML file.

- Convert CSV responses to hexdigest

The CSV responses are very large and cause issues when trying to insert into Postgres. To be able to include them for now we are converting the response CSV to a hexdigest that can be stored and compared. If we want more details we will have to make the requests manually to inspect the responses.

- Stop paginating on first match

In order to finish the parity check as soon as possible, we stop paginating responses as soon as we encounter a discrepency. A mismatch in a earlier page is likely to effect later pages anyway, so this is all we will be interested in (and other issues would be surfaced on future runs when its fixed).

- Remove CSV endpoints

These cause issues with timeouts on the database; I'm not sure why as we hash the large CSV response so it shouldn't cause issues.

Remove CSV endpoints; if we want to include them we can add another ticket to make them work.

- Persist formatted path to ResponseComparison

As the id we grab for the lead provider is random, we need to persist it in the `ResponseComparison` to be able to replicate the request.

Add `id` to `request_path` before saving (avoids adding a new attribute to `ResponseComparison`).

- Pretty format JSON in diff

When we output the JSON diff we want to pretty format it or we end up with a single line which is hard to compare.

- Deep sort responses

If the response is JSON we deep sort the resulting hash so that we can compare the contents irrespective of order (as the new serializers in NPQ reg may serialize the fields in a different order).

### Guidance for review

Best reviewed by-commit.

I left the CSV hexdigest commit in as we will still need it if we bring the CSV endpoints in again (we return all results, which would be a really big diff/thing to put into the database without hashing it).

| Results | Results (expanded) | Updated diff |
| -- | -- | -- |
| screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-2024-10-23-13_22_16 | screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-2024-10-23-13_22_36 | screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-response-comparisons-15-2024-10-23-16_14_57 |


config/parity_check_endpoints.yml
```diff
@@ -1,9 +1,74 @@
...
+
+  "/api/v1/participants/npq":
+    paginate: true
+  "/api/v2/participants/npq":
+    paginate: true
+  "/api/v3/participants/npq":
+    paginate: true
+
+  "/api/v1/participants/npq/:id":
+    id: 'User.includes(:applications).where(applications: { lead_provider: }).order("RANDOM()").limit(1).pick(:ecf_id)'
```
cwrw, line 52: can we make sure those are `accepted` applications?

app/services/migration/parity_check/client.rb
```diff
@@ -87,7 +91,11 @@ def url(app:)
     def formatted_path
       return path unless options[:id] && path.include?(":id")
 
-      path.sub(":id", eval(options[:id]).to_s) # rubocop:disable Security/Eval
+      id = eval(options[:id]) # rubocop:disable Security/Eval
```
cwrw: curious about this, why wouldn't we able to find an id 🤔 ideally all should be there?

config/parity_check_endpoints.yml
```diff
@@ -1,7 +1,66 @@
 get:
+  "/api/v1/npq-applications":
```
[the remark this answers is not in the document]
ethax-ross: I think we were going to pick up filtering in a future ticket; in the AC 😅 :

```
- Do not add filters yet
```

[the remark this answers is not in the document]
cwrw: lol ignore me then 😂

app/services/migration/parity_check/client.rb
```diff
@@ -70,8 +80,16 @@ def headers
...
-      Rails.application.config.npq_separation[:parity_check]["#{app}_url".to_sym] + path
+    def url(app:)
+      Rails.application.config.npq_separation[:parity_check]["#{app}_url".to_sym] + formatted_path
+    end
+
+    def formatted_path
+      @formatted_path ||= begin
+        return path unless options[:id] && path.include?(":id")
+
+        path.sub(":id", eval(options[:id]).to_s) # rubocop:disable Security/Eval
```
mooktakim, line 91: Instead of doing eval from the yaml file content, can we pass in instructions instead?
eg:
```yaml
  "/api/v1/npq-applications/:id":
    id: 'random_lead_provider_ecf_id`
```

And then in `Client` have method to translate to action:
```ruby
def process_instruct(n)
  case n
  when "random_lead_provider_ecf_id"
    lead_provider.applications.order("RANDOM()").limit(1).pick(:ecf_id)'
  end
end

ecf_id = process_instruct(options[:id])
```

config/parity_check_endpoints.yml
```diff
@@ -1,7 +1,66 @@
...
+    paginate: true
+  "/api/v2/npq-applications":
+    paginate: true
+  "/api/v3/npq-applications":
+    paginate: true
+
+  "/api/v1/npq-applications/:id":
+    id: 'lead_provider.applications.order("RANDOM()").limit(1).pick(:ecf_id)'
```
mooktakim, line 10: possible feature, option to run the single request multiple times? eg: "run_count: 100"

ethax-ross: Yeah one for the future maybe as this would have knock on effects to how we display/store the results

app/services/migration/parity_check/client.rb
cwrw: hmm somehow prefer it in the yaml file 🤔 otherwise we'll have a huge list in the client

ethax-ross: I prefer it out of the YAML file; easier to test, avoids `eval` and less duplication

config/parity_check_endpoints.yml
mooktakim: Hmm, i'm thinking we could do it now for free by duplicating the same URL like this:
- `/api/v1/npq-applications/:id#1`
- `/api/v1/npq-applications/:id#2`
- `/api/v1/npq-applications/:id#3`

lol

mooktakim (approved):

PR #1848, opened 2024-10-23
[CPDLP-3631] Ensure we migrate all relevant timestamps from ECF

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3631

* `Cohort`, `Schedule`, `ParticipantIdChange` and `Statement` migrations will use timestamps from ECF
* Updated `paper_trail`
  * `object` field type is now `json`
  * `object_changes` added
  * added `note` field
* `Application` & `User`
  * Can now include `version_note` which adds `versions.note` entry into `versions`
  * Migration will use ECF timestamps and include a `version_note`
  * If there has been any changes from ECF attrs, we set `updated_at` to `Time.zone.now`

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -11,7 +11,7 @@ class Application < ApplicationRecord
     establishment-ineligible
   ].freeze
 
-  has_paper_trail only: %i[lead_provider_approval_status participant_outcome_state]
+  has_paper_trail meta: { note: :version_note }
 
   belongs_to :user
   belongs_to :course
@@ -33,6 +33,8 @@ class Application < ApplicationRecord
   scope :accepted, -> { where(lead_provider_approval_status: "accepted") }
   scope :eligible_for_funding, -> { where(eligible_for_funding: true) }
   scope :with_targeted_delivery_funding_eligibility, -> { where(targeted_delivery_funding_eligibility: true) }
+
+  attr_accessor :version_note
 
   validate :schedule_cohort_matches
   # TODO: add constraints into the DB after separation
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,8 @@
 class User < ApplicationRecord
   devise :omniauthable, omniauth_providers: [:tra_openid_connect]
 
+  has_paper_trail meta: { note: :version_note }
+
   has_many :applications, dependent: :destroy
   has_many :declarations, through: :applications
   has_many :ecf_sync_request_logs, as: :syncable, dependent: :destroy
@@ -37,6 +39,8 @@ class User < ApplicationRecord
   EMAIL_UPDATES_ALL_STATES = [:empty] + EMAIL_UPDATES_STATES
 
   enum email_updates_status: EMAIL_UPDATES_ALL_STATES, _suffix: true
+
+  attr_accessor :version_note
 
   def latest_participant_outcome(lead_provider, course_identifier)
     declarations.eligible_for_outcomes(lead_provider, course_identifier)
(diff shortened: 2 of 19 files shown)


db/migrate/20241023112339_change_object_type_to_json_for_versions.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeObjectTypeToJsonForVersions < ActiveRecord::Migration[7.1]
+  def up
+    change_column :versions, :object, :json, using: "object::text::json"
```
ethax-ross, line 3: Are all existing version `object` values valid JSON?

ethax-ross (commented):
Looks great just one question

db/migrate/20241023112339_change_object_type_to_json_for_versions.rb
mooktakim: I've checked all the records in prod, they are valid JSON. (non-nil ones)

ethax-ross (approved):

db/migrate/20241023112339_change_object_type_to_json_for_versions.rb
cwrw: can we check on all envs please?

app/models/application.rb
```diff
@@ -11,7 +11,7 @@ class Application < ApplicationRecord
     establishment-ineligible
   ].freeze
 
-  has_paper_trail only: %i[lead_provider_approval_status participant_outcome_state]
+  has_paper_trail meta: { note: :version_note }
```
cwrw, line 14: could we add a comment in both application/user model to explain that the created at isn't the original please?

app/services/migration/migrators/application.rb
```diff
@@ -25,6 +25,8 @@ class Application < Base
       teacher_catchment_country
       notes
       funded_place
+      created_at
```
cwrw, line 28: Here and in user, are we making sure this is covered?
> Overwrite updated_at with the one from ECF, or set to the date of migration if anything changed from ECF

app/models/application.rb
mooktakim: @cwrw the version that would get created during migration will include the note here:
https://github.com/DFE-Digital/npq-registration/pull/1848/files#diff-a44bc9ad55c1f2d7565d5a1a513776fb08f5180b4934350ee83fba59bad1d52cR95

So all the changes from the original will be shown. Do you think we need to be more specific?

cwrw: discussed in person: let's just leave in papertrail

app/services/migration/migrators/application.rb
mooktakim: `updated_at` will be set to Time.now when:
- Application `ukprn` is changed from ECF
- User `trn` and `email` is changed from ECF

cwrw (approved):
🙌 thanks mook

leandroalemao (approved):
Looking very good tks @mooktakim 👍

app/models/user.rb
```diff
@@ -1,6 +1,8 @@
 class User < ApplicationRecord
   devise :omniauthable, omniauth_providers: [:tra_openid_connect]
 
+  has_paper_trail meta: { note: :version_note }
```
cwrw, line 4:
```suggestion
  has_paper_trail meta: { note: :version_note }, ignore: [:raw_tra_provider_data, :provider, :feature_flag_id, :get_an_identity_id_synced_to_ecf, :updated_from_tra_at, :trn_lookup_status, :notify_user_for_future_reg, :email_updates_status, :email_updates_unsubscribe_key]
```

cwrw (approved):

PR #1850, opened 2024-10-23
[CPDLP-3652] Add unique index to all models with ecf id in NPQ

leandroalemao:
### Context

Ticket: [CPDLP-3652](https://dfedigital.atlassian.net/browse/CPDLP-3652)

We currently do not have a unique index on models with `ecf_id` in NPQ. We will need to add those except for `Users` which needs fixing and is being done by the NPQ team.

- Add a unique index on `ecf_id` all models with an `ecf_id` in NPQ (except for User) as well as a DB constraint;
- Ensure there is no data in all other envs: staging/sandbox/separation/production which will break this index, and fix the data;
- Ensure the indices added will run quickly and not block deployments;

Tested in snapshot db:

```
== 20241022204916 AddUniqueConstraintToEcfIdInApplications: migrating =========
-- remove_index(:applications, :ecf_id)
   -> 0.2731s
-- add_index(:applications, :ecf_id, {:unique=>true, :algorithm=>:concurrently})
   -> 0.4602s
== 20241022204916 AddUniqueConstraintToEcfIdInApplications: migrated (0.7378s) 

== 20241023093413 AddUniqueConstraintToEcfIdInContractTemplates: migrating ====
-- remove_index(:contract_templates, :ecf_id)
   -> 0.0660s
-- add_index(:contract_templates, :ecf_id, {:unique=>true, :algorithm=>:concurrently})
   -> 0.0399s
== 20241023093413 AddUniqueConstraintToEcfIdInContractTemplates: migrated (0.1065s) 

== 20241023093913 AddUniqueConstraintToEcfIdInCourses: migrating ==============
-- add_index(:courses, :ecf_id, {:unique=>true, :algorithm=>:concurrently})
   -> 0.0331s
== 20241023093913 AddUniqueConstraintToEcfIdInCourses: migrated (0.0334s) =====

== 20241023125017 AddUniqueConstraintToEcfIdInDeclarations: migrating =========
-- remove_index(:declarations, :ecf_id)
   -> 0.1625s
-- add_index(:declarations, :ecf_id, {:unique=>true, :algorithm=>:concurrently})
   -> 0.0510s
== 20241023125017 AddUniqueConstraintToEcfIdInDeclarations: migrated (0.2171s) 

== 20241023141116 AddUniqueConstraintToEcfIdInLeadProviders: migrating ========
-- add_index(:lead_providers, :ecf_id, {:unique=>true, :algorithm=>:concurrently})
   -> 0.0357s
== 20241023141116 AddUniqueConstraintToEcfIdInLeadProviders: migrated (0.0362s) 

== 20241023141630 AddUniqueConstraintToEcfIdInParticipantIdChanges: migrating =
-- remove_index(:participant_id_changes, :ecf_id)
   -> 0.1510s
-- add_index(:participant_id_changes, :ecf_id, {:unique=>true, :algorithm=>:concurrently})
   -> 0.0367s
== 20241023141630 AddUniqueConstraintToEcfIdInParticipantIdChanges: migrated (0.1898s)
```

Tested in the remaining envs:

```
=> [Application, ContractTemplate, Course, Declaration, LeadProvider, ParticipantIdChange].map do |klass|
  klass.where.not(ecf_id: nil).select(:ecf_id).group(:ecf_id).having("count(*) > 1").size
end
=> [{}, {}, {}, {}, {}, {}]
```
[CPDLP-3652]: https://dfedigital.atlassian.net/browse/CPDLP-3652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/application_state.rb
+++ b/app/models/application_state.rb
@@ -4,7 +4,6 @@ class ApplicationState < ApplicationRecord
   belongs_to :application
   belongs_to :lead_provider, optional: true
 
-  # TODO: remove "allow_nil" and add default value "gen_random_uuid()" and constraints into the DB after separation
   validates :ecf_id, uniqueness: { case_sensitive: false }, allow_nil: true
 
   enum state: {
--- a/app/models/application_state.rb
+++ b/app/models/application_state.rb
@@ -4,6 +4,9 @@ class ApplicationState < ApplicationRecord
   belongs_to :application
   belongs_to :lead_provider, optional: true
 
+  # TODO: remove "allow_nil" and add default value "gen_random_uuid()" and constraints into the DB after separation
+  validates :ecf_id, uniqueness: { case_sensitive: false }, allow_nil: true
+
   enum state: {
     active: "active",
     deferred: "deferred",
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -15,6 +15,9 @@ class Cohort < ApplicationRecord
               message: "Choose true or false for funding cap",
             }
 
+  # TODO: remove "allow_nil" and add default value "gen_random_uuid()" and constraints into the DB after separation
+  validates :ecf_id, uniqueness: { case_sensitive: false }, allow_nil: true
+
   def registration_start_date_matches_start_year
     return if registration_start_date.blank?
 
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -14,8 +14,6 @@ class Cohort < ApplicationRecord
               in: [true, false],
               message: "Choose true or false for funding cap",
             }
-
-  # TODO: remove "allow_nil" and add default value "gen_random_uuid()" and constraints into the DB after separation
   validates :ecf_id, uniqueness: { case_sensitive: false }, allow_nil: true
 
   def registration_start_date_matches_start_year
--- a/app/models/participant_outcome.rb
+++ b/app/models/participant_outcome.rb
@@ -2,7 +2,7 @@ class ParticipantOutcome < ApplicationRecord
   belongs_to :declaration
   has_many :participant_outcome_api_requests
 
-  validates :ecf_id, uniqueness: true
+  validates :ecf_id, uniqueness: { case_sensitive: false }
   validates :state, presence: true
   validates :completion_date, presence: true
   validate :completion_date_not_in_the_future
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -91,6 +91,7 @@ RSpec.describe Application do
 
       # TODO: uncomment this when `before_validation` is removed from model, as `before_validation` is adding ecf_id regardless
       # it { is_expected.to validate_presence_of(:ecf_id).with_message("Enter an ECF ID") }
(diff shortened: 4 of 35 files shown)


cwrw (commented):
could we check data on all envs if they pass, except production as you already tested :)

spec/models/application_spec.rb
```diff
@@ -19,6 +19,8 @@
   end
 
   describe "validations" do
+    it { is_expected.to validate_uniqueness_of(:ecf_id).case_insensitive.with_message("ECF ID must be unique").allow_nil }
```
cwrw, line 22: can we add a test for the feature flag as well?

app/models/application_state.rb
```diff
@@ -4,6 +4,9 @@ class ApplicationState < ApplicationRecord
   belongs_to :application
   belongs_to :lead_provider, optional: true
 
+  # TODO: remove "allow_nil" and add default value "gen_random_uuid()" and constraints into the DB after separation
```
cwrw, line 7: this isn't surface in the API, so can allow it to remain `nil` after separation and delete that field as well

app/models/cohort.rb
```diff
@@ -15,6 +15,9 @@ class Cohort < ApplicationRecord
               message: "Choose true or false for funding cap",
             }
 
+  # TODO: remove "allow_nil" and add default value "gen_random_uuid()" and constraints into the DB after separation
```
cwrw, line 18: same comment as above, we left those intentionally not auto generated in the db: the ones that aren't surfaced in the API

app/models/participant_outcome.rb
```diff
@@ -2,7 +2,7 @@ class ParticipantOutcome < ApplicationRecord
   belongs_to :declaration
   has_many :participant_outcome_api_requests
 
-  validates :ecf_id, uniqueness: true
+  validates :ecf_id, uniqueness: { case_sensitive: false }
```
TBenedyk, line 5: Why are we making all these case insensitive?

leandroalemao: Unique constraints are `case insensitive` or `case_sensitive false` by default so the DB won't allow adding a value for ecf_id = "583162EE-00B8-4AB9-A668-996E8EAAD40B" if "583162ee-00b8-4ab9-a668-996e8eaad40b" is already there. 

Shoulda matches in the other hand[ is `case sensitive true` by default](https://github.com/thoughtbot/shoulda-matchers/blob/main/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L190), so we need to add a `.case_insensitive` to the matcher, so the test passes.

`it { is_expected.to validate_uniqueness_of(:ecf_id).case_insensitive.with_message("ECF ID must be unique") }`

As we had to add `case_insensitive` to the spec, I'm adding `{ case_sensitive: false }` into the model as well to keep it consistent with the rest.

mooktakim (approved):

ethax-ross (approved):

PR #1853, opened 2024-10-24
Exclude known differences from parity check

ethax-ross:
### Context

[Jira-3651](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3651)

- Exclude ECF-only attributes from responses

In some endpoints ECF will return additional, ECF-specific attributes. We want
to exclude these from the comparison so they don't effect the diff/equality
check.

Add `exclude` transient attribute to `ResponseComparison` and deep-remove
matching keys from the response payload prior to persisting.

- Filter statements to only NPQ

Add `filter[type]=npq` to the statements request to ensure we are only comparing NPQ statements from ECF.

Remove other statement endpoints for now; we will re-introduce filtered endpoints in a later PR.

- Filter declarations to only NPQ

Add `filter[type]=npq` to the declarations request to ensure we are only comparing NPQ statements from ECF (this is supported on the ECF migration environment only).

### Guidance for review

screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-2024-10-24-12_20_56

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Disable in migration environment
+Rack::Attack.enabled = !Rails.env.migration?
+
 # Throttle general requests by IP
 class Rack::Attack
   PROTECTED_ROUTES = [


cwrw (approved):

mooktakim (approved):

leandroalemao (approved):
Looking good tks @ethax-ross 👍 😄

PR #1859, opened 2024-10-25
[CPDLP-3638] Add backfill for ecf_ids and data in migration

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3638

* `Application`
  * Backfill for `ecf_id`
  * Backfill for `cohort_id`
  * Backfill for `lead_provider_approval_status `
* `User`
  * Backfill for `ecf_id`

### DIFF
--- a/app/services/migration/migrators/application.rb
+++ b/app/services/migration/migrators/application.rb
@@ -96,6 +96,10 @@ module Migration::Migrators
         application.version_note = "Changes migrated from ECF to NPQ"
         application.update!(ecf_npq_application.attributes.slice(*ATTRIBUTES))
       end
+
+      run_once { backfill_ecf_ids }
+      run_once { backfill_cohorts }
+      run_once { backfill_lead_provider_approval_statuses }
     end
 
   private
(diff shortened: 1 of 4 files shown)


spec/services/migration/migrators/application_spec.rb
```diff
@@ -219,6 +219,24 @@ def setup_failure_state
...
+      context "when backfilling existing applications" do
+        it "sets `ecf_id`" do
+          application = create(:application, ecf_id: nil)
+
+          instance.call
+
+          expect(application.reload.ecf_id).to be_present
+        end
+
+        it "sets `cohort_id` to current cohort when schedule is not set" do
```
cwrw, line 232: could we add more tests for schedule and lead provider status?

app/services/migration/migrators/application.rb
```diff
@@ -96,6 +96,10 @@ def call
         application.version_note = "Changes migrated from ECF to NPQ"
         application.update!(ecf_npq_application.attributes.slice(*ATTRIBUTES))
       end
+
+      run_once { backfill_ecf_ids }
```
cwrw, line 100: not sure how this will work with the run once logic, we'll find out on migration, and might need to have all in one `run_once` block, worth giving it a whirl and we'll discuss after :)

leandroalemao (approved):
Looks good. It'd be nice to run a test in migration env to see the results... tks 👍

app/services/migration/migrators/application.rb
mooktakim: migration run completed successfully.  👍

mooktakim:
Before run:
```
puts Application.where(ecf_id: nil).count
# 0
puts Application.where(cohort_id: nil).count
# 17
puts Application.where(lead_provider_approval_status: nil).count
# 1
puts User.joins(:applications).where(ecf_id: nil).distinct.count
# 0
```

After run:
```
puts Application.where(ecf_id: nil).count
# 0
puts Application.where(cohort_id: nil).count
# 0
puts Application.where(lead_provider_approval_status: nil).count
# 0
puts User.joins(:applications).where(ecf_id: nil).distinct.count
# 0
```

spec/services/migration/migrators/application_spec.rb
ethax-ross: I think we're still missing the case where the schedule is set? (sorry if you know this/just haven't got around to it yet)

ethax-ross (commented):
Looks good past the schedule fallback test from Abeer's comment

spec/services/migration/migrators/application_spec.rb
mooktakim: There's a validation that prevents `application.cohort` being different from `application.schedule.cohort`
Can't actually create that scenario.
https://github.com/DFE-Digital/npq-registration/blob/main/app/models/application.rb#L208

cwrw (approved):

ethax-ross (approved):

PR #1861, opened 2024-10-25
Update konduit make commands to point to correct web app

cwrw:
### Context

We are pointing to cpd ecf when we should npq registration in konduit make command

Amend make command to point to npq registration web app instead.

copy paste innit, thanks @jebw for the good spot!


jebw (approved):
LGTM

PR #1871, opened 2024-10-28
[CPDLP-3635] Swap out NPQ api config for feature flags

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3635

* Replaced `admin_portal_enabled` , `api_enabled`, `ecf_api_disabled` and `migration_enabled` to the feature flag `Feature.ecf_api_disabled?`
* Added runner for refresh migration action so it enables the feature for the `migration` env.

### DIFF
--- a/.github/workflows/refresh_migration_database.yml
+++ b/.github/workflows/refresh_migration_database.yml
@@ -45,5 +45,5 @@ jobs:
       - name: Generate known keys
         shell: bash
         run: |
-          kubectl -n cpd-production exec -ti --tty deployment/npq-registration-migration-web -- /bin/sh -c "cd /app && bundle exec rails runner \"Migration::ParityCheck::TokenProvider.new.generate!\""
-          
+          kubectl -n cpd-production exec -ti --tty deployment/npq-registration-migration-web -- /bin/sh -c "cd /app && bundle exec rails runner 'Migration::ParityCheck::TokenProvider.new.generate!' && bundle exec rails runner 'Feature.enable_ecf_api_disabled'"
+
--- a/.github/workflows/refresh_migration_database.yml
+++ b/.github/workflows/refresh_migration_database.yml
@@ -45,5 +45,9 @@ jobs:
       - name: Generate known keys
         shell: bash
         run: |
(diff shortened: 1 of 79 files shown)


leandroalemao (commented):
Quick question.. we're merging those configs into one feature flag. Does that mean for example that the migrations will be available to be run in the separation env?

.github/workflows/refresh_migration_database.yml
```diff
@@ -45,5 +45,5 @@ jobs:
       - name: Generate known keys
         shell: bash
         run: |
-          kubectl -n cpd-production exec -ti --tty deployment/npq-registration-migration-web -- /bin/sh -c "cd /app && bundle exec rails runner \"Migration::ParityCheck::TokenProvider.new.generate!\""
-          
\ No newline at end of file
+          kubectl -n cpd-production exec -ti --tty deployment/npq-registration-migration-web -- /bin/sh -c "cd /app && bundle exec rails runner 'Migration::ParityCheck::TokenProvider.new.generate!' && bundle exec rails runner 'Feature.enable_ecf_api_disabled'"
```
cwrw, line 48: could we add an extra step for enabling the feature flag? as this step's name is generate known keys?

app/services/feature.rb
```diff
@@ -58,5 +58,9 @@ def enable_registration!
     def ecf_api_disabled?
       Flipper.enabled?(ECF_API_DISABLED)
     end
+
+    def enable_ecf_api_disabled
```
cwrw, line 62:
```suggestion
    def enable_ecf_api_disabled!
```

ethax-ross (commented):
I think this should all be fine; like Leandro pointed out some stuff will be enabled where it wasn't before, but the migration aspect is manually triggered and the secrets would need to be in place for the ECF database in an environment for it to work. 

I do find `ecf_api_disabled` an awkward flag though; these double negatives hurt my head 😞 :

```
def enable_ecf_api_disabled!
  Flipper.enable(ECF_API_DISABLED)
end
```

Would it be clearer to have the feature as something like `lead_provider_api` then you would have `enable_lead_provider_api!`?

leandroalemao (commented):
How difficult is it to remove the `parity_check` from configs as well?

cwrw:
> How difficult is it to remove the parity_check from configs as well?

We agreed parity checks should remain there in the app config as we don't want that anywhere except on migration. There are some actions as well we don't want anyone running anywhere ever :)

leandroalemao (approved):
👍 All good tks @mooktakim

cwrw (approved):

ethax-ross (approved):

PR #1872, opened 2024-10-28
[CPDLP-3674] Fix attributes slice syntax

leandroalemao:
### Context

Ticket: [CPDLP-3674](https://dfedigital.atlassian.net/browse/CPDLP-3674)

The declarations `updated_at` timestamps is being changed in migrations when it shouldn’t be.

The `attributes` method returns a hash which keys are string, not symbol.. so it was not setting `statement_items` timestamps correctly in the statement item migrator, causing the [declarations serializer updated_at](https://github.com/DFE-Digital/npq-registration/blob/main/app/serializers/api/declaration_serializer.rb#L41) to return the wrong value.

```
User.last.attributes.slice(:id) => {}
User.last.attributes.slice("id") => {"id"=>216}
```

[CPDLP-3674]: https://dfedigital.atlassian.net/browse/CPDLP-3674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/migration/migrators/course_spec.rb
+++ b/spec/services/migration/migrators/course_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe Migration::Migrators::Course do
       it "sets the created Course attributes correctly" do
         instance.call
         course = Course.find_by(ecf_id: ecf_resource1.id)
-        expect(course).to have_attributes(ecf_resource1.attributes.slice(:identifier, :name))
+        expect(course).to have_attributes(ecf_resource1.attributes.slice("identifier", "name"))
         expect(course.course_group.name).to eq(Courses::DEFINITIONS.find { |d| d[:identifier] == ecf_resource1.identifier }[:course_group_name])
       end
 
--- a/spec/services/migration/migrators/declaration_spec.rb
+++ b/spec/services/migration/migrators/declaration_spec.rb
@@ -5,7 +5,9 @@ RSpec.describe Migration::Migrators::Declaration do
     let(:records_per_worker_divider) { 2 }
 
     def create_ecf_resource
-      create(:ecf_migration_participant_declaration, :ineligible)
+      travel_to(rand(100).hours.ago) do
+        create(:ecf_migration_participant_declaration, :ineligible)
+      end
     end
 
     def create_npq_resource(ecf_resource)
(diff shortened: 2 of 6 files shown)


spec/services/migration/migrators/declaration_spec.rb
```diff
@@ -26,7 +26,7 @@ def setup_failure_state
       it "sets the created Declaration attributes correctly" do
         instance.call
         declaration = Declaration.find_by(ecf_id: ecf_resource1.id)
-        expect(declaration).to have_attributes(ecf_resource1.attributes.slice(:created_at, :udpated_at, :declaration_date, :declaration_type, :state))
+        expect(declaration).to have_attributes(ecf_resource1.attributes.slice("created_at", "udpated_at", "declaration_date", "declaration_type", "state"))
```
cwrw, line 29: this is now failing as we aren't setting the updated at correctly in the declaration migrator, could we look into that please

spec/services/migration/migrators/course_spec.rb
```diff
@@ -23,7 +23,7 @@ def setup_failure_state
       it "sets the created Course attributes correctly" do
         instance.call
         course = Course.find_by(ecf_id: ecf_resource1.id)
-        expect(course).to have_attributes(ecf_resource1.attributes.slice(:identifier, :name))
+        expect(course).to have_attributes(ecf_resource1.attributes.slice("identifier", "name"))
```
cwrw, line 26: this is worrying 😓

spec/services/migration/migrators/declaration_spec.rb
```diff
@@ -5,16 +5,17 @@
...
-      create(:ecf_migration_participant_declaration, :ineligible)
+      travel_to(rand(100).hours.ago) do
+        create(:ecf_migration_participant_declaration, :ineligible)
+      end
     end
 
     def create_npq_resource(ecf_resource)
-      lead_provider = create(:lead_provider, ecf_id: ecf_resource.cpd_lead_provider.npq_lead_provider.id)
-      cohort = create(:cohort, ecf_id: ecf_resource.cohort_id, start_year: ecf_resource.cohort.start_year)
+      create(:lead_provider, ecf_id: ecf_resource.cpd_lead_provider.npq_lead_provider.id)
```
cwrw, line 14: could we revert those to what they were?

spec/services/migration/migrators/declaration_spec.rb
```diff
@@ -26,18 +27,21 @@ def setup_failure_state
...
+        expect(declaration.created_at.to_s).to eq(ecf_resource1.created_at.to_s)
+        expect(declaration.updated_at.to_s).to eq(ecf_resource1.updated_at.to_s)
+        expect(declaration.declaration_date.to_s).to eq(ecf_resource1.declaration_date.to_s)
       end
 
       context "when declaration date is before the schedule start" do
         before do
-          declaration = Declaration.find_by(ecf_id: ecf_resource1.id)
-          ecf_resource1.update!(declaration_date: declaration.application.schedule.applies_from.prev_week)
+          application = Application.joins(:user).find_by(users: { ecf_id: ecf_resource1.user.id })
```
cwrw, line 43: same here?

spec/services/migration/migrators/declaration_spec.rb
leandroalemao: ok will do.. but I thought without creating an existing declaration in NPQ it was a more real life scenario as we don't have any declarations in production.

mooktakim (approved):

PR #1874, opened 2024-10-29
Add POST/PUT endpoint support to parity check

ethax-ross:
### Context

We want to add POST and PUT endpoints to the parity check tool to ensure the responses and behaviour of the API is consistent between ECF and NPQ reg.

- Add POST endpoints to parity check

Support `post` requests in the parity check tool. In order to do this we expand the `Client` to support POST requests and allow specifying a `payload` in the YAML for a path. This can either be a hash, which will be sent as the request body or it can be a method name that can construct a dynamic payload.

- Add PUT endpoints to parity check

Support `put` requests in the parity check tool. In order to do this we expand the `Client` to support PUT requests and add specific use cases to the YAML file/client.

- Memoize request body

To avoid issues with dynamic payloads and ensure both the NPQ and ECF requests use the exact same body we can memoize it.

- Exclude fields that will be different in POST/PUT

Mainly the updated/created attributes that will have different values.

Minor bug fixes to some of the `id` queries off the back of errors we got during testing.

- Fix feature specs

- Treat non-200 responses as needs review

Currently, if both services return a non-200 response and its the same then it will be flagged as equal. Whilst this is technically correct its not very helpful as it can hide issues with the calls themselves.

Instead, we want to raise these as 'unexpected' results so that we get an indication something is wrong.

- Minor bug fixes and tweaks

Small changes to get the POST/PUT comparison results looking better/more reliable.

### Guidance for review

Best reviewed by-commit.

screencapture-0-0-0-0-3000-npq-separation-migration-parity-checks-2024-10-29-11_38_04

### DIFF
--- a/config/parity_check_endpoints.yml
+++ b/config/parity_check_endpoints.yml
@@ -91,18 +91,24 @@ get:
 
 post:
   "/api/v1/npq-applications/:id/accept#with_funded_place":
+    exclude:
+      - updated_at
     id: application_ecf_id_for_accept_with_funded_place
     payload:
       type: "npq-application-accept"
       attributes:
         funded_place: true
   "/api/v2/npq-applications/:id/accept#with_funded_place":
+    exclude:
+      - updated_at
     id: application_ecf_id_for_accept_with_funded_place
     payload:
       type: "npq-application-accept"
       attributes:
         funded_place: true
   "/api/v3/npq-applications/:id/accept#with_funded_place":
+    exclude:
+      - updated_at
     id: application_ecf_id_for_accept_with_funded_place
     payload:
       type: "npq-application-accept"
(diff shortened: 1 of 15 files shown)


cwrw (commented):
fantastic work! will test tomorrow on migration 🙌

config/parity_check_endpoints.yml
```diff
@@ -88,3 +88,233 @@ get:
...
+    exclude:
+      - id
+      - updated_at
+      - created_at
+      - mentor_id
+      - evidence_held
+      - delivery_partner_id
+    payload: post_declaration_payload
+
+  "/api/v1/participants/npq/:id/outcomes":
```
cwrw, line 184: I think we might need to exclude `updated_at` here as well

cwrw (approved):
🙌 !!

mooktakim (approved):

PR #1882, opened 2024-10-30
[CPDLP-3518] Add migration queue only worker to NPQ

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3518

* Migration jobs will use `migration` queue
* Existing worker runs the following queues: `default,low_priority,participant_outcomes`
* New migration worker runs `migration` queue only

### DIFF
--- a/app/jobs/parity_check_job.rb
+++ b/app/jobs/parity_check_job.rb
@@ -1,5 +1,5 @@
 class ParityCheckJob < ApplicationJob
-  queue_as :migration
+  queue_as :high_priority
 
   def perform
     parity_check = Migration::ParityCheck.new
--- a/app/jobs/parity_check_job.rb
+++ b/app/jobs/parity_check_job.rb
@@ -1,5 +1,5 @@
 class ParityCheckJob < ApplicationJob
-  queue_as :high_priority
+  queue_as :migration
 
   def perform
     parity_check = Migration::ParityCheck.new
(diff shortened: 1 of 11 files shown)


terraform/application/application.tf
```diff
@@ -69,11 +69,36 @@ module "worker_application" {
 
   docker_image = var.docker_image
 
-  command       = ["bundle", "exec", "rake", "jobs:work"]
+  command       = ["/bin/sh", "-c", "QUEUES=default,low_priority,participant_outcomes bundle exec rake jobs:work"]
```
ethax-ross, line 72: Worth adding a comment in the delayed_job initializer pointing to this? I'm just thinking if someone adds a queue in the future they would expect it to be picked up by default and not realise they'll need to add it here too

ethax-ross (commented):
Looks good, has this been deployed to migration yet?

app/jobs/parity_check_job.rb
```diff
@@ -1,5 +1,5 @@
 class ParityCheckJob < ApplicationJob
-  queue_as :high_priority
+  queue_as :migration
```
cwrw, line 2: I don't think this needs to move to migration 🤔

cwrw (commented):
looking good 🙌  happy to approve once manually we run plan/apply on migration and all is well :)

app/jobs/parity_check_job.rb
mooktakim: changed it back

leandroalemao (approved):
Approving so it can be tested in migration 👍  tks @mooktakim 👍

PR #1884, opened 2024-10-30
[CPDLP-3693] Data cleanup - lookup application directly by id in decl…

mooktakim:
…aration migrator

### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3693

* Use `ecf_id` from `participant_profile.npq_application.id` directly for declarations.

### DIFF
--- a/spec/services/migration/migrators/declaration_spec.rb
+++ b/spec/services/migration/migrators/declaration_spec.rb
@@ -32,6 +32,8 @@ RSpec.describe Migration::Migrators::Declaration do
         expect(declaration.state_reason).to eq(ecf_resource1.declaration_states.last.state_reason)
         expect(declaration.cohort.start_year).to eq(ecf_resource1.cohort.start_year)
         expect(declaration.lead_provider.ecf_id).to eq(ecf_resource1.cpd_lead_provider.npq_lead_provider.id)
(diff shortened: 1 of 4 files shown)


spec/services/migration/migrators/declaration_spec.rb
```diff
@@ -32,8 +32,7 @@ def setup_failure_state
         expect(declaration.state_reason).to eq(ecf_resource1.declaration_states.last.state_reason)
         expect(declaration.cohort.start_year).to eq(ecf_resource1.cohort.start_year)
         expect(declaration.lead_provider.ecf_id).to eq(ecf_resource1.cpd_lead_provider.npq_lead_provider.id)
-        expect(declaration.application.course.identifier.downcase).to eq(ecf_resource1.course_identifier.downcase)
```
cwrw, line 35: not sure why we removed those tests, they should still hold 🤔

cwrw (approved):
just one question

spec/services/migration/migrators/declaration_spec.rb
mooktakim: I thought they weren't need anymore as we're testing if application is correct. 
I've changed it back now.

leandroalemao (approved):

mooktakim:
Tested on migration env

PR #1891, opened 2024-10-31
Set swagger URL to current environment

ethax-ross:
[Jira-3637](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3637)

### Context

The swagger URL is hard-coded as the migration environment. Instead, we always want this to reflect the current environment.

- Set swagger URL to current environment

We currently hardcode the swagger URL; it is packaged into the Docker image and we deploy the same image to all environments.

We want to have the current domain in the URL for whichever environment the application is deployed to. In order to do this, we can replace the URL on application boot. It's not ideal that we do this on every boot, but the alternative would involve a separate docker image for each environment which would be a lot harder to manage.

### Guidance for review

[API docs v1](https://npq-registration-review-1891-web.test.teacherservices.cloud/api/docs/v1)
[API docs v2](https://npq-registration-review-1891-web.test.teacherservices.cloud/api/docs/v2)
[API docs v3](https://npq-registration-review-1891-web.test.teacherservices.cloud/api/docs/v3)


leandroalemao (approved):
👍 awesome! tks @ethax-ross 👍

config/initializers/swagger_server_url.rb
```diff
@@ -0,0 +1,6 @@
+return if ENV["DOMAIN"].blank?
+
+Dir[Rails.root.join("public/api/docs/**/swagger.yaml")].each do |swagger_file|
+  swagger_doc = File.read(swagger_file)
+  File.write(swagger_file, swagger_doc.gsub("http://0.0.0.0:3000", ENV["DOMAIN"]))
```
cwrw, line 5: should this include `https//:`? as that domain doesn't? Also not sure if this is what is affecting the authorisation and trying it out in the actual docs, as we get a url like this:
```
'https://npq-registration-review-1891-web.test.teacherservices.cloud/api/docs/v3/cpd-npq-review-1891-web.test.teacherservices.cloud/api/v3/npq-applications?cohort=2021
```

leandroalemao: Also `DOMAIN` is being set as the internal domain in the review app? `https://cpd-npq-review-1891-web.test.teacherservices.cloud`

So seeing an error:

```
Failed to fetch.
Possible Reasons: CORS
Network Failure URL scheme must be "http" or "https" for CORS request.
```

public/api/docs/v1/swagger.yaml
```diff
@@ -923,7 +923,7 @@ paths:
               "$ref": "#/components/schemas/ParticipantChangeScheduleRequest"
         required: true
 servers:
-- url: https://npq-registration-separation-web.teacherservices.cloud/
+- url: http://0.0.0.0:3000
```
mooktakim, line 926: Minor comment
Instead of "http://0.0.0.0:3000", should we just have the variable name:
```yaml
url: ACCESS_EXTERNAL_DOMAIN
```
That way the gsub is more specific.

mooktakim (approved):

terraform/application/application.tf
```diff
@@ -1,7 +1,7 @@
 locals {
   environment  = "${var.environment}${var.pull_request_number}"
   service_name = "cpd-npq"
-  domain       = var.environment == "review" ? "cpd-npq-${local.environment}-web.test.teacherservices.cloud" : module.web_application.hostname
+  domain       = var.environment == "review" ? "cpd-registration-${local.environment}-web.test.teacherservices.cloud" : module.web_application.hostname
```
cwrw, line 4: `
```suggestion
  domain       = var.environment == "review" ? "npq-registration-${local.environment}-web.test.teacherservices.cloud" : module.web_application.hostname
```

terraform/application/config/staging.yml
```diff
@@ -0,0 +1,2 @@
+---
+ACCESS_EXTERNAL_DOMAIN: npq-registration-staging-web.teacherservices.cloud
```
cwrw, line 2:
```suggestion
ACCESS_EXTERNAL_DOMAIN: npq-registration-staging-web.test.teacherservices.cloud
```

public/api/docs/v1/swagger.yaml
ethax-ross: Good suggestion, but it would need to be a valid URL or rswag complains; it could be `http://swagger.url` or something, but I figured it was easier to leave it as the development URL so it just works in that environment (otherwise we'd need to have more logic to set it explicitly as dev doesn't have the `ACCESS_EXTERNAL_DOMAIN` env var)

leandroalemao (approved):

mooktakim (approved):

cwrw (approved):
looks good 🙌  do you mind applying this (after rebasing) on migration after it's been PO tested just incase we missed something just incase since we're changing tf stuff :)

ethax-ross:
> looks good 🙌 do you mind applying this (after rebasing) on migration after it's been PO tested just incase we missed something since we're changing tf stuff :)

Deployed to the migration just now and I've re-tested

PR #1893, opened 2024-10-31
Run parity check in parallel

ethax-ross:
[Jira-3673](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3673)

### Context

The parity check service takes a long time to run in production (~2hr 45m). We want to improve the performance so it can run more quickly.

- Run parity check in parallel

The parity check currently takes around 3 hours to run. Most of the time is spent waiting for network requests to complete.

Instead of making the service more complex by splitting across jobs, we should be able to get a decent performance bump by running it across multiple threads, so we can make several network requests in parallel.

Split parity check across threads; one thrad per lead provider (there are 10 lead providers, so this should give a decent performance bump).

Run request to each service in parallel in the `Client`. This should effectively double the throughput.

Raise errors earlier so we don't have to deal with individual threads raising errors.

- Split parity check across multiple workers

As the workers only have 4 CPU cores we found multithreading to not be that efficient.

Instead, this commit will split the requests across multiple workers. It uses a mutex on the database to ensure the last worker to finish will finalise the parity check.

The testing becomes quite complicated/contrived due to relying on accessing `Delayed::Job` directly for the check status (so we can't use the `:test` queue adapter).

Also adds progress % during run to get an idea of how far along it is; that being said, the few endpoints that are paginated can skew this so the last ~5% can take ~30% of the total time, for example.

### Guidance to review

Roughly 2hrs 2m quicker by splitting across workers. It is also now stress testing the applications, which is showing in the response times 😬 

| Before | After | After (progress) |
| -- | -- | -- |
| Screenshot 2024-11-01 at 14 03 22 | Screenshot 2024-11-01 at 14 05 20 | Screenshot 2024-11-01 at 14 07 00 |


ethax-ross:
Closing this as there are issues with ordering and running async (the POST requests can effect the outcome of GET requests)

PR #1899, opened 2024-11-01
[CPDLP-3694] Refresh record in action endpoint services

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3694

* Updated `Applications::Accept` to reload application after action
* Updated `Declarations::Void` to reload declaration after action


spec/support/shared_examples/application_action_support.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.shared_examples "a application accept action" do
```
cwrw, line 5:
```suggestion
RSpec.shared_examples "an application accept action" do
```
everywhere :)

app/services/applications/accept.rb
```diff
@@ -26,6 +26,8 @@ def accept
         reject_other_applications_in_same_cohort!
       end
 
+      application.reload
```
ethax-ross, line 29: If you comment this reload out the request specs still pass; I think it might make more sense to keep the request specs generic and have the more specific service specs cover this 🤔

mooktakim: The service spec won't test this as we're concerned with what is returned by the endpoint, which would use the serializer. We can be limited with how much we do with the request tests.

cwrw (commented):
looks good to me past Ross's comment around the tests and ensuring they fail 👍

app/services/applications/accept.rb
leandroalemao: Maybe you could add something like `expect_any_instance_of(Application).to receive(:reload).and_call_original` to `accept_spec.rb`?

mooktakim: The issue was with Declaration service classes. Added reloads and kept the test just in case.
@leandroalemao added your suggestion.

cwrw (approved):

app/services/applications/accept.rb
ethax-ross: I still think these tests would be better in the service layer; the new request specs are inconsistent with what we're got at the moment and I think it'll make it confusing going forward.

The existing request specs test that a request is handled by the correct service and the correct serializer is given resource returned by that service. The issue is that we're serializing an object without reloading it and since we're fixing that by changing the service, that's where the test should be IMO. As @leandroalemao suggested we could check it calls reload. To make it more specific you could stub `reload` on the actual object with something like:

```
allow(service.application).to receive(:reload)
service.accept
expect(service.application).to have_received(:reload)
```

mooktakim: Moved tests to service, instead of the requests. 👍

ethax-ross (approved):

PR #1912, opened 2024-11-05
Fix bug with providers not populating in serializers

ethax-ross:
[Jira-3713](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3713)

### Context

We are finding that some ITT providers and private childcare providers are coming back as `nil` in the API response where they should be populated (as they are in ECF).

We also want to ensure that the `updated_at` of the application in NPQ reg is set to the current date if there is a provider in NPQ reg but not in ECF, so that the change in data will be surfaced to lead providers.

- Ensure disabled providers are included in serializers

We have a default scope on the providers that omits disabled providers by default. This is causing the providers to be `nil` in the serializers, but the tests are passing as the scope is bypassed due to the providers being preloaded in the spec already.

Update specs to ensure the application is reloaded and the default scopes are correctly applied.

Update `Application` to descope via new `belongs_to` associations suffixed with `including_disabled` so we return disabled providers when accessed via the application.

Improve test coverage of nullable values in `ApplicationCsvSerializer`.

- Update updated_at of application if provider only in NPQ reg

If we migrate an application from ECF that doesn't have a private childcare provider and/or itt provider, but the corresponding application in NPQ reg does have one of these set, we want to ensure the `updated_at` is updated to the current time so that the change will be surfaced to providers.

I've opted to include disabled providers here for consistency, but all the
current providers in NPQ reg should be active.

### DIFF
--- a/spec/serializers/api/application_serializer_spec.rb
+++ b/spec/serializers/api/application_serializer_spec.rb
@@ -116,7 +116,7 @@ RSpec.describe API::ApplicationSerializer, type: :serializer do
 
       subject(:attributes) { JSON.parse(described_class.render(application.reload))["attributes"] }
 
-      context "when the `itt_provider`` is set" do
+      context "when the `itt_provider` is set" do
         let(:itt_provider) { create(:itt_provider) }
 
         it "serializes the `itt_provider`" do
--- a/spec/serializers/api/application_serializer_spec.rb
+++ b/spec/serializers/api/application_serializer_spec.rb
@@ -110,9 +110,18 @@ RSpec.describe API::ApplicationSerializer, type: :serializer do
     end
 
     describe "itt_provider serialization" do
-      it "serializes the `itt_provider`" do
-        itt_provider.legal_name = "provider"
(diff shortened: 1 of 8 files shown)


cwrw (approved):

spec/serializers/api/application_serializer_spec.rb
```diff
@@ -110,9 +110,18 @@
...
-      it "serializes the `itt_provider`" do
-        itt_provider.legal_name = "provider"
-        expect(attributes["itt_provider"]).to eq(itt_provider.legal_name)
+      # We need to persist and then reload the application
+      # to ensure the default scope is applied/accounted for.
+      before { application.save! }
+
+      subject(:attributes) { JSON.parse(described_class.render(application.reload))["attributes"] }
+
+      context "when the `itt_provider`` is set" do
```
leandroalemao, line 119: There's an extra ` 

context "when the `itt_provider` is set" do

leandroalemao (approved):

PR #1916, opened 2024-11-06
[CPDLP-3715] Favour the school urn from profile over the application urn in application migrator

leandroalemao:
### Context

Ticket: [CPDLP-3715](https://dfedigital.atlassian.net/browse/CPDLP-3715)

In some applications in the parity check the application school urn is different to that on the profile. We believe the profile should be the source of truth as it can be amended manually after being accepted. Since we only have one value in NPQ we have to favour the profile school urn.

- Amend the application migrator to favour the school urn from profile first then the application urn to link the school.
- If the school urn is updated we need to ensure the application updated_at also shows this change
- include the following fields in the comparison for updated at:
  - `ineligible_for_funding_reason`

[CPDLP-3715]: https://dfedigital.atlassian.net/browse/CPDLP-3715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/migration/migrators/application.rb
```diff
@@ -85,7 +86,14 @@ def call
 
         application.user_id = find_user_id!(ecf_id: ecf_npq_application.user.id)
         application.version_note = "Changes migrated from ECF to NPQ"
-        application.update!(ecf_npq_application.attributes.slice(*ATTRIBUTES))
+
+        attrs = ecf_npq_application.attributes.slice(*ATTRIBUTES)
+
+        if touch_updated_at?(application, ecf_npq_application)
+          attrs[:updated_at] = Time.zone.now
```
mooktakim, line 93: fyi `updated_at` needs to be string.

mooktakim (approved):

app/services/migration/migrators/application.rb
```diff
@@ -105,6 +108,16 @@ def provider_only_in_npq?(application, ecf_npq_application)
         ecf_npq_application.private_childcare_provider_urn.nil? && application.private_childcare_provider_including_disabled.present?
     end
 
+    def touch_updated_at?(application, ecf_npq_application)
+      ecf_npq_application.profile&.school_urn.presence != ecf_npq_application.school_urn ||
+        application.ineligible_for_funding_reason != ecf_npq_application.ineligible_for_funding_reason ||
```
cwrw, line 113: I don't think this will work properly, as we might not have all the applications from the migrator done yet as the calculation is expected on all applications for that user. Lets leave ineligible for funding out for now
I'll rethink how we add it in again

cwrw (approved):

mooktakim (approved):

ethax-ross (approved):

app/services/migration/migrators/application.rb
```diff
@@ -124,6 +129,10 @@ def touch_updated_at?(application, ecf_npq_application)
       false
     end
 
+    def ecf_profile_school_urn_or_ecf_npq_application_school_urn?(ecf_npq_application)
+      ecf_npq_application.profile&.school_urn.presence || ecf_npq_application.school_urn
```
cwrw, line 133: when including the profile school urn, some schools don't exist with some urns (around 6). Can we add a method which finds the school id but doesn't fail if it can't find it, we can also keep the existing method as its needed elsewhere for finding schools and failing

app/services/migration/migrators/application.rb
```diff
@@ -70,7 +70,7 @@ def call
 
         school_urn = ecf_profile_school_urn_or_ecf_npq_application_school_urn?(ecf_npq_application)
         if school_urn
-          application.school_id = find_school_id!(urn: school_urn)
+          application.school_id = find_school_id(urn: school_urn)
```
cwrw, line 73: can we add a test for this please?

cwrw (approved):

leandroalemao:
Closing the PR to open a new one.

PR #1917, opened 2024-11-06
Add analytics to npq

slawosz:
This PR is introducing [def-analytics](https://github.com/DFE-Digital/dfe-analytics) to `NPQ`.
Currently analytics will be send only to `staging` environment to test integration.

There is extra PR to be merge to allow excluding certain models, for the time being `npq` will be using fork: https://github.com/DFE-Digital/dfe-analytics/pull/169


config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,60 @@
...
+  config.environment = ENV.fetch('RAILS_ENV', 'development')
+
+  # A proc which will be called with the user object, and which should
+  # return the identifier for the user. This is useful for systems with
+  # users that don't use the id field.
+  #
+  # config.user_identifier = proc { |user| user&.id
+
+  # if part of any model name is migration, it wont be loaded
+ config.excluded_models_proc = proc { |x| x.to_s =~ /Migration::/ }
```
peteryates, line 59: Can the migration things not be added to the blocklist for now?

peteryates (approved):

peteryates:
Looks good, I think you just need to regenerate the blocklists.

config/initializers/dfe_analytics.rb
slawosz: The migration needs to be added, otherwise plugin will not work.
Currently we have models like `Statement` and `Migration::Statement`, which are pointing to 2 databases, but both are using `statements` table. The `dfe-analytics` operates on tables names so it gets confused by 2 tables that have the same name but are using 2 different schemas.

Excluding `Migration::` tables does not give much difference - all important data will be exported anyway.

peteryates: Ah, makes sense! Thanks.

config/analytics.yml
```diff
@@ -0,0 +1,303 @@
...
+    - name
+    - created_at
+    - updated_at
+    - ecf_id
+    - description
+    - position
+    - display
+    - identifier
+    - course_group_id
+  :contract_templates:
```
cwrw, line 194: are we sure we want to send this across, in ECF we avoided sending payment data?

config/analytics.yml
```diff
@@ -0,0 +1,303 @@
...
+    - full_name
+    - date_of_birth
+    - trn_verified
+    - active_alert
+    - national_insurance_number
+    - trn_auto_verified
+    - feature_flag_id
+    - provider
+    - uid
+    - raw_tra_provider_data
```
cwrw, line 296: should we be sending this 🤔

config/analytics.yml
```diff
@@ -0,0 +1,303 @@
...
+    - reason
+    - created_at
+    - updated_at
+    - ecf_id
+  :applications:
+    - id
+    - user_id
+    - course_id
+    - lead_provider_id
+    - DEPRECATED_school_urn
```
cwrw, line 237: maybe we can take out the deprecated fields from here?

config/analytics_pii.yml
```diff
@@ -0,0 +1,21 @@
+---
+shared:
+  :registration_interests:
+    - email
+  :private_childcare_providers:
+    - registered_person_name
+  :contracts:
```
cwrw, line 7: not sure the contract is PII?

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,60 @@
...
+  # config.bigquery_retries = 3
+
+  # Passed directly to the timeout: option on the BigQuery client
+  #
+  # config.bigquery_timeout = 120
+
+  # A proc which returns true or false depending on whether you want to
+  # enable analytics. You might want to hook this up to a feature flag or
+  # environment variable.
+  config.enable_analytics = proc { Rails.env.staging? }
```
cwrw, line 44: should we add a feature flag instead so we can enable it in the envs we want incase?

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,60 @@
...
+
+
+  # The environment we’re running in. This value will be attached
+  # to all events we send to BigQuery.
+  #
+  config.environment = ENV.fetch('RAILS_ENV', 'development')
```
cwrw, line 50: not sure we want to enable this?

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  # A proc which will be called with the user object, and which should
+  # return the identifier for the user. This is useful for systems with
+  # users that don't use the id field.
+  #
+  # config.user_identifier = proc { |user| user&.id
```
cwrw, line 56: could we add 
```
config.user_identifier = proc { |user| user&.id if user.respond_to?(:id) }
```
incase as it can be an issue trying to get the user out sometimes, hopefully not an issue in NPQ but just incase?

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,60 @@
...
+  #
+  # config.log_only = true
+
+  # Whether to use ActiveJob or dispatch events immediately.
+  #
+  # config.async = true
+
+  # Which ActiveJob queue to put events on
+  #
+  # config.queue = :default
```
leandroalemao, line 12: Shouldn't be better to have a specific queue for these events?
ie: `config.queue = :dfe_analytics`

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,60 @@
...
+  # Passed directly to the timeout: option on the BigQuery client
+  #
+  # config.bigquery_timeout = 120
+
+  # A proc which returns true or false depending on whether you want to
+  # enable analytics. You might want to hook this up to a feature flag or
+  # environment variable.
+  config.enable_analytics = proc { Rails.env.staging? }
+
+
```
leandroalemao, line 46: We also have this job enabled in ECF.. does it make any difference if we also add here?

https://github.com/DFE-Digital/dfe-analytics?tab=readme-ov-file#entity-table-check-job

leandroalemao (commented):
Is there any spec to add maybe?
Also It'd be a nice idea to have a look at the yml files in ECF as a base to generate the lists here?

https://github.com/DFE-Digital/early-careers-framework/blob/main/config/analytics_blocklist.yml
and
https://github.com/DFE-Digital/early-careers-framework/blob/main/config/analytics_pii.yml

PR #1934, opened 2024-11-08
[CPDLP-3743] Add Redis cache to production env

leandroalemao:
### Context

Ticket: [CPDLP-3743](https://dfedigital.atlassian.net/browse/CPDLP-3743)

We currently use Redis in migration for the data migration, we need to set a similar one up in production.

Add Redis cache to production env.

[CPDLP-3743]: https://dfedigital.atlassian.net/browse/CPDLP-3743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/terraform/application/config/production.tfvars.json
+++ b/terraform/application/config/production.tfvars.json
@@ -18,5 +18,9 @@
     "start_minute": 0
   },
   "external_url": "https://register-national-professional-qualifications.education.gov.uk/healthcheck",
-  "enable_logit": true
+  "enable_logit": true,
+  "deploy_redis_cache": true,
+  "redis_cache_family": "P",
+  "redis_cache_capacity": 2,
+  "redis_cache_sku_name": "Premium"
 }
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -44,7 +44,21 @@ Rails.application.configure do
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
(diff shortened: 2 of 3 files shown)


terraform/application/config/production.tfvars.json
```diff
@@ -18,5 +18,9 @@
     "start_minute": 0
   },
   "external_url": "https://register-national-professional-qualifications.education.gov.uk/healthcheck",
-  "enable_logit": true
+  "enable_logit": true,
+  "deploy_redis_cache": true,
+  "redis_cache_family": "P",
+  "redis_cache_capacity": 2,
```
cwrw, line 24: maybe 1 instead? as we have 1 in ECF right now?

ethax-ross: Might be worth running this past Colin; I think we tend to use smaller capacity for queues (1/P/Premium) and larger capacity for cache (2/C/Standard)

cwrw: Colin suggested we use a premium with high network performance

leandroalemao: So should we keep (2/P/Premium) then?

image

cwrw: yes you're right, let's keep it

cwrw (approved):

cwrw (approved):

ethax-ross (approved):

PR #1943, opened 2024-11-11
Bump number of web apps

ethax-ross:
We want to get the performance on par with ECF; during testing we're a bit slower and it seems the web apps are the bottleneck; going to 12 to see if that improves the situation in the short-term until we have time to optimise.


ethax-ross:
Going to re-raise as review app is busted.

PR #1956, opened 2024-11-12
[CPDNPQ-2117] qualifications API spike - with legacy_passed_participant_outcomes from DQT

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2117

DQT is being decommissioned and TRS needs an API to call

- Added a table to store DQT data.
- API will include entries that match in this table, as well as NPQ participant outcomes.
- Includes all changes in https://github.com/DFE-Digital/npq-registration/pull/1910

### Things that could be improved

- optimise the query
- check if additional indexes are required on the tables

Note - this is just a spike, and won't be merged until made production-ready, and should not be merged before separation.


config/routes.rb
```diff
@@ -219,6 +219,10 @@
 
       resources :statements, only: %i[index show], param: :ecf_id
     end
+
+    namespace :teacher_record_service, path: "teacher-record-service", defaults: { format: :json } do
```
cwrw, line 223: this seems to be under v3 although the controller is v1, I suggest this is decoupled from the lead provider API as we also have feature flags hiding the API currently. and possibly move this to a v1 as they shouldn't be connected or block each other esp regarding versions?

app/controllers/api/v1/qualifications_controller.rb
```diff
@@ -0,0 +1,26 @@
+module API
```
cwrw, line 1: talked in person: left over from other pr to remove

app/serializers/api/qualifications_serializer.rb
```diff
@@ -0,0 +1,18 @@
+module API
+  class QualificationsSerializer
+    def self.render(participant_outcomes, root:, trn:)
```
cwrw, line 3: discussed in person: although a bit hard we're going to try to use blue printer

cwrw:
closing as review app sending too many logs, will reopen when @alkesh is back to investigate :)

PR #1962, opened 2024-11-13
Add logging for DQT API requests

ethax-ross:
Add logging so that we can monitor that these requests are happening and completing successfully when we switch them on in production.

### DIFF
--- a/app/services/dqt/v1/teacher.rb
+++ b/app/services/dqt/v1/teacher.rb
@@ -14,7 +14,11 @@ module Dqt
         }
         query[:nino] = nino if nino
 
+        Rails.logger.info("DQT API request started")
+
         response = get(path, query:)
+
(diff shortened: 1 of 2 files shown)


app/services/dqt/v1/teacher.rb
```diff
@@ -14,8 +14,12 @@ def self.find(trn:, birthdate:, nino: nil)
         }
         query[:nino] = nino if nino
 
+        Rails.logger.info("DQT API request started")
```
cwrw, line 17: I guess this is after the day as those will trickle in but not instantly?

ethax-ross: I have a snippet to check the integration on the day which will verify the keys/secrets work, but this is just an extra check to ensure we can monitor requests are actually happening/successful

PR #1964, opened 2024-11-13
Bump web apps in prod/migration

ethax-ross:
[Jira-3634](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3634)

We want to get the performance on par with ECF; during testing we're a bit slower and it seems the web apps are the bottleneck; going to 12 to see if that improves the situation in the short-term until we have time to optimise. Move to 6 workers now we are running 12 web apps.

Bump migration workers to 30 to match migration environment ahead of running migration in production.


terraform/application/config/migration.tfvars.json
```diff
@@ -7,8 +7,8 @@
     "redis_cache_capacity": 0,
     "redis_cache_sku_name": "Basic",
     "enable_logit": true,
-    "webapp_replicas": 6,
-    "worker_replicas": 4,
+    "webapp_replicas": 12,
```
cwrw, line 10: let's scale down to 3/2 for now as discussed?

cwrw (approved):

PR #1967, opened 2024-11-13
Make konduit work in Docker Compose

rwrrll:
### Context

(No ticket)

Installs packages needed for Konduit database connection, makes `sed` invocations compatible with GNU sed, and updates docs


jebw (approved):
LGTM

PR #1971, opened 2024-11-14
Ensure placeholder TRNs are unverified

ethax-ross:
[Jira-3763](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3763)

### Context

We have a number of placeholder TRNs in the ECF dataset; we want to ensure when we migrate these that they are always in an unverified state.

- Ensure placeholder TRNs are unverified

### Guidance for review

Checked on migration:

```
User.where(trn: %w[0000000 1234567], trn_verified: true).exists?
=> false
```

### DIFF
--- a/app/services/migration/migrators/user.rb
+++ b/app/services/migration/migrators/user.rb
@@ -1,5 +1,7 @@
 module Migration::Migrators
   class User < Base
+    PLACEHOLDER_TRNS = %w[0000000 1234567].freeze
+
     class << self
       def record_count
         ecf_users.count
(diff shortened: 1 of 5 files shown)


cwrw (commented):
Looks good, could we amend the user factory to change the trn from 1234567 to something like this `sprintf("%07i", Random.random_number(9_999_999))`?

leandroalemao (approved):
All good tks @ethax-ross 👍

cwrw (approved):

PR #1905, opened 2024-11-04
[CPDNPQ-2170] Run smoke test during deployment

jebw:
### Context

Ticket: [CPDNPQ-2170](https://dfedigital.atlassian.net/browse/CPDNPQ-2170)

Current the application does not run smoke tests during deployment - the healthchecks are tested as part of the AKS deployment module but only that they respond with success. This does additional checks such as ensuring the migrations have completed correctly

### Changes proposed in this pull request

1. Added a `bin/smoke` program to smoke test the new application
2. Added a github action to run the `bin/smoke` program to perform the smoke tests
3. Integrated the github action in (2) with the deploy_to_aks action so that smoke tests occur during deployment into all environments

### Failed feature specs screenshots

If any of the feature specs would fail, there will be page on the wiki with all
failures:

https://github.com/DFE-Digital/npq-registration/wiki/

[CPDNPQ-2170]: https://dfedigital.atlassian.net/browse/CPDNPQ-2170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1979, opened 2024-11-15
Re-instate SHA check on deploy

ethax-ross:
### Context

We reverted #1905 due to seeing failures in the review app deployments (see #1973). We think these failures were caused by other flakiness/issues with the pipeline, so we want to re-instate the SHA check.

- Re-instate SHA check on deploy

Re-instate the SHA check on deploy that was removed in case it was causing flakiness; we were seeing failures on the SHA check, but they appear to be related to other issues with the pipeline.

Check the correct SHA is deployed, once per second for a maximum of 10 seconds (to give time for the old pods to drain and new pods to spin up).

### DIFF
--- a/bin/smoke
+++ b/bin/smoke
@@ -1,20 +1,35 @@
 #!/usr/bin/env bash
 
-sleep 10 # Allow time for old app instances to drain
-
 url=$1
 if [[ -z $url ]]; then
   echo `date`" - smoke test failed (URL is missing)"
   exit 1
 fi
-response=$(curl -sL $url/healthcheck)
-response_sha=$(jq ".git_commit_sha" <<< $response)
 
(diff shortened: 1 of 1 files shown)


bin/smoke
```diff
@@ -1,21 +1,36 @@
...
 current_commit_sha=\"$2\"
 if [[ -z $current_commit_sha ]]; then
   echo `date`" - smoke test failed (head sha is missing)"
   exit 1
 fi
 
+max_retries=10
+
+for ((i=1; i<=max_retries; i++)); do
+  response=$(curl -sL $url/healthcheck)
```
cwrw, line 18: I think response needs to remain outside the loop as it's used elsewhere 🤔

cwrw (approved):

PR #1984, opened 2024-11-18
Cleanup dfe analytics config

slawosz:


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,8 @@
 class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
+  include DfE::Analytics::Requests
```
ethax-ross, line 4: Worth adding a quick request spec to cover this? something like:

```
it "sends DFE Analytics web request event" do
  expect { get root_path }.to have_sent_analytics_event_types(:web_request)
end
```

app/jobs/crons/check_analytics_entity.rb
```diff
@@ -0,0 +1,15 @@
+class Crons::CheckAnalyticsEntity < CronJob
+  include Sentry::Cron::MonitorCheckIns
+
+  self.cron_expression = "0 2 * * *"
+
+  sentry_monitor_check_ins slug: "check-analytics-entity"
+
+  def perform
+    DfE::Analytics::EntityTableCheckJob.perform_later
```
ethax-ross, line 9: Can we add a test for this job? might be worth adding a commit message/comment around why we're adding this as well?

app/services/feature.rb
```diff
@@ -69,5 +71,9 @@ def maintenance_banner_enabled?
     def enable_ecf_api_disabled!
       Flipper.enable(ECF_API_DISABLED)
     end
+
+    def dfe_analytics_enabled?
```
ethax-ross, line 75: Can we test this please

app/controllers/application_controller.rb
cwrw: could we also add this in the `API::BaseController` to cover API requests as well?

config/initializers/dfe_analytics.rb
```diff
@@ -55,4 +55,6 @@
 
   # if part of any model name is migration, it wont be loaded
   config.excluded_models_proc = proc { |x| x.to_s =~ /Migration::/ }
+
+  config.entity_table_checks_enabled = true
 end
```
cwrw, line 60: could we add `config.user_identifier = proc { |user| user&.id if user.respond_to?(:id) }` here as well incase as we had issues in Ecf because of user?

cwrw (approved):
thanks @slawosz!

leandroalemao (approved):
Looks good tks @slawosz ! 👍

PR #1990, opened 2024-11-19
[CPDNPQ-2218] Follow up from CPDNPQ-2141

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2218 

Follow up from CPDNPQ-2141 to improve specs and refactor (no new functionality)

I've left commits unsquashed to help with review, I can squash before merge if preferred!

### DIFF
--- a/app/components/npq_separation/admin/course_payment_overview_component.rb
+++ b/app/components/npq_separation/admin/course_payment_overview_component.rb
@@ -19,7 +19,7 @@ module NpqSeparation
 
       def counts
         allowed_declaration_types.inject({}) { |m, dt|
-          m.merge(dt.titleize => billable_declarations_count_for_declaration_type(dt))
+          m.merge(dt.titleize => declaration_count_for_declaration_type(dt))
         }.merge({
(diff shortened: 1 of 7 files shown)


spec/features/npq_separation/admin/finance/course_payment_breakdown_spec.rb
```diff
@@ -1,437 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
-
-RSpec.feature "ECF legacy spec for course payment breakdown", :ecf_api_disabled, :js, type: :feature do
```
cwrw, line 5: same here around feature coverage, are we covering everything elsewhere 🤔 ?

cwrw (commented):
looking good, a few minor comments

cwrw (approved):
thanks @rwrrll looks good, I think worth getting Jake to test with you on the NPQ snapshot and comparing it with prod just to be safe incase we missed anything :)

PR #1991, opened 2024-11-19
Add `significantly_updated_at` attribute to `User`

ethax-ross:
[Jira-3747](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3747)

### Context

At the moment, we use the `User#updated_at` to surface changes to lead providers via the API. The issue with this is that actions/changes that lead providers don't care about result in the user being marked as 'updated'. Instead, we want to have an additional `updated_at` attribute that we can use to surface the changes lead providers care about to the underlying `User` model.

- Add significantly_udpated_at attribute to User 

We have an `updated_at` on the `User` model, however this is updated frequently and for actions that lead providers don't need to care about (such as when a user signs in).

We want to add a second `updarted_at` attribute that only updates when something significant changes with the `User`, such as their name, TRN, etc.

As the `Application` can touch a `User` we also need this attribute to update when the user is touched.

I opted for `significantly_updated_at` as the attribute name as opposed to `api_updated_at` to try and keep it more generic in case we want to surface it elsewhere.

Index the field to ensure endpoints that filter by `updated_at` will be performant when we switch them to  significantly_updated_at`.

- Switch User migrator to populate significantly_updated_at

We want to populate `significantly_udpated_at` instead of `updated_at` as we will be exposing this attribute in the serializers for the lead provider API.

- Update Participants::Query to use significantly_udpated_at

We want to filter by the `significantly_updated_at` attribute instead of the `updated_at` attribute when querying users for the lead provider API. This will prevent insignificant changes to a `User` resulting in them being re-surfaced in the API.

- Update ParticipantSerializer to use significant_updated_at

We want to surface the `sigificant_updated_at` attribute of participants in the lead provider API responses so that providers only see changes/updates when something they care about has changed with the user.

### Guidance to review

As it stands any users already in the NPQ reg database that aren't touched as part of the migration will end up with a default `significantly_updated_at` time of 'now' (when we roll this out). Do we want to backfill to match with the user `updated_at` for these records?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -54,7 +54,7 @@ class User < ApplicationRecord
 
   enum email_updates_status: EMAIL_UPDATES_ALL_STATES, _suffix: true
 
-  attr_accessor :version_note
+  attr_accessor :version_note, :skip_touch_significantly_updated_at
 
   def latest_participant_outcome(lead_provider, course_identifier)
     declarations.eligible_for_outcomes(lead_provider, course_identifier)
@@ -216,6 +216,8 @@ class User < ApplicationRecord
 private
 
   def touch_significantly_updated_at
+    return if skip_touch_significantly_updated_at
+
     changed_attributes = saved_changes.keys
 
     explicitly_updating_significantly_updated_at = changed_attributes.include?("significantly_updated_at")
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,15 @@
 class User < ApplicationRecord
+  INSIGNIFICANT_ATTRIBUTES = %w[
+    raw_tra_provider_data
+    feature_flag_id
+    get_an_identity_id_synced_to_ecf
+    updated_from_tra_at
+    trn_lookup_status
+    notify_user_for_future_reg
+    email_updates_status
+    email_updates_unsubscribe_key
+  ].freeze
+
   devise :omniauthable, omniauth_providers: [:tra_openid_connect]
 
   has_paper_trail meta: { note: :version_note }, ignore: [:raw_tra_provider_data]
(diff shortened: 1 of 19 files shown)


leandroalemao (commented):
It's looking good tks.. 👍  I reckon we can backfill `significantly_updated_at` to be the same as `updated_at` for the other records as part of the ticket as well?

app/services/participants/query.rb
```diff
@@ -37,7 +37,7 @@ def where_lead_provider_is(lead_provider)
     def where_updated_since(updated_since)
       return if ignore?(filter: updated_since)
 
-      users_updated_since = User.where(updated_at: updated_since..)
+      users_updated_since = User.where(significantly_updated_at: updated_since..)
```
cwrw, line 40: can we also do this in the application query as well?

app/services/migration/migrators/user.rb
```diff
@@ -58,12 +58,12 @@ def call
           active_alert: npq_application.active_alert || user.active_alert,
           trn_verified:,
           created_at: ecf_user.created_at,
-          updated_at: ecf_user.updated_at,
+          significantly_updated_at: ecf_user.updated_at,
```
leandroalemao, line 61: Maybe we still can keep migrating `updated_at` over or not?

app/models/user.rb
```diff
@@ -199,4 +212,18 @@ def unsubscribe_from_email_updates
   def archived?
     archived_email.present?
   end
+
+private
+
+  def touch_significantly_updated_at
```
cwrw, line 218: discussed in person: disable on migration

app/services/migration/migrators/user.rb
cwrw: discussed in person: let's keep both

leandroalemao (approved):
All good 👍 tks @ethax-ross 😄

cwrw (approved):

PR #1993, opened 2024-11-19
Remove migrators and associated code

ethax-ross:
> ⚠️ Do not merge until post-separation!

[Jira-3786](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=unassigned%2C712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3786)

### Context

Once we have ran the migration in production we no longer need/want the migrators and associated code in the codebase (we will no longer have a need for it and it becomes a potential liability).

- Remove migrators and associated code

Once we have ran the migration in production we no longer need/want the migrators and associated code in the codebase (we will no longer have a need for it and it becomes a potential liability).

Remove all migrators and associated code.

Leave parity check tool in place; this will be removed in a separate commit.

- Remove nightly migration GitHub action

Remove the GitHub action that runs and reports on a nightly migration.

- Drop DataMigration model/table

- Remove migration workers

We no longer need these workers as we're finished with the migration.


terraform/application/application.tf
```diff
@@ -79,28 +79,3 @@ module "worker_application" {
 
   enable_logit = var.enable_logit
 }
-
-module "migration_worker_application" {
```
cwrw, line 83: should we also change the other worker to pick up all queues now (remove specifying the queue on the other worker)?

ethax-ross: Ah, yeah good catch I'll update it

cwrw (approved):
😢

leandroalemao (approved):

PR #1994, opened 2024-11-20
Add extra fields to analytics with rake task to backfill

slawosz:
https://dfedigital.atlassian.net/browse/CPDNPQ-2227


config/analytics.yml
```diff
@@ -254,6 +254,9 @@ shared:
     - schedule_id
     - referred_by_return_to_teaching_adviser
     - accepted_at
+    - senco_in_role
+    - senco_start_date
+    - application_trn
```
cwrw, line 259: I think this makes me a bit worried as we should avoid having multiple points of data that point to the same thing, i.e. trn which should only be on User. What is the application trn field? Can we just send the `raw_application_data` instead to avoid this and they have access to this data instead, I think that could be easier?

slawosz: Raw application data is a json field and I am not sure if  it would be fine with data team. Also fields differ from application to application.

cwrw: could we ask them and test this maybe on a review app and set the creds to allow testing on review to confirm how it looks like, they don't need to parse the whole field just the fields we care about? I really would like to avoid having application trn on the application as it's very risky data wise, and definitely cause confusion down the line

AntPage1: Just to clarify, the application trn is the value provided by the user when applying. The TRN from users is the true value. We need both to be able to measure the volume of applicants who provide incorrect information at the point of application and this also acts as the unique identifier for applicants who are never accepted.

cwrw: Thanks Tony that makes sense. However since this is already stored in the the raw data field can we just use that and send it over? We have this set up in ECF in a similar way in the model and it has caused a lot of confusion and issues down the line, so hoping to avoid this early on in NPQ model-wise

AntPage1: Thanks Abeer, we should be able to handle the JSON record column. The sooner we can have this the sooner we can test it. As I'm not 100% sure what is held in this `raw_application_data` field from applications we do not want any PII data coming through so if there is a way of limiting the JSON to these 3 items only and we receive it as one column we should be able to make this work.

cwrw: unfortunately we have to follow this route in the end as filtering the raw application data on backfill didn't work

cwrw (commented):
looks good, might be worth adding a service for the backfill with testing. I was wondering about new applications and saving those values going forward, as we're only backfilling existing values?

db/schema.rb
```diff
@@ -526,10 +529,10 @@
     t.boolean "active_alert", default: false
     t.text "national_insurance_number"
     t.boolean "trn_auto_verified", default: false
+    t.string "feature_flag_id"
```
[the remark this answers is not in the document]
slawosz: Its auto-generated by rails.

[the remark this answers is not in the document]
cwrw: I don't get this diff when running migrations locally, let's keep them out for now I think

cwrw (approved):

PR #2006, opened 2024-11-22
[CPDNPQ-2254] users not able to register fix

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2254

When a user has logged in via DfE Identity, we attempt to update the user record with the latest attributes from DfE Identity.
Sometimes, the email change clashes with another user account we already have, so the error causes the page to redirect to the start of the journey.

When this scenario occurs, and there is a clashing user account, we will archive the clashing user account, and move its applications over to the current user account.


app/controllers/omniauth_controller.rb
```diff
@@ -100,6 +104,10 @@ def try_to_extract_user_uid
   rescue StandardError
     "unknown-provider-uid"
   end
+
+  def user_has_email_change_that_will_clash_with_existing_users
+    @user.changes[:email] && User.where(email: @user.changes[:email].last).any?
```
cwrw, line 109: having a look at the method `find_or_create_from_provider_data`, we are saving the user there before we get here, so we might have the user already with the new email and this might always return true if email changed and return the user we actually want to keep?

app/services/users/archive_by_email.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+      users_with_matching_email.each do |user_with_matching_email|
+        Users::Archiver.new(user: user_with_matching_email).archive!
+      end
+    end
+
+  private
+
+    def users_with_matching_email
+      @users_with_matching_email = User.where(email: user.email).all
```
cwrw, line 25: can we ignore the user we're passing in from here as in `User.where(email: user.email).where.not(id: user.id)` otherwise we might archive the primary user

cwrw: just incase should we guard against when email could be `nil` just incase saving that user went wrong here and return early:
```
return [] unless user.email
```

app/services/users/archive_by_email.rb
```diff
@@ -0,0 +1,34 @@
...
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :user
+
+    def call
+      ApplicationRecord.transaction do
+        users_with_matching_email.each do |user_with_matching_email|
+          move_applications(from_user: user_with_matching_email, to_user: user)
+          user.participant_id_changes.create!(from_participant_id: user_with_matching_email.ecf_id, to_participant_id: user.ecf_id) if Feature.ecf_api_disabled?
```
cwrw, line 14:
```suggestion
           # Move previous change history to new to_user
          from_user.participant_id_changes.update!(user: to_user)
          
          user.participant_id_changes.find_or_create_by!(from_participant_id: user_with_matching_email.ecf_id, to_participant_id: user.ecf_id) if Feature.ecf_api_disabled?
```
just incase we have id changes on old user, also if there are duplicates on the to user

app/controllers/omniauth_controller.rb
alkesh: If the email changed to one that doesn't clash, then the user will be saved successfully, and user.changes will be empty, as the new record has been persisted.
I'll add a feature spec to cover this scenario.

app/services/users/archiver.rb
```diff
@@ -14,6 +14,8 @@ def archive!
         user.archived_email = user.email
         user.archived_at = Time.zone.now
         user.email = "archived-#{user.email}"
+        user.uid = nil
```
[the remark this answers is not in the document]
alkesh: Even in migration - should we be making the UID nil?
What will happen otherwise is a user will log in with an archived user account, and it will be matched via the UID, and the email, name, etc will be updated to the latest from DfE Identity - and it will actually not really be an archived user - except that `archived_at` will still be populated, and `archived?` will return true.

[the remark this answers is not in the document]
cwrw: good point, although we are clearing it elsewhere I think worth doing that here anyway let's keep it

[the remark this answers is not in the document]
cwrw: might be worth just double checking the code to ensure that we never associate anyone with an archived user, I think this helps ensure that, but maybe the last thing would be ecf id, so worth checking the code (possibly without code changes), happy to chat through this

app/models/user.rb
```diff
@@ -59,7 +59,8 @@ def self.find_or_create_from_provider_data(provider_data, feature_flag_id:)
 
     if user.persisted?
       unless user.valid?
-        Rails.logger.info("[GAI] User persisted BUT not valid, #{user.errors.full_messages.join(';')}, ID=#{user.id}, UID=#{provider_data.uid}, trying to join account")
+        Rails.logger.info("[GAI] User persisted BUT not valid, #{user.errors.full_messages.join(';')}, ID=#{user.id}, UID=#{provider_data.uid}, archiving account with clashing email")
```
alkesh, line 62: this error message seemed to be a cut-and-paste of the one just below, and it wasn't accurate (no joining of accounts occurred here)

app/models/user.rb
```diff
@@ -59,7 +59,8 @@ def self.find_or_create_from_provider_data(provider_data, feature_flag_id:)
 
     if user.persisted?
       unless user.valid?
-        Rails.logger.info("[GAI] User persisted BUT not valid, #{user.errors.full_messages.join(';')}, ID=#{user.id}, UID=#{provider_data.uid}, trying to join account")
```
alkesh, line 62: Changing this error message, as it is identical to one further down, so seems to be a cut-and-paste error.

app/services/users/archive_by_email.rb
```diff
@@ -0,0 +1,44 @@
...
+      ApplicationRecord.transaction do
+        users_with_matching_email.each do |user_with_matching_email|
+          move_applications(from_user: user_with_matching_email, to_user: user)
+          if Feature.ecf_api_disabled?
+            move_participant_id_changes(from_user: user_with_matching_email, to_user: user)
+            user.participant_id_changes.find_or_create_by!(from_participant_id: user_with_matching_email.ecf_id, to_participant_id: user.ecf_id)
+          end
+        end
+      end
+      users_with_matching_email.each do |user_with_matching_email|
```
leandroalemao, line 20: Is there any reason why this block is outside the transaction? As you're doing a loop in `users_with_matching_email` twice..

app/services/users/archive_by_email.rb
```diff
@@ -0,0 +1,44 @@
...
+        Users::Archiver.new(user: user_with_matching_email).archive!
+      end
+    end
+
+  private
+
+    def users_with_matching_email
+      return [] unless user.email
+
+      @users_with_matching_email = User.where(email: user.email).where.not(id: user.id).all
```
leandroalemao, line 31: There's a unique `email` constraint in the DB, so this query will always return none or 1 record max..

cwrw: we also don't need the `.all` here

app/services/users/archive_by_email.rb
cwrw: the archiver has it's own transaction

app/services/users/archive_by_email.rb
alkesh: Good point - I think I'll rename this method to `user_with_matching_email` and make it return either `nil` or a user.

app/services/users/archive_by_email.rb
```diff
@@ -0,0 +1,42 @@
...
+      if user_with_matching_email
+        ApplicationRecord.transaction do
+          move_applications(from_user: user_with_matching_email, to_user: user)
+          if Feature.ecf_api_disabled?
+            move_participant_id_changes(from_user: user_with_matching_email, to_user: user)
+            user.participant_id_changes.find_or_create_by!(from_participant_id: user_with_matching_email.ecf_id, to_participant_id: user.ecf_id)
+          end
+        end
+        Rails.logger.info("Archiving user with clashing email address ID=#{user_with_matching_email.id}")
+        Users::Archiver.new(user: user_with_matching_email).archive!
```
cwrw, line 20: just incase before this could we add:
```
user.reload
user_with_matching_email.reload
```
we hit issues when we tried to update records before so just incase :)

cwrw (approved):
Looks great, thank you @alkesh! just a very minor comment but otherwise 👍 

Discussed in person but for posterity:
- Jake to test on review
- Confirm no lookups made via `ecf_id` for user: doesn't look like it
- ⚠️ This can be merged after separation - hopefully on the 28th in the morning
Thanks!

app/services/users/archive_by_email.rb
alkesh: I've added the code to reload `user_with_matching_email`.
For `user` - I've made this attribute private, so it doesn't need to be reloaded.
I can see in other services, the attributes need to be reloaded, because they are accessed outside the service, once the service has been called - I don't think that kind of access is needed in this service.

PR #2022, opened 2024-11-28
Fix rate limiting

ethax-ross:
[Jira-3806](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3806)

We had a catch-all/generic rate limit that was set to 300 requests/5 minutes. This is lower than all of our other throttles in RackAttack and was causing API requests to be throttled incorrectly.

Instead, we want to add an additional throttle for non-API requests at 300 requests/minute and a catch all/backstop throttle that will ensure no request to the service will exceed 1500 requests/5 minutes.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -61,6 +61,11 @@ class Rack::Attack
     auth_token(request) if api_request?(request) && !public_api_path?(request)
   end
 
+  # Throttle public /api requests by ip (300 requests per 5 minutes)
+  throttle("public API requests by ip", limit: 300, period: 5.minutes) do |request|
+    request.ip if public_api_path?(request)
+  end
+
   # Throttle non-api requests (300 requests per 5 minutes)
   throttle("non-API requests by ip", limit: 300, period: 5.minutes) do |request|
     request.ip unless api_request?(request)
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -17,32 +17,59 @@ class Rack::Attack
     "/api/docs",
   ].freeze
 
+  def self.api_request?(request)
+    request.path.starts_with?("/api/")
+  end
+
+  def self.public_api_path?(request)
+    PUBLIC_API_PATH_PREFIXES.any? { |prefix| request.path.starts_with?(prefix) }
+  end
+
+  def self.protected_path?(request)
+    PROTECTED_ROUTES.include?(request.path)
+  end
+
+  def self.csp_report_path?(request)
+    request.path == "/csp_reports"
+  end
+
+  def self.get_an_identity_webhook_path?(request)
+    request.path.starts_with?("/api/v1/get_an_identity/webhook_messages")
+  end
+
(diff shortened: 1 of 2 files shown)


config/initializers/rack_attack.rb
```diff
@@ -17,32 +17,59 @@ class Rack::Attack
...
 
+  def self.api_request?(request)
+    request.path.starts_with?("/api/")
+  end
+
+  def self.public_api_path?(request)
+    PUBLIC_API_PATH_PREFIXES.any? { |prefix| request.path.starts_with?(prefix) }
+  end
+
+  def self.protected_path?(request)
```
cwrw, line 28: if these aren't reused methods do we need to move them out here 🤔

ethax-ross: we don't _have_ to, but I thought it made the rules easier to read 🤷‍♂️ happy to put them back if you feel differently

config/initializers/rack_attack.rb
```diff
@@ -17,32 +17,59 @@ class Rack::Attack
...
-      request.get_header("HTTP_AUTHORIZATION")
-    end
+  # Throttle non-api requests (300 requests per 5 minutes)
+  throttle("non-API requests by ip", limit: 300, period: 5.minutes) do |request|
+    request.ip unless api_request?(request)
   end
 
-  # Throttle all requests by IP (300 requests per 5 minutes)
-  throttle("general requests by ip", limit: 300, period: 5.minutes, &:ip)
+  # Catch all/backstop; throttle all requests by IP (3000 requests per 5 minutes).
```
cwrw, line 69:
```suggestion
  # Catch all/backstop; throttle all requests by IP (1500 requests per 5 minutes).
```

cwrw (approved):
looks good let's test!

config/initializers/rack_attack.rb
```diff
@@ -17,32 +17,59 @@ class Rack::Attack
...
-    public_api_path = PUBLIC_API_PATH_PREFIXES.any? { |prefix| request.path.starts_with?(prefix) }
+    auth_token(request) if api_request?(request) && !public_api_path?(request)
+  end
 
-    if request.path.starts_with?("/api/") && !public_api_path
-      request.get_header("HTTP_AUTHORIZATION")
-    end
+  # Throttle non-api requests (300 requests per 5 minutes)
+  throttle("non-API requests by ip", limit: 300, period: 5.minutes) do |request|
+    request.ip unless api_request?(request)
```
cwrw, line 66: should we also ensure `public_api_path` is included here?

ethax-ross: We could do, might be worth having a separate, explicit throttle for them or as it stands they just fallback to the catchall

ethax-ross: Pushed a commit with an explicit rate limit for them

mooktakim (approved):

leandroalemao (approved):

PR #2023, opened 2024-11-28
[CPDLP-3795] Add missing indexes to statements

leandroalemao:
### Context

From the performance analysis, we noticed NPQ does not perform as good as ECF. We need to understand why since we have a bit of time

- add_index :statements, %i[lead_provider_id output_fee]
- add_index :statements, :payment_date


leandroalemao:
Closing this as the benchmarks, made with the indexes created, scored worse than the benchmarks made without the indexes created.. not sure why. c/c @cwrw

PR #2025, opened 2024-11-28
[CPDNPQ-2270] port rake file and service from ECF (bulk script for revert to pending)

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2270

bulk script for revert to pending - rake file and service ported from ECF - updated to work with the NPQ `Applications::RevertToPending` service.

### DIFF
--- a/lib/tasks/bulk_change_applications_to_pending.rake
+++ b/lib/tasks/bulk_change_applications_to_pending.rake
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "rake"
-
 namespace :one_off do
   # Bulk change NPQ applications to pending.
   # Accepts a CSV of application (ecf) IDs (without a header row) and a dry run
@@ -21,7 +19,7 @@ namespace :one_off do
     dry_run = args[:dry_run] != "false"
     unless File.exist?(csv_file_path)
       logger.error "File not found: #{csv_file_path}"
-      return
+      exit 1
     end
 
     application_ecf_ids = CSV.read(csv_file_path, headers: false).flatten
--- a/lib/tasks/bulk_change_applications_to_pending.rake
+++ b/lib/tasks/bulk_change_applications_to_pending.rake
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rake"
+
+namespace :one_off do
+  # Bulk change NPQ applications to pending.
+  # Accepts a CSV of application (ecf) IDs (without a header row) and a dry run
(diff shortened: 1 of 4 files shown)


lib/tasks/bulk_change_applications_to_pending.rake
```diff
@@ -0,0 +1,35 @@
...
+    dry_run = args[:dry_run] != "false"
+    unless File.exist?(csv_file_path)
+      logger.error "File not found: #{csv_file_path}"
+      return
+    end
+
+    application_ecf_ids = CSV.read(csv_file_path, headers: false).flatten
+
+    logger.info "Bulk changing #{application_ecf_ids.size} applications to pending#{' (dry run)' if dry_run}..."
+
```
jebw, line 30: `sleep(10)` - gives the developer time to think 'oops' and press ctrl-c ? Might be belt and braces though, there's already a dry run feature? thoughts?

spec/lib/tasks/one_off/bulk_change_applications_to_pending_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    subject { Rake::Task["one_off:bulk_change_to_pending"].invoke(csv_file.path) }
+
+    it "does not update any applications" do
+      subject
+      expect(Application.all.pluck(:lead_provider_approval_status)).to eq %w[accepted accepted]
+    end
+  end
+
+  context "when dry run true" do
+    subject { Rake::Task["one_off:bulk_change_to_pending"].invoke(csv_file.path, "true") }
```
jebw, line 29: `subject(:run_task)`

spec/lib/tasks/one_off/bulk_change_applications_to_pending_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    it "does not update any applications" do
+      subject
```
jebw, line 32: `run_task` ? Clarifies what subject is doing ?

jebw (approved):
Couple of minor code clarity suggestions but looks great :raised_hands:

lib/tasks/bulk_change_applications_to_pending.rake
alkesh: I think this one might be overkill.
I'd expect the developer will already run this with dry-run first, and the effort in creating/uploading the CSV file in the first place means it shouldn't be a trivial thing to run accidentally.

PR #2026, opened 2024-11-28
Remove parity check service and ECF integration points

ethax-ross:
[Jira-3791](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3791)

### Context

We have separated NPQ from ECF and no longer need to run parity checks; so we want to remove all of the parity check-related code.

- Remove parity check service and ECF integration points

Remove parity check service, remaining ECF models and views.

Revert rack_attack to be enabled in migration.

Remove gems used for parity check.

Remove ECF schema and databse connection.

### Guidance for review

Once this has been merged down we can remove the following secrets from keyvault:

```
ECF_DB_DATABASE
ECF_DB_USERNAME
ECF_DB_PASSWORD
ECF_DB_HOST
ECF_DB_PORT
```


app/services/migration/parity_check/token_provider.rb
```diff
@@ -1,36 +0,0 @@
-module Migration
-  class ParityCheck::TokenProvider
```
cwrw, line 2: can we remove this from the ECF side as well?

cwrw: can we also remove this from `refresh_migration_database`?

cwrw (commented):
looking good, could we also remove the specs from the exclusions in `.erdconfig` and also could we regenerate the `knapsack_rspec_report`?

cwrw (approved):

leandroalemao (approved):
All good tks @ethax-ross

PR #2027, opened 2024-11-29
CPDNPQ-2154 Manage training status

jebw:
### Context

Ticket: [CPDNPQ-2154](https://dfedigital.atlassian.net/browse/CPDNPQ-2154)

We want to allow admin users to assign or change the training status for users

1. Add a service + specs to validate and update the change to the training status
2. Added a controller + view + specs to utilise the service to update the training status
3. Added a feature spec to verify the workflow

[CPDNPQ-2154]: https://dfedigital.atlassian.net/browse/CPDNPQ-2154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/applications/change_training_status.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Applications
+  class ChangeTrainingStatus
```
cwrw, line 4: we have existing services for defer/resume/withdraw, can we use those instead?

jebw:
Code updated to use the `Participants::Action` services - I have added a #call method to those - the alternatives I could see were either

1) a `case` statement to build the delegated service, then a second `case` statement to call the action so that I could maintain a handle onto the delegated service to extract errors onto the calling service
2) errors not making it through to the UI

alkesh (approved):

app/views/npq_separation/admin/applications/change_training_statuses/new.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <div class="govuk-inset-text">
+      <p class="govuk-body">
+        <strong>User ID / Participant ID</strong>
+        <br/>
+        <%= @application.user.id %>
+      </p>
+      <p class="govuk-body">
+        <strong>Training status</strong>
+        <br/>
+        <%= @application.training_status&.humanize || "- not set - " %>
```
alkesh, line 18:
```suggestion
        <%= @application.training_status&.humanize || "<em>- not set -</em>".html_safe %>
```
wdyt?

PR #2029, opened 2024-11-29
[CPDNPQ-2155] Manage funding eligibility

jebw:
### Context

Ticket: [CPDNPQ-2155](https://dfedigital.atlassian.net/browse/CPDNPQ-2155)

As an admin user I need to update the funding eligibility on an application

1. Add a service + specs to validate and update the change to the training
2. Added a controller + view + specs to utilise the service to update the training status
3. Added a feature spec to verify the workflow

[CPDNPQ-2155]: https://dfedigital.atlassian.net/browse/CPDNPQ-2155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/applications/change_funding_eligibility.rb
+++ b/app/services/applications/change_funding_eligibility.rb
@@ -35,8 +35,8 @@ module Applications
   private
 
     def validate_funding_eligiblity_status_code_change
-      if declared_as_billable? && eligible_for_funding == false
-        errors.add(:base, :billable_declaration_exists)
+      if declared_as_billable_or_changeable? && eligible_for_funding == false
+        errors.add(:base, :declaration_exists)
       end
     end
 
@@ -46,8 +46,8 @@ module Applications
       end
     end
 
-    def declared_as_billable?
-      application.declarations.billable.count.positive?
+    def declared_as_billable_or_changeable?
+      application.declarations.billable_or_changeable.count.positive?
     end
   end
 end
--- a/app/services/applications/change_funding_eligibility.rb
+++ b/app/services/applications/change_funding_eligibility.rb
@@ -5,18 +5,25 @@ module Applications
     include ActiveModel::Model
     include ActiveModel::Attributes
 
+    OPTIONS = {
+      true => I18n.t("shared.yes"),
+      false => I18n.t("shared.no"),
+    }.freeze
+
     attribute :application
     attribute :eligible_for_funding, :boolean
 
-    validates :eligible_for_funding, inclusion: [true, false]
     validates :application, presence: true
+    validates :eligible_for_funding, inclusion: OPTIONS.keys
+
+    def eligible_for_funding_options
+      OPTIONS
+    end
 
     def change_funding_eligibility
-      Application.transaction do
-        return false if invalid?
+      return false if invalid?
 
-        application.update!(eligible_for_funding:)
-      end
+      application.update!(eligible_for_funding:)
     end
   end
 end
--- a/app/services/applications/change_funding_eligibility.rb
+++ b/app/services/applications/change_funding_eligibility.rb
@@ -23,7 +23,10 @@ module Applications
     def change_funding_eligibility
       return false if invalid?
 
-      application.update!(eligible_for_funding:)
+      funding_eligiblity_status_code =
+        (eligible_for_funding ? :marked_funded_by_policy : :marked_ineligible_by_policy)
+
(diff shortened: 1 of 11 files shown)


app/services/applications/change_funding_eligibility.rb
```diff
@@ -0,0 +1,29 @@
...
+    OPTIONS = {
+      true => I18n.t("shared.yes"),
+      false => I18n.t("shared.no"),
+    }.freeze
+
+    attribute :application
+    attribute :eligible_for_funding, :boolean
+
+    validates :application, presence: true
+    validates :eligible_for_funding, inclusion: OPTIONS.keys
```
cwrw, line 17: I think we need to take into account `funded_place` and if it's set, as `funded_place` cannot be true if `eligible_for_funding` is false, worth confirming if its on accepted vs pending applications as `funded_place` is set on accepted applications from cohort 2024 onwards

cwrw: also worth checking declaration logic, if this is allowed on applications with declarations or not, if it is this becomes much more complex

jebw: This appears to be the relevant logic in ECF, noting it only gets applied within the admin area so I've added it to the service rather than Application itself

* https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/npq_application.rb#L74-L75

spec/requests/npq_separation/admin/applications/change_funding_eligibilities_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        post npq_separation_admin_applications_change_funding_eligibility_path(application, params:)
+      end
+
+      let(:params) { { applications_change_funding_eligibility: { eligible_for_funding: "1" } } }
+
+      it { is_expected.to redirect_to npq_separation_admin_application_path(application) }
+    end
+  end
+
+  context "when not logged in" do
```
slawosz, line 35: This probably could be extracted to some `shared_context`?

app/services/applications/change_funding_eligibility.rb
```diff
@@ -0,0 +1,53 @@
...
+      funding_eligiblity_status_code =
+        (eligible_for_funding ? :marked_funded_by_policy : :marked_ineligible_by_policy)
+
+      application.update!(eligible_for_funding:, funding_eligiblity_status_code:)
+    end
+
+  private
+
+    def validate_funding_eligiblity_status_code_change
+      if declared_as_billable? && eligible_for_funding == false
```
cwrw, line 38: I think if we have submitted declarations and we mark the eligible for funding as true, that means those declarations need to become eligible. I think worth checking with Jake the intended logic, should we allow this at all on applications with declarations, or should we change the declarations state themselves?

spec/views/npq_separation/admin/applications/change_funding_eligibilities/new.html.erb_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let :form_path do
+    npq_separation_admin_applications_change_funding_eligibility_path(application)
+  end
+
+  it { is_expected.to have_css("h1", text: application.user.full_name) }
+  it { is_expected.to have_css(%(form[action="#{form_path}"]), count: 1) }
+  it { is_expected.to have_css(%(.govuk-form-group fieldset label), text: "Yes") }
+  it { is_expected.to have_css(%(.govuk-form-group fieldset label), text: "No") }
+  # it { is_expected.to have_css("optgroup", count: 2) }
+  # it { is_expected.to have_css("optgroup option") }
```
alkesh, line 25: can these be removed? or uncommented?

jebw: d'oh!

spec/requests/npq_separation/admin/applications/change_funding_eligibilities_controller_spec.rb
slawosz: I see it would be not beneficial - likely we would end up with half of the controllers using shared context, half using direct code.

alkesh (approved):

PR #2032, opened 2024-12-02
Update validations post separation

ethax-ross:
[Hira-3653](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3653)

### Context

We have a `npq_separation` context on `User` and `Application` that is no longer used; we want to make this validation apply to all models. We also want to relax the validation on `User` so that they can change their `uid`.

- Update validations post-separation

Now that we have separated, we want to udpate the `User` and `Application` validation as we no longer need the `on: :npq_separation` context (the data is gnow consistent post-migration).

We also want to relax the `uid` validation on `User` as it is not valid for a user to change their `uid` to another value.

### Guidance for review

We have checked production data to ensure all models will be valid with the de-scoped validation rules.


app/models/user.rb
```diff
@@ -28,7 +28,6 @@ class User < ApplicationRecord
             notify_email: true
 
   validates :uid, uniqueness: { allow_blank: true }
-  validates :uid, inclusion: { in: ->(user) { [user.uid_was] } }, on: :npq_separation, if: -> { uid_was.present? }
   # TODO: add constraints into the DB after separation
```
leandroalemao, line 31: Should we remove those conditions and add DB constraints as well in this another? or Is it in another ticket?

ethax-ross: The `TODO:`'s are for another ticket 👍

cwrw (approved):
Thanks Ross. As discussed worth creating an application via the FE adding criteria

PR #2034, opened 2024-12-02
[CPDLP-3774] Remove internal APIs (NPQ Post Separation Cleanup)

leandroalemao:
### Context

Ticket: [CPDLP-3774](https://dfedigital.atlassian.net/browse/CPDLP-3774)

Now separation is complete we need to remove the internal APIs that used to talk to ECF

- Remove ECF internal APIs and any logic that used to be calculated regarding that - also ONLY cleanup the feature flags there, as there is another ticket to remove the feature flag completely
- Remove unnecessary jobs that used to be used in the internal APIs
- Remove any unnecessary cron jobs as well
- Check for any keys that need removing


cwrw (commented):
@leandroalemao thank you for this. This is a very big pr and would be great if we can break it down to meaningful explained commits so it will make it easier to review? happy to chat

app/services/participant_validator.rb
```diff
@@ -9,58 +9,19 @@ def initialize(trn:, full_name:, date_of_birth:, national_insurance_number: nil)
   end
 
   def call
-    if Feature.ecf_api_disabled?
-      call_with_dqt
-    else
-      call_with_ecf
+    result = Dqt::RecordCheck.new(**payload.merge(check_first_name_only: true)).call
+    if result.total_matched >= 3
```
ethax-ross, line 13: Why is 3 the magic number? 🤔 is it worth having this as a constant to help explain?

spec/features/journeys/happy_paths/when_get_an_identity_returns_no_trn_spec.rb
```diff
@@ -1,15 +1,15 @@
...
   include ApplicationHelper
 
   include_context "retrieve latest application data"
-  include_context "Stub previously funding check for all courses" do
-    # In this situation we fallback to a non-pilot set of checks
-    let(:api_call_trn) { manually_entered_trn }
-  end
+  # include_context "Stub previously funding check for all courses" do
+  #   # In this situation we fallback to a non-pilot set of checks
+  #   let(:trn) { manually_entered_trn }
```
ethax-ross, line 11: Can we just delete these setup if its not used?

mooktakim: I have actually removed that context on my pr:
https://github.com/DFE-Digital/npq-registration/pull/2042

app/services/participant_validator.rb
mooktakim: The mystical number that came from the ancient times of ecf lol

ethax-ross (approved):

mooktakim (approved):

spec/support/helpers/journey_assertion_helper.rb
```diff
@@ -36,9 +36,6 @@ def expect_applicant_reached_end_of_journey
         expect(page).to have_text("Registration successfully submitted")
         expect(page).to have_link("Register for another NPQ", href: /\/registration\/provider_check/)
       end
-
-      expect(User.count).to be(1)
```
cwrw, line 40: Im wondering why those needed to go?

cwrw (commented):
Looks great thanks @leandroalemao well done. Only one question otherwise happy to approve after

spec/support/helpers/journey_assertion_helper.rb
leandroalemao: Discussed via a call

cwrw (approved):
thanks @leandroalemao

mooktakim (approved):

PR #2042, opened 2024-12-03
[CPDLP-3775] NPQ Post Separation Cleanup - Remove feature flag

mooktakim:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3775

* Updated `Application.ecf_id` to auto generated UUID
* Removed feature flag from specs
* Deleted some ecf endpoint services that no longer needed


app/jobs/application_submission_job.rb
```diff
@@ -1,88 +0,0 @@
-class ApplicationSubmissionJob < ApplicationJob
```
cwrw, line 1: Thos are in Leandro's pr, can we rebase off his branch otherwise we're stepping on each other's toes?

mooktakim: Sure, will rebase after Leandro's PR

db/migrate/20241203191058_auto_gen_uuid_ecf_id_for_applications.rb
```diff
@@ -0,0 +1,6 @@
+class AutoGenUuidEcfIdForApplications < ActiveRecord::Migration[7.1]
+  def change
+    change_column_default :applications, :ecf_id, from: nil, to: "gen_random_uuid()"
```
cwrw, line 3: let's look into user id as well

PR #2060, opened 2024-12-09
move specs out of lib into correct matching folders

alkesh:
### Context

We have a  lot of specs in the lib folder, instead of the correct matching folder for the code.

move specs in `specs/lib` to the correct `specs/forms` or `specs/services` folders.


alkesh:
moved to a new PR due to failing review app deploy: https://github.com/DFE-Digital/npq-registration/pull/2064

PR #2064, opened 2024-12-10
move specs out of lib into correct matching folders

alkesh:
### Context

We have a lot of specs in the lib folder, instead of the correct matching folder for the code.

move specs in `specs/lib` to the correct` specs/forms` or `specs/services` folders


cwrw (commented):
can we hold off on this pr, we have some big prs removing large changes from the old implementation and integration with ECF that touches most of those, as this will mess up those prs please?

slawosz (approved):

PR #2066, opened 2024-12-11
Updating contract data, fixes 2350 data request

slawosz:
https://dfedigital.atlassian.net/browse/CPDNPQ-2350

Rake task has hardcoded arguments as the request is quite critical, so it would make sense to create some sort of documentation for arguments. In this case, hard coding is a sensible option.


jebw (approved):
LGTM

PR #2071, opened 2024-12-11
Remove migration environment

ethax-ross:
### Context

Now that we have separated NPQ from ECF we want to decommission the migration environment. We will still have access to a copy of production data via the snapshot database.

- Remove migration environment

Remove all terraform, Rails environment, GitHub workflows, Makefile commands, DB config and readme references to the migration environment as we are tearing it down.

- Switch migration maintenance to sandbox

The migration environment is being removed; we want to retain the maintenance page functionality but switch it to the sandbox environment.

### Guidance for review

We should be able to check out this branch then run `make ci migration terraform-destroy` - once destroyed we should drop the temp commit before merging.

Tested with `terraform plan -destroy`:

```
DOCKER_IMAGE=ghcr.io/dfe-digital/npq-registration:6e016ce6b924732ca9ce22659f5c59dd3e6d6281 make ci migration terraform-plan-destroy

...

Plan: 0 to add, 0 to change, 22 to destroy.
```


cwrw (approved):

Makefile
```diff
@@ -41,7 +41,7 @@ sandbox: production-cluster
 
 migration: production-cluster
```
leandroalemao, line 42: Should we remove this as well as `migration.sh`?

PR #2075, opened 2024-12-13
Remove whenever gem

leandroalemao:
### Context

Ticket: N/A

There was only one job being scheduled via `whenever` gem, which got removed as it was linked to ECF.
The other jobs use `delayed_cron_job` instead.

Remove whenever gem.


mooktakim (approved):
Thanks Leandro

PR #2076, opened 2024-12-13
Document funnding/capping behaviour

ethax-ross:
### Context

We want to document the behaviour around funding/capping as since we introduced `funded_place` there is a fair amount going on when dealing with applications, declarations and eligibility.

- Document funding/capping behaviour

Add documenbtation for funding to detail how application funding, eligibility and capping behaviour works both pre and post 2024 when the `funded_place` attribute was added.

- Clean up rebranded_alternative_courses

I found the switch statement confusing and it makes the behaviour seem more complicated than it actually is.

If the course is a 'rebranded alternative' course return all possible courses otherwise just return the current course.

### DIFF
--- a/docs/funding.md
+++ b/docs/funding.md
@@ -1,0 +1,61 @@
+[< Back to Navigation](../README.md)
+
+# Funding
+
+## Funding eligibility during registration
+
+TODO: to be completed by NPQ reg team
+
+## Accepting an application
+
+When accepting an application we evaluate the `funded_place` attribute passed in the payload against the `funding_cap` of the cohort and `eligible_for_funding` status of the application:
+
+- The `funded_place` attribute in the accept application payload will be evaluated only for cohorts that have a `funding_cap`. 
+- The `funded_place` must be `true` or `false` if the cohort has a `funding_cap`.
+
+When accepting an application with a `funded_place`:
+
+- The `funded_place` can be `true` or `false` if the application is `eligible_for_funding`.
+- The `funded_place` can only be `false` if the application is not `eligible_for_funding`.
+
+## Changing the funded place of an application
+
+Lead providers can change the `funded_place` of an application via the API:
+
+- The application must be `accepted`.
+- The application must be `eligible_for_funding` if changing to `true`.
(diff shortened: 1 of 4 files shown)


docs/funding.md
```diff
@@ -0,0 +1,56 @@
...
+## Funding eligibility during registration
+
+TODO: to be completed by NPQ reg team
+
+## Accepting an application
+
+When accepting an application we evaluate the `funded_place` attribute passed in the payload against the `funding_cap` of the cohort and `eligible_for_funding` status of the application:
+
+- The `funded_place` attribute in the accept application payload will be evaluated only for cohorts that have a `funding_cap`. 
+- An application must be `eligible_for_funding` if the application is being accepted for a `funded_place`.
```
cwrw, line 14: can we elaborate further here with examples, maybe:
- we can set `funded_place` true/false on `eligible_for_funding` true
- and `funded_place` false on `eligible_for_funding` false only

docs/funding.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Previously funded
+
+An application is considered `previously_funded` if:
+
+- Another application exists with an equivalent course (as determined by `rebranded_alternative_courses`).
+  - If the application course is not `npq-additional-support-offer` or `npq-early-headship-coaching-offer` it must be the same course.
+  - If the application course is `npq-additional-support-offer` or `npq-early-headship-coaching-offer` it can be an application with either of these two courses.
+- The other application is `accepted` and `eligible_for_funding`.
+- The `funded_place` of the application is either not set (`nil`) or is `true`.
```
cwrw, line 33: maybe we can say in cohort prior to 2024?

cwrw (commented):
a few minor comments otherwise 👍 thanks for fixing rebranded logic :)

leandroalemao (approved):
Seems it's covering all proposed bullets from the ticket. Not sure what we could add more to this doc. Thanks 👍 😄

mooktakim (approved):
other than Abeers comments, looks good

cwrw (approved):

PR #2079, opened 2024-12-16
[CPDNPQ-2172] Improve financial seeds

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2172

The objective is to make seed data statements more realistic — previously we created a handful more or less at random, after this PR we'll create 2 years of monthly statements in each cohort for each lead provider and course.

We also now create declarations starting in each schedule, with subsequent declaration_types (previously known as "milestones") spread over time, some voids, and a couple of clawbacks on the current output fee statement. Creating these declarations is very slow, so this only happens for the first Lead Provider, otherwise review apps would take too long to spin up.

The latest output fee statement for each cohort / lead provider is also now payable.

Still plenty more improvement on the table, but bringing down the scope hammer here due to time constraints.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -38,5 +38,7 @@ Rails.logger.info("Seeding database")
   "process_statements.rb",
 ].each do |seed_file|
   Rails.logger.info("seeding #{seed_file}")
-  load_base_file(seed_file)
+  ApplicationRecord.transaction do
+    load_base_file(seed_file)
+  end
 end
--- a/db/seeds/base/add_declarations.rb
+++ b/db/seeds/base/add_declarations.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-if Rails.env.in? %w[development review]
-  helpers            = Class.new { include ActiveSupport::Testing::TimeHelpers }.new
-  lead_provider      = LeadProvider.alphabetical.first # it's very slow to do this for all lead providers
-  application_count  = 1
+helpers            = Class.new { include ActiveSupport::Testing::TimeHelpers }.new
+lead_providers     = LeadProvider.alphabetical.limit(2) # it's very slow to do this for all lead providers
+application_count  = 1
 
+lead_providers.each do |lead_provider|
   Schedule.find_each do |schedule|
     schedule.courses.each do |course|
       application_count.times do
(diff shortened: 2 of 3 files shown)


db/seeds/base/process_statements.rb
```diff
@@ -0,0 +1,29 @@
...
+Cohort.all.find_each do |cohort|
+  LeadProvider.all.find_each do |lead_provider|
+    statement_scope  = Statement.where(cohort:, lead_provider:).order(deadline_date: :desc)
+    latest_statement = statement_scope.with_output_fee.where("deadline_date < ?", Time.zone.today).first
+
+    next unless latest_statement
+
+    # Mark past statements (up to latest output fee statement) as paid
+    statement_scope.where("deadline_date < ?", latest_statement.deadline_date).find_each do |statement|
+      Statements::MarkAsPayable.new(statement:).mark
```
jebw, line 12: Can we generate some payable statements ? Or is that difficult because of timings?

db/seeds/base/add_declarations.rb
```diff
@@ -1,106 +1,64 @@
 # frozen_string_literal: true
 
-helpers = Class.new { include ActiveSupport::Testing::TimeHelpers }.new
+if Rails.env.in? %w[development review]
```
jebw, line 3: I don't think we need this in here - the declaration seeds don't need to be handled differently to the contract seeds for example

db/seeds.rb
```diff
@@ -35,6 +35,7 @@ def load_base_file(file)
   "add_contracts.rb",
   "add_declarations.rb",
   "add_api_tokens.rb",
+  "process_statements.rb",
 ].each do |seed_file|
   Rails.logger.info("seeding #{seed_file}")
   load_base_file(seed_file)
```
jebw, line 41: On my old slow laptop, wrapping this in a transaction drops `time rails db:schema:load db:seed`  from 2:57 to 1:43

```ruby
ApplicationRecord.transaction do
  load_base_file(seed_file)
end
```

db/seeds/base/add_declarations.rb
```diff
@@ -1,106 +1,64 @@
...
+  helpers            = Class.new { include ActiveSupport::Testing::TimeHelpers }.new
+  lead_provider      = LeadProvider.alphabetical.first # it's very slow to do this for all lead providers
```
jebw, line 5: Can we change it to do this for 2 lead providers - this allows the testers to do API tests and verify changes to one lead providers data does not impact the other

db/seeds.rb
jebw: not vital but seems like a quick win

jebw (changes requested):
LGTM - some minor queries

db/seeds/base/add_declarations.rb
jebw: Yeh, think we want to be able to tests just as well locally as in a review app and we'll need it to run in staging as well

db/seeds.rb
jebw: just re-run this testing locally and speed up is reproducible

jebw (approved):
LGTM :+1:

PR #2086, opened 2024-12-17
[CPDNPQ-2394] Don't allow duplicate statements

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2394

It’s currently possible to create more than one statement per lead provider / cohort / year / month

This is incorrect, there should be a maximum of one statement per lead provider / cohort / year / month

This currently affects seed data; although #2079  will correct the seeds to no longer create duplicate statements, it should be technically impossible to do so in any circumstances

#### 1: Check that no production data violates this

In a production console just now:

```
irb(main):001> Statement.find_each.map { [_1.cohort_id, _1.lead_provider_id, _1.year, _1.month] }.tally.values.all?(1)
=> true
```

I think this means we're all good, unless there are any other checks I haven't considered?

#### 2: Introduce necessary model validation and database constraint(s) to make this impossible in future

See Pull Request — one unique index, one `uniqueness` validation 🔒

### DIFF
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -20,11 +20,18 @@ class Statement < ApplicationRecord
               only_integer: true,
               message: "Month must be a number between 1 and 12",
             }
+
   validates :year,
             numericality: {
               in: 2020..2050,
               only_integer: true,
               message: "Year must be a 4 digit number",
+            }
+
+  validates :lead_provider_id,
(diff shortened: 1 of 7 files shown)


alkesh (approved):

jebw (approved):
LGTM

PR #2090, opened 2024-12-18
2326 update statements

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2326

### Introduces FinancialChangeLog

FinancialChangeLog is an easy way to track all financial related data changes.

While our app has already papertrail, the changing financial data through different CSV files or using different
data sources need full transparency.
It provides one place where we can easy track:
* which process or ticket or dataset changed given data.
* which records and attributes of those records were changed
* if record is deleted, the log will still stay in database
* unlike per model versions, log can be easily searchable

While NPQ is not fully a financial app, the FinancialChangeLog replicates a ledger functionality that will make tracing any changes much easier.


slawosz:
I see my Rubymine is doing some silly things with newline characters on the rake files, working on it.

app/services/one_off/create_or_update_statements.rb
```diff
@@ -0,0 +1,84 @@
...
+          LeadProvider.where.not(name: "National Institute of Teaching").find_each do |lead_provider|
+            year, month = parse_date(row["name"])
+            statements = Statement.where(year:, month:, cohort: @cohort, lead_provider: lead_provider)
+            return ArgumentError if statements.count > 1
+
+            statement = statements.first
+            if statement
+              statement.update!(output_fee: output_fee(row["output_fee"]))
+              FinancialChangeLog.log!(description: FinancialChangeLog::ONE_OFF_2326, data: { updated_statement_id: statement.id })
+            else
```
alkesh, line 42:
```suggestion
              statement.update!(output_fee: output_fee(row["output_fee"]))
              FinancialChangeLog.log!(description: FinancialChangeLog::ONE_OFF_2326, data: { updated_statement_id: statement.id, , changes: statement.changes })
            else
```

things this change does:
- logs the data being changed
- only logs if the the output fee actually changes

wdyt?

spec/services/one_off/create_or_update_statements_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      before do
+        LeadProvider.where.not(name: "Ambition Institute").map(&:destroy)
+      end
+
+      it "updates a statement" do
+        expect {
+          OneOff::CreateOrUpdateStatements.new.call(cohort_year: 2021, csv_path:)
+        }.to change { statement_1.reload.output_fee }.from(true).to(false)
+      end
```
alkesh, line 86: missing a spec for "it creates log record" here

spec/models/financial_change_log_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe FinancialChangeLog, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
jebw, line 4: Something in here just to check the roundtripping of the json data would be great

app/services/one_off/create_or_update_statements.rb
jebw: great suggestion

app/models/financial_change_log.rb
```diff
@@ -0,0 +1,19 @@
...
+# * which process or ticket or dataset changed given data.
+# * which records and attributes of those records were changed
+# * if record is deleted, the log will still stay in database
+# * unlike per model versions, log can be easily searchable
+#
+# While NPQ is not fully a financial app, the FinancialChangeLog replicate to some extend a ledger functionality often being found in
+# proper financial apps.
+class FinancialChangeLog < ApplicationRecord
+  ONE_OFF_2326 = "OneOff 2326".freeze
+
```
jebw, line 15: can we add some minimal validations, just check description and data changes are present - don't need to worry about error messages

jebw (changes requested):
Hey, looks great - can some minimal specs be added to the new data model - just checking the json data roundtrips from the db correctly

Also validating its been set in the model would be useful

app/services/one_off/create_or_update_statements.rb
slawosz: Its not working as expected, we can investigate later - possibly `changes` gets reset when saved. With ticket number and the file available in the ticket we can easily track the change.

alkesh: yeah - `changes` and `changed` are reset when saved.
you would have to update the attribute, then check for `changed?`, and then log and save.

slawosz: Thats an idea - as we are less likely to miss the important changes. How about we can do it in the next iteration?

app/services/one_off/create_or_update_statements.rb
```diff
@@ -0,0 +1,84 @@
...
+      csv_file = CSV.read(csv_path, headers: true)
+
+      ActiveRecord::Base.transaction do
+        @cohort = Cohort.find_by!(start_year: cohort_year)
+
+        csv_file.each do |row|
+          LeadProvider.where.not(name: "National Institute of Teaching").find_each do |lead_provider|
+            year, month = parse_date(row["name"])
+            statements = Statement.where(year:, month:, cohort: @cohort, lead_provider: lead_provider)
+            return ArgumentError if statements.count > 1
```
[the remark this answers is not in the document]
slawosz: Stupid me. Removing anyway as we have unique index on statements now.

alkesh (approved):

spec/services/one_off/create_or_update_statements_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+          OneOff::CreateOrUpdateStatements.new.call(cohort_year:, csv_path:)
+        }.to change { statement_1.reload.output_fee }.from(true).to(false)
+      end
+
+      it "creates log records" do
+        OneOff::CreateOrUpdateStatements.new.call(cohort_year:, csv_path:)
+
+        log = FinancialChangeLog.first
+        expect(log.operation_description).to eq("OneOff 2326")
+        expect(log.data_changes).to eq({ "updated_statement_id" => statement_1.id })
```
alkesh, line 83:
```suggestion
        expect(log.data_changes).to eq({ "changes" => { "output_fee" => [true, false] }, "updated_statement_id" => statement_1.id })```

PR #2095, opened 2024-12-20
[CPDNPQ-2419] Fix some flakey tests

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2419

tl;dr: fix some flakey tests — think I got them all, but some of the failures are veeeery intermittent and can't be reliably reproduced even with the same seed! So we'll have to wait and see.

### DIFF
--- a/spec/services/assurance_reports/query_spec.rb
+++ b/spec/services/assurance_reports/query_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe AssuranceReports::Query do
   let(:assurance_report)    { query.declarations.first }
   let(:lead_provider)       { create(:lead_provider) }
   let(:other_lead_provider) { create(:lead_provider) }
-  let(:statement)           { create(:statement, lead_provider:) }
+  let(:statement)           { create(:statement, :next_period, lead_provider:) }
   let(:training_status)     { :active }
 
   let :declaration do


spec/features/npq_separation/admin/finance/voided_declarations_spec.rb
```diff
@@ -20,7 +20,7 @@
     expect(page).to have_css("td:nth-child(1)", text: voided_declaration.id)
     expect(page).to have_css("td:nth-child(2)", text: voided_declaration.user.id)
     expect(page).to have_css("td:nth-child(3)", text: voided_declaration.declaration_type)
-    expect(page).to have_css("td:nth-child(4)", text: voided_declaration.course.name)
+    expect(page).to have_css("td:nth-child(4)", text: voided_declaration.reload.course.name)
```
jebw, line 23: If the declaration is needing to be reloaded, presumably because data may have changed, shouldn't the first usage of the declaration on line 20 be the one doing the reload?

jebw (commented):
LGTM - only query is if `reload`ing only on later usage leaves the earlier lines still being flaky?

alkesh (approved):

PR #2108, opened 2025-01-07
[CPDNPQ-2407] Track user in papertrail records

jebw:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2407

We wish to know who changed record, as well as what was changed.

1. Upgraded `paper_trail` to latest version
2. Changed to include current user when recording updates to records via `paper_trail`


config/initializers/paper_trail.rb
```diff
@@ -1 +1,7 @@
 PaperTrail.serializer = PaperTrail::Serializers::JSON
+
+Rails.application.configure do
+  console do
+    PaperTrail.request.whodunnit = "Rails Console"
```
slawosz, line 5: In `ECF` the rails console is asking person for a name, I think it was a good practice.

slawosz (approved):

config/initializers/paper_trail.rb
jebw: Yeah, good call - I'd debated that approach, I'll add that in :+1:

PR #2109, opened 2025-01-07
[CPDNPQ-2442] Default to running rails console from workers

jebw:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2442

Currently the default `Makefile` commands will run consoles on the web nodes which can impact a running web request if the console usage triggers an out of memory scenario.

A secondary concern is the console command by default provides read-write and in production this will often only need to be read-only for investigation purposes

1. Run all commands from worker nodes instead
2. Provide an alternative to ssh into a web node
3. Changed the default `aks-console` to use rails' `--sandbox` parameter
4. Added a new `aks-rw-console` command to run without the `--sandbox` parameter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,19 +174,23 @@ install-konduit: ## Install the konduit script, for accessing backend services
 
 aks-console: get-cluster-credentials
 	$(if $(PULL_REQUEST_NUMBER), $(eval export APP_ID=review-$(PULL_REQUEST_NUMBER)) , $(eval export APP_ID=$(CONFIG_LONG)))
-	kubectl -n ${NAMESPACE} exec -ti --tty deployment/npq-registration-${APP_ID}-web -- /bin/sh -c "cd /app && bundle exec rails c"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -174,19 +174,23 @@ install-konduit: ## Install the konduit script, for accessing backend services
 
 aks-console: get-cluster-credentials
 	$(if $(PULL_REQUEST_NUMBER), $(eval export APP_ID=review-$(PULL_REQUEST_NUMBER)) , $(eval export APP_ID=$(CONFIG_LONG)))
-	kubectl -n ${NAMESPACE} exec -ti --tty deployment/npq-registration-${APP_ID}-web -- /bin/sh -c "cd /app && bundle exec rails c"
+	kubectl -n ${NAMESPACE} exec -ti --tty deployment/npq-registration-${APP_ID}-worker -- /bin/sh -c "cd /app && bundle exec rails c --sandbox"
+
+aks-rw-console: get-cluster-credentials
```
alkesh, line 179: worth updating `docs/connecting-to-azure.md` to mention these new commands

jebw: Done :+1:

alkesh (approved):

PR #2110, opened 2025-01-08
[CPDNPQ-2363] one off rake task to create contract

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2363

EHCO contract missing for LLSE in 2023 cohort

new rake task - as existing `update_contracts_2350.rake` rake task only updates existing contract values

### DIFF
--- a/lib/tasks/one_off/create_contract.rake
+++ b/lib/tasks/one_off/create_contract.rake
@@ -1,0 +1,54 @@
+COHORT_YEAR = 2023
+STATEMENT_YEAR = 2025
+STATEMENT_MONTH = 1
+LEAD_PROVIDER_NAME = "LLSE".freeze
+
+namespace :one_off do
+  # Example usage (dry run):
+  # bundle exec rake 'one_off:create_contract[tmp/contract.csv]'
+  #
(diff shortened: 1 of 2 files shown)


lib/tasks/one_off/create_contract.rake
```diff
@@ -0,0 +1,54 @@
+COHORT_YEAR = 2023
+STATEMENT_YEAR = 2025
+STATEMENT_MONTH = 1
+LEAD_PROVIDER_NAME = "LLSE".freeze
```
jebw, line 4: This will end up in the global namespace for anything reading the rake task file

I think its fine for there to either be constants or variables in the rake task

jebw (approved):
LGTM but would be nice to move the constants out - if nothing else these will end up defined when then test suite runs. Must admit I'm not sure what conditions dictate when a rake file is read

lib/tasks/one_off/create_contract.rake
alkesh: that's a very good point - I've wrapped the rake task definition in a class - so those constants should not leak into the global namespace now.

PR #2112, opened 2025-01-08
Validate declaration_date only when its changed

slawosz:
`declaration_date` does not need to be validated when is not being changed.
https://dfedigital.atlassian.net/browse/CPDNPQ-2432

### DIFF
--- a/spec/models/declaration_spec.rb
+++ b/spec/models/declaration_spec.rb
@@ -54,11 +54,7 @@ RSpec.describe Declaration, type: :model do
 
         let(:application) { create(:application, :accepted, user:, course:) }
 
-        subject do
-          d = build(:declaration, course:, user:, application:, declaration_date:)
-          d.save!(validate: false)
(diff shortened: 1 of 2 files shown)


spec/models/declaration_spec.rb
```diff
@@ -37,17 +37,42 @@
...
+        let(:user) { create(:user) }
+        let(:course) { create(:course) }
+
+        let(:application) { create(:application, :accepted, user:, course:) }
+
+        subject do
+          d = build(:declaration, course:, user:, application:, declaration_date:)
+          d.save!(validate: false)
+          d
+        end
```
alkesh, line 61:
```suggestion
        subject { build(:declaration, course:, user:, application:, declaration_date:).tap { |d| d.save!(validate: false) } }
```

alkesh (approved):

jebw (approved):
LGTM

PR #2113, opened 2025-01-09
[CPDNPQ-2314] rake task to merge duplicate users

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2314

Need a rake task to deduplicate users:
- move applications from duplicate user to user to keep
- move uid over if neccessary
- archive duplicate user

The existing `Users::ArchiveByEmail` was designed only to be used when there was an email clash when a user logs in.
This PR moves the common underlying functionality into a new service `Users::MergeAndArchive` which is used both by the log in case, and the new rake task for deduplicating users (`one_off:merge_user_into_another`)


app/services/users/archiver.rb
```diff
@@ -10,14 +10,12 @@ class Archiver
     def archive!
       raise ArgumentError, "User already archived" if user.archived?
 
-      ApplicationRecord.transaction do
```
alkesh, line 13: This Archiver was previously called by the separation migration code - but this is no longer the case,
it is only called by `Users::MergeAndArchive` now, which wraps it in its own transaction.

lib/tasks/one_off/merge_user_into_another.rake
```diff
@@ -0,0 +1,30 @@
+namespace :one_off do
```
slawosz, line 1: It is not technically a one off as this script will be runned few times I guess?

spec/lib/tasks/one_off/merge_user_into_another_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  let!(:user_to_merge_accepted_application) { create(:application, :accepted, user: user_to_merge) }
+  let!(:existing_participant_id_change) { create(:participant_id_change, user: user_to_merge) }
+  let(:user_to_merge) { create(:user, :with_get_an_identity_id) }
+  let(:user_to_keep) { create(:user) }
+
+  after do
+    Rake::Task["one_off:merge_user_into_another"].reenable
+  end
+
+  context "when dry run false" do
```
slawosz, line 18: Wouldn't it better to test actuall funcionality in service class?

lib/tasks/one_off/merge_user_into_another.rake
alkesh: 👍🏽 good point - I'll move this out of one_off

slawosz (approved):

PR #2115, opened 2025-01-10
[CPDNPQ-2225] qualifications API for TRS - part of DQT decommissioning

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2225

- new qualifications API - returns a combination of passed participant outcomes from NPQ and the legacy participant outcomes loaded from the DQT data extract.
- added `scope` to APIToken - this is either `lead_provider` or `teacher_record_service` - and a token from one scope cannot access the APIs from the other scope
- new one off rake task to load legacy participant outcomes - which is from an extract from DQT
- new rake task to create a teacher_record_service APIToken
- document describing the API (`docs/qualifications_api.md`)


config/routes.rb
```diff
@@ -208,6 +208,12 @@
 
       resources :statements, only: %i[index show], param: :ecf_id
     end
+
+    namespace :teacher_record_service, path: "teacher-record-service", defaults: { format: :json } do
+      namespace :v1 do
+        resources :qualifications, path: "npq-qualifications", only: %i[show], param: :trn
```
jebw, line 214: Do we need to namespace the `path` with a `npq-` prefix? 

Its a new API against our npq specific host, some of the existing APIs are like it to maintain compatibility with old lead provider code carried over from ECF integration but thats not the case here?

db/migrate/20241104164154_add_scope_to_api_tokens.rb
```diff
@@ -0,0 +1,13 @@
+class AddScopeToAPITokens < ActiveRecord::Migration[7.1]
+  def change
+    add_column :api_tokens, :scope, :string, default: "lead_provider", null: false
+
+    reversible do |direction|
+      direction.up do
+        execute <<-SQL
+          UPDATE api_tokens SET scope = 'lead_provider';
+        SQL
```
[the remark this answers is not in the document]
jebw: Yeah, that would be my expectation as well

db/schema.rb
```diff
@@ -70,11 +70,12 @@
   end
 
   create_table "api_tokens", force: :cascade do |t|
-    t.bigint "lead_provider_id", null: false
+    t.bigint "lead_provider_id"
```
jebw, line 73: So I'm not following here - I'm seeing a schema change to drop the `null: false` but no corresponding migration to achieve that? Is something missing here

You can also add a DB constraint in place of the `, null: false` requirement to require a `lead_provider_id IS NOT NULL WHERE scope = "lead_provider"`

spec/factories/api_tokens.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :api_token do
+    lead_provider { create(:lead_provider) }
+    hashed_token { Devise.token_generator.digest(APIToken, :hashed_token, "token") }
+
+    trait :teacher_record_service do
+      scope { APIToken.scopes[:teacher_record_service] }
+      hashed_token { Devise.token_generator.digest(APIToken, :hashed_token, "trs_token") }
+    end
```
jebw, line 9: The developer needs to be able to either predict the raw token or control it, moving it to a transient block allows for generation of multiple tokens of the same scope in the future?

```ruby
    transient do
      raw_token { "token" }
    end

    lead_provider { create(:lead_provider) }
    hashed_token { Devise.token_generator.digest(APIToken, :hashed_token, raw_token) }

    trait :teacher_record_service do
      raw_token { "trs_token" }

      lead_provider { nil }
      scope { APIToken.scopes[:teacher_record_service] }
    end
```

spec/rails_helper.rb
```diff
@@ -125,6 +125,7 @@
     Course::IDENTIFIERS.each do |identifier|
       FactoryBot.create(identifier)
     end
+    Rails.application.load_tasks
```
jebw, line 128: Can this get scoped to just rake specs, so they're not loaded when the other specs are individually run?

jebw (commented):
LGTM - couple of comments around schema changes and API path

The actual API format makes me wince a little since this is what JSON:API has `include` and entity relationships for _but_ it is following the pattern of our other APIs and does make it simpler to consume for someone just parsing as raw JSON (instead of using a JSON:API client)

db/schema.rb
alkesh: that is odd - not sure how that migration got lost - I'll add it back in.
Good idea about using a check constraint.

jebw (approved):
LGTM 

I've tested the migration both 
* loading schema from `main`, switching to branch and running db:migrate
* loading schema from branch

AFAICT it constrains correctly in all scenarios now, using both approaches above :+1:

PR #2116, opened 2025-01-13
[CPDNPQ-2210] Statement + contract setup

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2210

Ability to create statements (including contracts) for a cohort

### DIFF
--- a/db/migrate/20250109205930_set_statement_default_reconcile_amount.rb
+++ b/db/migrate/20250109205930_set_statement_default_reconcile_amount.rb
@@ -1,5 +1,0 @@
-class SetStatementDefaultReconcileAmount < ActiveRecord::Migration[7.1]
-  def change
-    change_column_default :statements, :reconcile_amount, from: nil, to: 0
-  end
-end
--- a/app/services/statements/bulk_creator.rb
+++ b/app/services/statements/bulk_creator.rb
@@ -77,7 +77,7 @@ module Statements
     def statement_attributes_for(statement_row, contract_row)
       lead_provider = lead_provider_for(contract_row)
 
-      statement_row.attributes.merge(cohort:, lead_provider:)
+      statement_row.attributes.merge(cohort:, lead_provider:, reconcile_amount: 0)
     end
 
     def contract_attributes_for(contract_row)
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_01_09_205930) do
+ActiveRecord::Schema[7.1].define(version: 2025_01_06_140920) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "citext"
@@ -542,7 +542,7 @@ ActiveRecord::Schema[7.1].define(version: 2025_01_09_205930) do
     t.bigint "cohort_id", null: false
     t.bigint "lead_provider_id", null: false
     t.datetime "marked_as_paid_at"
-    t.decimal "reconcile_amount", precision: 8, scale: 2, default: "0.0"
+    t.decimal "reconcile_amount", precision: 8, scale: 2
     t.enum "state", default: "open", null: false, enum_type: "statement_states"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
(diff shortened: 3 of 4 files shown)


app/controllers/npq_separation/admin/statements_controller.rb
```diff
@@ -0,0 +1,84 @@
+class NpqSeparation::Admin::StatementsController < NpqSeparation::AdminController
```
slawosz, line 1: How about inheriting from `SuperAdminController`?

spec/features/npq_separation/admin/creating_statements_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      expect {
+        click_on "Create statements"
+        expect(page).to have_text("6 statements created successfully")
+      }.to change(Statement, :count).by(6)
+          .and change(Contract, :count).by(9)
+          .and change(ContractTemplate, :count).by(3)
+    end
+
+    scenario "downloading examples" do
+      page.driver.browser.download_path = download_path
```
jebw, line 70: Probably makes sense to just set this in `rails_helper.rb` for all feature specs?

db/migrate/20250109205930_set_statement_default_reconcile_amount.rb
```diff
@@ -0,0 +1,5 @@
+class SetStatementDefaultReconcileAmount < ActiveRecord::Migration[7.1]
+  def change
+    change_column_default :statements, :reconcile_amount, from: nil, to: 0
```
jebw, line 3: Pretty against this one - we shouldn't have a 'default amount' for a number we should be calculating - if its 0 then some code should have calculated its zero and set it?

`nil` shows nothing has calculated the amount

config/initializers/custom_types.rb
```diff
@@ -0,0 +1,22 @@
+class StrictDecimal < ActiveModel::Type::Integer
```
jebw, line 1: Love this :raised_hands:

app/services/statements/bulk_creator/contract_row.rb
```diff
@@ -0,0 +1,44 @@
+module Statements
+  class BulkCreator
```
jebw, line 2: Historically I've had issues with multiple files defining the same class (which is what this does)

If this file is loaded before `bulk_creator.rb` then you'll end up with an empty instance of `BulkCreator` class defined and it'll never try to autoload the `bulk_creator.rb` file

2 common approaches are either

1) Use a pluralized name as a module - `module BulkCreators`
2) Less pleasant, replace with `BulkCreator.class_eval do` - this will force autoloading the original class, then reopen it and define this `ContractRow` class within it

app/controllers/npq_separation/admin/statements_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      contracts_csv_id: bulk_creator_params[:contracts_csv_id],
+    )
+
+    @statements = @service.call(dry_run:)
+
+    if @service.errors.any?
+      render :new
+    elsif dry_run
+      set_preview
+      render
```
jebw, line 21: isn't this redundant? it'll just `render :create` by default anyway?

app/controllers/npq_separation/admin/statements_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+  end
+
+  def show
+    row_class = {
+      "statements" => Statements::BulkCreator::StatementRow,
+      "contracts" => Statements::BulkCreator::ContractRow,
+    }.fetch(params[:id])
+
+    render plain: row_class.example_csv.lines.first, content_type: "text/csv", filename: "#{params[:id]}.csv"
```
jebw, line 34: I'd expect csv would already be registered in mime types (so `content_type: :csv`) but if its not lets add it there

jebw: Thinking on this shouldn't this (and other similar lines) be `send_data` ?

jebw (changes requested):
Looks great, I'd second @slawosz comment re: not reimplementing the SuperAdmin checks

Only point that's not a 'comment' is the schema change for default value on a statement seems wrong to me?

app/controllers/npq_separation/admin/statements_controller.rb
jebw: Hmm, maybe split the super admin checks into a module and share between classes, I'm keen for us not to be repeating auth code - can't see a problem with whats there _but_ its better to have a single implementation

jebw (approved):
LGTM :raised_hands:

PR #2117, opened 2025-01-13
CPDNPQ-2461 fix schedule calculations

jebw:
### Context

Ticket: [CPDNPQ-2461](https://dfedigital.atlassian.net/browse/CPDNPQ-2461)

Currently lead providers are unable to accept applications because our schedule calculation does not return any schedule for today for Leadership and Specialist courses

1. The `CourseGroup::Leadership` calculator is updated to special case the extended autumn 2024 schedule
2. The `CourseGroup::Specialist` calculator is updated to special case the extended autumn 2024 schedule

[CPDNPQ-2461]: https://dfedigital.atlassian.net/browse/CPDNPQ-2461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/course_groups/leadership_spec.rb
+++ b/spec/services/course_groups/leadership_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe CourseGroups::Leadership do
       end
     end
 
-    context "when date is between December of cohort start year and April of 2026" do
+    context "when date is between December and April of cohort start year" do
       let(:cohort) { create(:cohort, start_year: 2025) }
 
       it "returns Spring schedule" do
--- a/spec/services/course_groups/leadership_spec.rb
+++ b/spec/services/course_groups/leadership_spec.rb
@@ -19,7 +19,9 @@ RSpec.describe CourseGroups::Leadership do
       end
     end
 
-    context "when date is between December of cohort start year and April of the next year" do
+    context "when date is between December of cohort start year and April of 2026" do
+      let(:cohort) { create(:cohort, start_year: 2025) }
+
       it "returns Spring schedule" do
         travel_to Date.new(cohort.start_year, 12, 26) do
           expect(subject.schedule).to eq(spring_schedule)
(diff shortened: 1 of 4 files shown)


spec/services/course_groups/leadership_spec.rb
```diff
@@ -19,7 +19,9 @@
       end
     end
 
-    context "when date is between December of cohort start year and April of the next year" do
+    context "when date is between December of cohort start year and April of 2026" do
+      let(:cohort) { create(:cohort, start_year: 2025) }
```
alkesh, line 23: The context `when` does not match the `let`.
Should the context be `when the cohort is 2025`?
or should there be a before block with the `travel_to` in it?
Not a blocker for merging, just a nice to have.

spec/services/course_groups/specialist_spec.rb
```diff
@@ -19,7 +19,9 @@
       end
     end
 
-    context "when date is between December of cohort start year and April of the next year" do
+    context "when date is between December of cohort start year and April of 2026" do
+      let(:cohort) { create(:cohort, start_year: 2025) }
```
alkesh, line 23: need the `travel_to` in a before block here?
(nice to have)

alkesh (approved):

spec/services/course_groups/leadership_spec.rb
jebw: I'd argue no - this is for the 2025 cohort but spring 2026 schedule _that being said_ I'll put the names back to approximately how they were to avoid confusion

spec/services/course_groups/specialist_spec.rb
jebw: I'd probably argue yes but without changing the whole file it will be inconsistent so for simplicity for now I'll leave it as is

PR #2127, opened 2025-01-20
Display stalled statements list in financial console

slawosz:
image
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2440


app/models/statement.rb
```diff
@@ -1,4 +1,6 @@
 class Statement < ApplicationRecord
+  AUTHORISATION_GRACE_TIME = 15 * 60
```
jebw, line 2: This can use rails' period utilities, ie `15.minutes`

app/models/statement.rb
```diff
@@ -41,6 +43,7 @@ class Statement < ApplicationRecord
   scope :unpaid, -> { with_state(%w[open payable]) }
   scope :paid, -> { with_state("paid") }
   scope :next_output_fee_statements, -> { with_state("open").with_output_fee.order(:deadline_date).where("deadline_date >= ?", Date.current) }
+  scope :with_delayed_authorisations, -> { with_state("payable").where("marked_as_paid_at < ?", AUTHORISATION_GRACE_TIME.seconds.ago) }
```
jebw, line 46: This would then change to `AUTHORISATION_GRACE_TIME.ago`

app/controllers/npq_separation/admin/finance/statements/stale_controller.rb
```diff
@@ -0,0 +1,5 @@
+class NpqSeparation::Admin::Finance::Statements::StaleController < NpqSeparation::AdminController
+  def index
+    @statements = Statement.with_delayed_authorisations
```
jebw, line 3: Is there a reason not to use Statements::Query like `PaidController` does

Also I'm maybe missing something but you've got the pagination helper in the view but nothing seems to be passing the page into the above scope?

jebw (changes requested):
Looks great, really nice addition :raised_hands: I've made a couple of minor points around the scope in the controller inline.

Another request is to add a request spec for the controller - it can be pretty minimal, just verify 200 response for authorised and 401 for unauthorised but it does need to a request spec to be checking the authorization

https://github.com/DFE-Digital/npq-registration/blob/main/spec/requests/npq_separation/admin/schedules_controller_spec.rb#L124-L139

jebw (commented):
Looks great, but linter and some of the specs are failing?

jebw (approved):
LGTM :raised_hands:

jebw:
@slawosz this has been sat around for a long time - presumably its no longer needed and can be closed?

PR #2132, opened 2025-01-21
Revert "[CPDNPQ-2357] Make staging auth operate like production"

jebw:
Reverts DFE-Digital/npq-registration#2089

Put staging back to how it was


alkesh (approved):

slawosz (approved):
Thanks :)

PR #2139, opened 2025-01-22
[BAU] Added an initial Architecture Decision Record introducing usage of ADRs

jebw:
### Context

Ticket: BAU

We have a lot of existing code which is contains many decisions, the majority are not documented

1. Add an initial ADR to introduce usage of Architecture Decision Records

### DIFF
--- a/docs/adr/00001-introduce-architecture-decision-records.md
+++ b/docs/adr/00001-introduce-architecture-decision-records.md
@@ -1,4 +1,4 @@
-# Record architecture decisions
+# 1. Record architecture decisions
 
 Date: 2025-01-22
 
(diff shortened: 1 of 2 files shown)


alkesh (approved):

docs/adr/00001-introduce-architecture-decision-records.md
alkesh, line 1: `adr tools` uses 4 digits for the ID.
so the first ADR will be `0001`

alkesh: oh - another thing with `adr tools`.
I think there needs to be a file `.adr-dir` containing `docs/adr` (this is the result of running `adr init docs/adr`

PR #2149, opened 2025-01-24
[BAU] Upgrade GovUK Frontend

jebw:
### Context

Ticket: BAU

We should stay current with the GovUK frontend

1. Upgrade the npm govuk-frontend package
2. Upgrade the ruby form builder to a matching version
3. Upgrade the ruby components to a matching version


peteryates (approved):
Nice. Surprised the legacy syntax worked this long

PR #2165, opened 2025-01-29
[CPDNPQ-2524] Fix search for applications with no school

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2524

With production data we noticed that the edge case applications index is missing lots of Referred By Return To Teaching Adviser applications — there are 283 in the database, but only 4 are showing

Turns out this is because ApplicationsSearch uses an INNER JOIN rather than a LEFT JOIN i.e. discards rows without a `school_id` foreign key

This PR converts the join to a LEFT JOIN so that rows with no school are included

### DIFF
--- a/app/services/admin_service/applications_search.rb
+++ b/app/services/admin_service/applications_search.rb
@@ -22,6 +22,6 @@ class AdminService::ApplicationsSearch
 private
 
   def default_scope
-    Application.left_joins(:school, :user).includes(:course, :lead_provider, :school, :user).order(id: :asc)
+    Application.left_joins(:school).joins(:user).includes(:course, :lead_provider, :school, :user).order(id: :asc)
   end
 end


alkesh (approved):

spec/lib/services/admin_service/applications_search_spec.rb
alkesh, line 1: minor comment - but can the `spec/lib/services/admin_service` folder be moved out of lib, and into`spec/services/admin_service`?

app/services/admin_service/applications_search.rb
```diff
@@ -22,6 +22,6 @@ def call
 private
 
   def default_scope
-    Application.joins(:school, :user).includes(:course, :lead_provider, :school, :user).order(id: :asc)
+    Application.left_joins(:school, :user).includes(:course, :lead_provider, :school, :user).order(id: :asc)
```
jebw, line 25: Hmm - I think you just want to drop the school from the joins? From the code it doesn't look like they can search by school anyway?

spec/lib/services/admin_service/applications_search_spec.rb
jebw: @alkesh maybe revive your PR to move all the specs around?

alkesh: good idea 👍🏽 
https://github.com/DFE-Digital/npq-registration/pull/2064

alkesh (approved):

PR #2181, opened 2025-02-05
[CPDNPQ-2485] content security policy updates

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2485

Current CSP needs improvements

* remove `unsafe-inline` and `unsafe-eval` for scripts
* replace use of `<script>` with `nonced_javascript_tag` in the views
* remove `unsafe-inline` for styles
* change `report_uri` to report to sentry
* remove CSP reports controller and route
* fix javascript console error that was only occurring on separation admin (every page) because we hadn't included `layouts/shared/govuk_javascript`


jebw (commented):
Looks really great, only query is around ENV var name

(and possible lack of variable assignment but I could be wrong on that)

jebw (approved):
TIL I learned about vars getting assign in `if` clauses that are false :thinking:

PR #2183, opened 2025-02-05
Enable bullet in test+review and fix N+1

jebw:
### Context

Ticket: BAU

We want to know when our pages are not optimised automated testing (ie the `test` environment) and manual testing during production review (ie `review` environment) are our best options

1. Remove 'ignores' from bullet in test environment
2. Enable bullet and make it hard fail in review environment
3. Fix an N+1 which was discovered as part of this


jebw:
terraform has tied itself in knots. closing and reopening

PR #2188, opened 2025-02-06
[CPDNPQ-2308] send email when eligible_for_funding changes from false to true

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2308

send email when application `eligible_for_funding` changes from `false` to `true`
don't send an email when it changes from `true` to `false`, or when it doesn't change.

### DIFF
--- a/app/services/applications/change_funding_eligibility.rb
+++ b/app/services/applications/change_funding_eligibility.rb
@@ -29,7 +29,11 @@ module Applications
       funding_eligiblity_status_code =
         (eligible_for_funding ? :marked_funded_by_policy : :marked_ineligible_by_policy)
 
-      application.update!(eligible_for_funding:, funding_eligiblity_status_code:)
+      application.assign_attributes(eligible_for_funding: eligible_for_funding, funding_eligiblity_status_code: funding_eligiblity_status_code)
+      eligible_for_funding_change = application.changes["eligible_for_funding"]
+      application.save!.tap do
+        send_eligible_for_funding_email if eligible_for_funding_change == [false, true]
+      end
     end
 
   private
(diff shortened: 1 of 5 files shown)


app/services/applications/change_funding_eligibility.rb
```diff
@@ -29,7 +29,11 @@ def change_funding_eligibility
       funding_eligiblity_status_code =
         (eligible_for_funding ? :marked_funded_by_policy : :marked_ineligible_by_policy)
 
-      application.update!(eligible_for_funding:, funding_eligiblity_status_code:)
+      application.assign_attributes(eligible_for_funding: eligible_for_funding, funding_eligiblity_status_code: funding_eligiblity_status_code)
```
jebw, line 32: Presumably you can still use the hash shorthand here (as was used when this was (`#update!`) ?

app/services/applications/change_funding_eligibility.rb
```diff
@@ -29,7 +29,11 @@ def change_funding_eligibility
...
+      eligible_for_funding_change = application.changes["eligible_for_funding"]
+      application.save!.tap do
+        send_eligible_for_funding_email if eligible_for_funding_change == [false, true]
```
jebw, line 35: Can you use the `.saved_changes` to determine the change instead? seems simpler?

app/services/applications/change_funding_eligibility.rb
```diff
@@ -49,5 +53,15 @@ def validate_funding_eligiblity_status_with_funded_place
...
     end
+
+    def send_eligible_for_funding_email
+      ApplicationFundingEligibilityMailer.eligible_for_funding_mail(
+        to: application.user.email,
+        full_name: application.user.full_name,
+        provider_name: application.lead_provider.name,
+        course_name: application.course.name,
+        ecf_id: application.ecf_id,
+      ).deliver_now
```
jebw, line 64: `.deliver_later` ? That way if there's flakiness connecting to Notify this will still get delivered rather than the action failing?

jebw (approved):
Looks good, you can probably use `.saved_changes` to simplify the code a bit

And I think it should probably be sending the email async ? That may need some tweaks to your specs though?

app/services/applications/change_funding_eligibility.rb
alkesh: yes - not sure why rubocop didn't pick this up.
I've switched back to the original `update!` method, as I can use `saved_changes` (thanks for the suggestion).

PR #2206, opened 2025-02-10
[CPDNPQ-2605] Remove invalid EHCO provider options

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2605

After choosing the EHCO NPQ, applicants are taken to a screen to specify their provider

The list of providers is hard-coded and hasn’t been updated to reflect a change in which providers offer this course

Ultimately these lists should not be hard-coded, but since the service is currently open, to solve this quickly we should remove invalid providers from this list


jebw (approved):
LGTM

alkesh (approved):

PR #2214, opened 2025-02-11
[CPDNPQ-2369] validate declaration_type at both model and service layer

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2369

Creating a declaration via the API with an invalid `declaration_type` causes a 500 error

The issue is that a query that uses an invalid `declaration_type` will raise the error:
 `ActiveRecord::StatementInvalid: PG::InvalidTextRepresentation: ERROR:  invalid input value for enum declaration_types`.
This is happening in `Declarations::Create#existing_declaration` which is used in the `output_fee_statement_available` validation.

Rather than catching the `ActiveRecord::StatementInvalid` error, I think it's cleaner to just add an inclusion validation on the `Declarations::Create` service.

The validation added to the Declaration model won't be used in this scenario, but I've added it for thoroughness.

### DIFF
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -80,12 +80,12 @@ class Declaration < ApplicationRecord
     end
   end
 
-  enum declaration_type: {
+  enum :declaration_type, {
     started: "started",
     "retained-1": "retained-1",
     "retained-2": "retained-2",
     completed: "completed",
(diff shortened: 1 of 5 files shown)


jebw (approved):
LGTM :raised_hands:

PR #2220, opened 2025-02-13
[CPDNPQ-2480] Admin console - change lead provider

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2480

* a new link next to the Lead Provider name: 'Transfer'
* a page showing all Lead Providers
* validation to ensure a LP is selected

I've used a `show` page instead of a `new` page - our admin pages seem to mostly use `new`.
Advantages of using 'show' instead of 'new'
* the URL is better I think: `/admin/applications/:id/change_lead_provider`
* the controller doesn't need an empty `def new; end` method
* if using 'new', and there's an error on the page, then user ends up on the show page anyway, which if they reload, causes an error. Using 'show' avoids this.

example page:
Screenshot 2025-02-14 at 13 54 54

### DIFF
--- a/app/controllers/npq_separation/admin/applications/change_lead_provider_controller.rb
+++ b/app/controllers/npq_separation/admin/applications/change_lead_provider_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module NpqSeparation
+  module Admin
+    module Applications
+      class ChangeLeadProviderController < NpqSeparation::AdminController
+        before_action :set_application, :set_change_lead_provider
+
+        def create
+          @change_lead_provider.assign_attributes(lead_provider_params)
+
+          if @change_lead_provider.change_lead_provider
+            redirect_to npq_separation_admin_application_path(@application)
+          else
+            render :show, status: :unprocessable_entity
+          end
+        end
+
+      private
+
+        def set_application
+          @application = Application.find(params[:id])
+        end
+
+        def set_change_lead_provider
+          @change_lead_provider =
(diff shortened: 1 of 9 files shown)


app/controllers/npq_separation/admin/applications/change_lead_provider_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+          end
+        end
+
+      private
+
+        def set_application
+          @application = Application.find(params[:id])
+        end
+
+        def set_lead_provider
```
jebw, line 25: `set_change_lead_provider` - names a bit misleading?

app/services/applications/change_lead_provider.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+  private
+
+    def lead_provider
+      @lead_provider ||= LeadProvider.find(lead_provider_id)
+    end
+
+    def different_lead_provider
+      errors.add(:lead_provider_id, :inclusion) if lead_provider_id == application.lead_provider.id
```
jebw, line 39: `unless lead_provider_id_changed?` ?

jebw (approved):
LGTM - couple of really minor suggestions

app/services/applications/change_lead_provider.rb
alkesh: I couldn't use any `_changed` or `_change` methods because `lead_provider_id` is just an integer and not an ActiveModel - and when the validation kicks in, neither `lead_provider` nor `application.lead_provider` have registered any changes yet.

PR #2238, opened 2025-02-19
[CPDNPQ-2645] Rake task to correct unverified users

jebw:
### Context

Ticket: [CPDNPQ-2645](https://dfedigital.atlassian.net/browse/CPDNPQ-2645)

We have a subset of users who have incorrectly have their TRN flagged as unverified. This happened because when the user logs in via DfE Identity, their `trn_verified` field and `trn_lookup_status` are always updated according to whether DfE Identity holds a TRN for the user.

Their TRN is _not_ updated if DfE Identity does not hold a TRN for the user.

This leads to a situation where users who have gone down the DQT path are left with their TRN set, but having had that marked as verified, any subsequent login will mark it back as unverified.

This provides a script to identify these users and correct their `trn_verified` status

1. It finds users who have `trn_auto_verified` set to `true` (so they went down the DQT check process) but `trn_verified` set to false. This is a pattern we have identified for users who've been impacted by this issue. The 'login' code never updates the `trn_auto_verified` field but the DQT check does so this is an approximate identifier for the users impacted.
2. It checks the users version history looking for evidence `trn_verified` was updated from `true` to `false`
4. It checks that the user did not have their TRN changed which would be a legitimate reason to mark their TRN unverified
5. If all checks are passed then the users `trn_validated` field is updated
6. It also attempts to use the earliest value for `trn_lookup_status` from the versions which changed `trn_verified` to false - we use this version because once its gone to nil, it will remain nil. Its not just set to `Found` because it was nil on some users previously before they were marked as unverified

[CPDNPQ-2645]: https://dfedigital.atlassian.net/browse/CPDNPQ-2645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tasks/one_off/backfill_trn_verified_status.rake
+++ b/lib/tasks/one_off/backfill_trn_verified_status.rake
@@ -3,7 +3,7 @@ namespace :one_off do
   task :backfill_trn_verified_status, %i[dry_run] => :environment do |_task, args|
     logger = Rails.env.test? ? Rails.logger : Logger.new($stdout)
     dry_run = args[:dry_run] != "false"
-    issue_introduced = Date.parse("2024-10-03").at_beginning_of_day
+    issue_introduced = Date.parse("2024-11-28").at_beginning_of_day
 
     User.transaction do
       # Find potential affected users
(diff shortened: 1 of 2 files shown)


jebw:
Yeah, safety measure

PR #2244, opened 2025-02-24
[CPDNPQ-2368] stop invalid UUID reaching query object

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2368

Using an invalid UUID in the filter param for declarations, raises an error.
The other APIs cope with an invalid UUID fine (no error raised).

I've added a `uuids` option to the `extract_conditions` method - which if true will remove invalid UUIDs from the conditions.

I don't think we need additional validation - especially because you can pass (for example) multiple participant_ids - so if only one is invalid, it is better to just return values for the IDs that _are_ valid, rather than raising an error for the one that is not valid.

Whilst investigating this, I noticed our 'try it out' feature on the swagger docs was not working, because the filter, page and sort parameters were not quite described correctly - I've fixed that in this PR.

### DIFF
--- a/app/services/declarations/query.rb
+++ b/app/services/declarations/query.rb
@@ -44,7 +44,7 @@ module Declarations
     def where_participant_ids_in(participant_ids)
       return if ignore?(filter: participant_ids)
 
-      scope.merge!(Declaration.where(user: { ecf_id: extract_conditions(participant_ids) }))
+      scope.merge!(Declaration.where(user: { ecf_id: extract_conditions(participant_ids, uuids: true) }))
     end
 
     def where_cohort_start_year_in(cohort_start_years)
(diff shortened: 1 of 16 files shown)


jebw (approved):
LGTM

jebw (commented):
Adding the UUID gem feels like an unneeded and unmaintained dependency

Ruby can do generation of UUIDs itself and for validation a regex is sufficient for UUIDv4

An example is here - https://stackoverflow.com/a/74948952 - Note the comment below, the example regex is 'correct' but needs modifying to check the entire string a a UUIDv4 rather than just containing a UUIDv4

jebw (approved):

PR #2245, opened 2025-02-24
CPDNPQ-2650 Fix nil trn issue

jebw:
### Context

Ticket: [CPDNPQ-2650](https://dfedigital.atlassian.net/browse/CPDNPQ-2650)

Currently we are not getting the TRN through from our DfE Identity integration. This appears to be because our custom OIDC provider implementation derives from the stock OAuth2 one with custom code to then fetch the additional OIDC token.

This in turn is receiving 2 tokens (I think) - the original `auth_token` and the `id_token` - the second contains the additional data we want but is getting thrown away automatically by the OAuth2 gem.

Answer is to use the standard OIDC provider and avoid maintaining our own code. This brings with it some complications regarding differences in behaviour, the two main ones are

1) Users signing up for Email updates - the only implementation did a dance where it set a flag in the users session, sent them off to the OIDC provider, then used this session flag to determine the apps behaviour on their return. We can just pass this param straight through the OIDC provider, and have it provided to us in `omniauth.params` on their return so thats what I'm doing

2) Force the user to relogin after creating an Application. I'm not sure the origin of this feature but this has forced me to create a new custom provider which very minimally appends the `prompt=login` param when handing the user over to the OIDC provider when appropriate.

1. Replace the previous OIDC strategy inheriting from the OAuth2 strategy with a more minimal one inheriting from the OIDC strategy
2. Moved logic for extracting additional user data out to the user model where its more easily testable and still gives us access to other custom data
3. Changed the Signup for Email Updates implementation to use normal OIDC param passing
4. Added a minimal override in our custom provider to force the user to relogin after completing an application
5. Removed the stubbing of the ENV global variable in specs - this removed access to other ENV variables and had potential to cause unintended spec failures. We can avoid doing it so do so.
6. Now we're relying on HOSTING_DOMAIN being set locally we need to prevent the initializer from rewriting the api docs constantly on developers machines
7. Replaced that CSRF check that got removed as part of another bugfix

[CPDNPQ-2650]: https://dfedigital.atlassian.net/browse/CPDNPQ-2650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env.test
```diff
@@ -3,3 +3,7 @@ QUALIFIED_TEACHERS_API_URL="https://qualified-teachers-api.example.com"
 QUALIFIED_TEACHERS_API_KEY="some-apikey-guid"
 DQT_API_URL="https://dqt-api.example.com"
 DQT_API_KEY="test-apikey"
+TRA_OIDC_DOMAIN=https://tra-domain.com/
+TRA_OIDC_CLIENT_ID=npq
+TRA_OIDC_CLIENT_SECRET=randomsecret
+HOSTING_DOMAIN=http://test.host
```
alkesh, line 9: missing linefeed at end of this file

app/models/user.rb
```diff
@@ -195,13 +185,23 @@ def set_closed_registration_feature_flag
...
+    self.updated_from_tra_at = Time.zone.now
+    self.full_name = extra_info&.preferred_name.presence || provider_data.info.name
+
+    if extra_info&.birthdate.present?
+      self.date_of_birth = Date.parse(extra_info.birthdate, "%Y-%m-%d")
+    end
+
+    # The user's TRN should remain unchanged if the TRA returns an empty TRN
+    if extra_info&.trn.present?
+      self.trn = extra_info.trn.strip
```
alkesh, line 201: is this `strip` needed?
if so, it needs a spec

lib/omniauth/strategies/tra_openid_connect.rb
```diff
@@ -1,67 +1,21 @@
...
-      def build_access_token
-        verifier = request.params["code"]
-        redirect_uri = full_host + callback_path
-        client.auth_code.get_token(
-          verifier,
-          { redirect_uri: }.merge(token_params.to_hash(symbolize_keys: true)),
-          deep_symbolize(options.auth_token_params),
-        )
+      def force_login_prompt?
+        request.session["clear_tra_login"] == true
```
alkesh, line 18: this `clear_tra_login` isn't covered by a spec

alkesh (approved):
does `docs/get_an_identity.md` need updating too?

lib/omniauth/strategies/tra_openid_connect.rb
jebw: There isn't really any specs around any of this so I think we actually need to be adding proper specs for this but I think thats a bigger piece of work so I'll punt that to a separate ticket and continue relying upon manual testing for now

slawosz (approved):

PR #2256, opened 2025-02-26
[CPDNPQ-2647] do not use archived users when logging in

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2647

When logging in, users that have been archived are being found and used (because they have a UID that matches the one of the user logging in)

* do not use an archived user
* before saving the new user using the DfE Identity data, blank the UID of the clashing archived user - there is a uniqueness constraint on UID that will fail otherwise.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -71,14 +71,15 @@ class User < ApplicationRecord
       return user
     end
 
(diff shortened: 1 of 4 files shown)


jebw (approved):
LGTM :rocket:

PR #2257, opened 2025-02-26
Remove defunct framework defaults initializers

rwrrll:
### Context

(no ticket)

All these files are fully commented out and we already have `config.load_defaults 7.0`, so we can remove these initializers.


jebw (approved):
Nice :+1:

PR #2261, opened 2025-02-27
[CPDNPQ-2638] cope with expired/non-existent session on qualified_teacher_check step

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2638

If the user's session expires, or does not exist, and the page they have in their browser is the qualfied_teacher_check page - then submitting this form causes an error.

The `after_save` on this step, and also the `next_step` on the next step (course_start_date) rely on having a current_user.
In the expired/non-existent session scenario, that user is a `NullUser`.
The fix is:
* cope with NullUser in the qualfied_teacher_check step
* check if the user is is not a NullUser before the `dqt_mismatch` and `course_start_date` steps (either of which could follow the qualfied_teacher_check step) , and redirect them to the start of the journey if it is a NullUser

### DIFF
--- a/app/forms/questionnaires/course_start_date.rb
+++ b/app/forms/questionnaires/course_start_date.rb
@@ -42,7 +42,7 @@ module Questionnaires
     end
 
     def requirements_met?
-      query_store.current_user.present?
+      query_store.current_user&.actual_user?
     end
 
     def next_step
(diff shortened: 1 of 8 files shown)


jebw (approved):
LGTM

PR #2262, opened 2025-02-27
[CPDNPQ-2411] Redact mailer logs

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2411

When Rails mailers process and deliver outbound mail, recipient addresses are leaking into Rails log output, which is a PII concern. This PR:

- applies Rails filter_parameters to mailer log entries
- disables ActiveJob argument logging for mailer jobs (arguments will still be logged by the actual mailer, but now redacted)
- adds basic test coverage to mailers to make sure they still work
- adds `full_name` to the list of filtered parameters

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -15,5 +15,10 @@ Rails.application.config.filter_parameters += %i[
   national_insurance_number
   nino
   to
+  cc
+  bcc
   full_name
+  key
+  code
+  trn
 ]


config/initializers/filter_parameter_logging.rb
```diff
@@ -14,4 +14,6 @@
   ssn
   national_insurance_number
   nino
+  to
+  full_name
```
jebw, line 18: can we add `cc`, `bcc`, `key`, `code`, `trn`, `ecf_id` in here

config/initializers/mailer_log_redactor.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+private
+
+  def redact(payload)
+    @filter ||= ActiveSupport::ParameterFilter.new(filter_parameters)
+    @filter.filter(payload)
+  end
+
+  def filter_parameters
```
jebw, line 14: super minor but you can use delegate for this?

```ruby
delegate :filter_parameters, to: Rails.application.config
```

jebw (commented):
Looks really good but can you expand the params to be redacted, thanks

jebw (approved):
LGTM

PR #2268, opened 2025-02-28
fix konduit

alkesh:
### Context

Using konduit (e.g. `make konduit-snapshot`) puts the database URL with the wrong indentation in our `config/database.yml`.

This is because the primary database was removed in https://github.com/DFE-Digital/npq-registration/pull/2055, the indentation changed, but the Makefile was not updated.

Fix the indentation, so the URL is put in the right place.


jebw (approved):
:raised_hands:

PR #2273, opened 2025-03-05
[CPDNPQ-2518] Rename declarations CSV

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2518

When finance users download declaration CSVs, they would like a specific filename rather than "assurance_report.csv"

Previous work had copied over the `filename` method but it wasn't being used, so I've hooked this up and tweaked the method to make it work.

### DIFF
--- a/spec/requests/npq_separation/admin/finance/statements/assurance_reports_controller_spec.rb
+++ b/spec/requests/npq_separation/admin/finance/statements/assurance_reports_controller_spec.rb
@@ -9,6 +9,8 @@ RSpec.describe NpqSeparation::Admin::Finance::Statements::AssuranceReportsContro
     subject { response }
 
     before do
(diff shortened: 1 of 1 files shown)


app/controllers/npq_separation/admin/finance/statements/assurance_reports_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
 
     respond_to do |format|
       format.csv do
-        render body: @serializer.serialize
+        send_data @serializer.serialize, filename: @serializer.filename
```
alkesh, line 11: needs a spec in `spec/requests/npq_separation/admin/finance/statements/assurance_reports_controller_spec.rb`.
should be able to check the `subject.headers["content-disposition"]`

alkesh (approved):

PR #2280, opened 2025-03-06
Distinguish between school and fe setting where only one is in the PP50 list

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2672


app/services/funding_eligibility.rb
```diff
@@ -231,4 +231,8 @@ def not_eligible_england_ehco?
   def childminder?
     query_store.childminder?
   end
+
+  def work_setting
+    query_store.work_setting
+  end
```
alkesh, line 237: could replace the above two with
```
delegate :childminder?, :work_setting, to: :query_store
```

alkesh (approved):

PR #2281, opened 2025-03-06
[CPDNPQ-2463] ensure errors are JSON for API requests

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2463

The Not Found page always returned HTML

We don't actually return formats based on the Content-Type header,
we always return JSON from API endpoints.

So now, the errors controller will check the request URL, and if it is API, it will return JSON.
Only the Not Found page was actually incorrectly returning HTML - the other responses were already correct (401, 422, 500).

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,6 +1,8 @@
 class ErrorsController < ApplicationController
   layout "application"
 
+  before_action :set_format
+
   def not_found
     respond_to do |format|
       format.html { render status: :not_found }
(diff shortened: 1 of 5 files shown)


alkesh:
> Change looks good!
> 
> Worth mentioning we already respond in the correct format according to the `Accept` header:
> 
> ```
> curl https://register-national-professional-qualifications.education.gov.uk/api/v3/lolwut -H "Accept: application/json"
> {"error":"Resource not found"}
> ```
> 
> So this change is to _force_ JSON regardless of the client/request. Assuming we really want that, LGTM 👍

I think that's the best.
Currently, if you use our API with another Accept header (e.g. text/xml, or text/plain) we don't actually return any error, we just return JSON.

jebw (approved):
LGTM

PR #2285, opened 2025-03-11
[CPDNPQ-2688] add extra info to sentry error

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2688

There are a couple of errors appearing in sentry - for one of them, I can't be sure how we're getting to that point.
errors we're seeing:
- Email address must be unique - last seen: Mar 9, 2025 1:56:55 PM
- uid has already been taken - last seen: Feb 14, 2025 4:43:59 PM - this was resolved by CPDNPQ-2647

I've tried a few scenarios, but cannot get to a point where we hit the errors we're seeing.

Adding extra info to the sentry error

### DIFF
--- a/app/controllers/omniauth_controller.rb
+++ b/app/controllers/omniauth_controller.rb
@@ -14,11 +14,14 @@ class OmniauthController < Devise::OmniauthCallbacksController
 
     # @user.persisted? checks that it exists and has been persisted to the database
     # @user.save checks that any changes made to an existing record have been persisted to that persisted record
-    if @user.persisted? && @user.save
+    if @user.persisted? && (user_saved = @user.save)
       session["user_id"] = @user.id
       @user.set_closed_registration_feature_flag
       sign_in_and_redirect @user
     else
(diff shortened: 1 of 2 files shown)


app/controllers/omniauth_controller.rb
```diff
@@ -14,19 +14,25 @@ def tra_openid_connect
...
     # @user.save checks that any changes made to an existing record have been persisted to that persisted record
-    if @user.persisted? && @user.save
+    if @user.persisted? && (user_saved = @user.save)
       session["user_id"] = @user.id
       @user.set_closed_registration_feature_flag
       sign_in_and_redirect @user
     else
+      # TODO: need a feature test for the error scenario:
+      # - uid has already been taken
+      user_id_with_clashing_email = User.find_by(email: provider_data.info.email).id
```
jebw, line 24: defensive _but_ this is already doing something we don't understand so better to use safe navigation operator `&.id`

jebw (approved):
LGTM - one minor comment

PR #2291, opened 2025-03-12
switch from selenium webdriver to cuprite

alkesh:
### Context

There are lots of specs failing because [chrome was updated](https://github.com/actions/runner-images/pull/11761) in the github actions runner image. There currently isn't a selenium webdriver version that supports this version.

switching from selenium-webdriver to cuprite fixes the issues

Things that don't work:
* our Axe accessibility tests (they won't fail if something is not accessible) - however, I don't think they're testing what we think they should be testing - the [wcag22aa](https://github.com/dequelabs/axe-core/blob/master/doc/rule-descriptions.md#wcag-22-level-a--aa-rules) ruleset has only one rule it checks (for `target-size`) - we should probably be using the [wcag2a](https://github.com/dequelabs/axe-core/blob/master/doc/rule-descriptions.md#wcag-20-level-a--aa-rules) ruleset as well.

Other things:
* creating an admin - would strip the space from the full name because we were using an email field - cuprite picked this up as a test failure


app/views/admin/admins/new.html.erb
```diff
@@ -19,10 +19,10 @@
                               label: { text: "Email address" },
                               hint: { text: "This can be for a new or existing user" },
                               width: "three-quarters" %>
-      <%= f.govuk_email_field :full_name,
-                              label: { text: "Full name" },
-                              hint: { text: "This will set the name for the admin created using the above entered email" },
-                              width: "three-quarters" %>
+      <%= f.govuk_text_field :full_name,
```
alkesh, line 22: oh - good spot @rwrrll !

jebw (approved):
LGTM

PR #2299, opened 2025-03-13
Add documentation for eligiblity list

slawosz:
### Context
[Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2507](https://dfedigital.atlassian.net/browse/CPDNPQ-2507)


jebw (changes requested):
Can this be expanded to include where the logic which uses these lists is, and some high level comments on what that logic is? thanks

jebw (approved):
Looks great, thanks :+1:

PR #2308, opened 2025-03-18
[CPDNPQ-2453] set delivery partners on declarations

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2453

Allow `delivery_partner_id` and `secondary_delivery_partner_id` to be specified when creating a declaration.

Call the validations for delivery partners from the `Declaration` model directly.
Update the validation error messages on the Declaration model, so they are public-facing API error messages.
Allow two new attributes: `delivery_partner_id`, `secondary_delivery_partner_id`.

### DIFF
--- a/app/services/declarations/create.rb
+++ b/app/services/declarations/create.rb
@@ -30,7 +30,7 @@ module Declarations
     validate :output_fee_statement_available
     validate :validate_has_passed_field, if: :validate_has_passed?
     validate :validates_billable_slot_available
-    validate :declaration_date_not_in_the_future
+    validate :declaration_valid
 
     attr_reader :raw_declaration_date, :declaration
 
@@ -167,10 +167,6 @@ module Declarations
       errors.add(:base, :declaration_already_exists)
     end
 
-    def declaration_date_not_in_the_future
-      errors.add(:declaration_date, :future_declaration_date) if declaration_date&.future?
-    end
-
     def validate_has_passed_field
       self.has_passed = has_passed.to_s
 
@@ -205,6 +201,13 @@ module Declarations
       else
         raise ArgumentError, I18n.t(:cannot_create_completed_declaration)
       end
+    end
+
+    def declaration_valid
+      return if errors.any?
+
+      declaration = Declaration.new(declaration_parameters_for_create)
+      errors.merge!(declaration.errors) unless declaration.valid?
     end
   end
 end
--- a/app/services/declarations/create.rb
+++ b/app/services/declarations/create.rb
@@ -11,6 +11,8 @@ module Declarations
     attribute :declaration_date, :datetime
     attribute :course_identifier, :string
     attribute :has_passed
+    attribute :delivery_partner_id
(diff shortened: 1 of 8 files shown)


app/services/declarations/create.rb
```diff
@@ -30,6 +32,15 @@ class Create
...
 
+    validates :delivery_partner_id, presence: true, if: -> { Feature.declarations_require_delivery_partner? }
+    validates :delivery_partner_id, inclusion: { in: :available_delivery_partners }, if: -> { delivery_partner_id }
+
+    validates :secondary_delivery_partner_id, absence: true, unless: :delivery_partner_id
+    validates :secondary_delivery_partner_id, inclusion: { in: :available_delivery_partners },
+                                              if: -> { secondary_delivery_partner_id }
+
+    validate :delivery_partners_are_not_the_same, if: :delivery_partner_id
+
```
jebw, line 43: Rather than duplicate this validation - it might be nice if we can add a way of copying errors for matching attributes over from the model(s) the service is wrapping?

My validations are on the relationships rather than ids but I think they could equally be moved to validate on the `_id` attributes themselves?

app/services/declarations/create.rb
```diff
@@ -30,6 +32,15 @@ class Create
     validate :validates_billable_slot_available
     validate :declaration_date_not_in_the_future
 
+    validates :delivery_partner_id, presence: true, if: -> { Feature.declarations_require_delivery_partner? }
```
jebw, line 35: This rule is a bit more complicated now in the Declaration model and is going to get more complicated (we need to handle overseas applications differently it turns out)

spec/services/declarations/create_spec.rb
```diff
@@ -361,20 +404,5 @@
...
-
-      it "returns false" do
-        expect(service.create_declaration).to be false
-      end
-
-      it "has an error" do
-        service.create_declaration
-        expect(service).to have_error(:declaration_type, :inclusion, "The entered '#/declaration_type' is not recognised.")
-      end
-    end
```
jebw, line 378: Whats the reason for dropping this spec?

jebw (approved):
LGTM - some 'nice to have' comments but nothing strictly necessary - the delivery partner id

I wasn't following why a spec was getting removed?

spec/services/declarations/create_spec.rb
alkesh: that spec just moved - into the _validations_ section - line 294 above.

app/services/declarations/create.rb
```diff
@@ -28,7 +30,7 @@ class Create
     validate :output_fee_statement_available
     validate :validate_has_passed_field, if: :validate_has_passed?
     validate :validates_billable_slot_available
-    validate :declaration_date_not_in_the_future
```
alkesh, line 31: this validation is in the declaration validations, which are now checked in this service

jebw (approved):
LGTM

PR #2313, opened 2025-03-19
Fix feature flags bug

EmmaFrith:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2510

Change how feature flagged users are retrieved to improve performance and reduce the likelihood of nil values, and write associated tests.


jebw (approved):
LGTM :rocket:

PR #2322, opened 2025-03-20
[CPDNPQ-2727] pick schedule correctly for 2025 and 2024 cohorts

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2727

Providers are unable to accept applications when they don't specify `schedule_identifier`.
Most accepts do not specify `schedule_identifier`.

Right now - there are 2 schedules that could be used for the current cohort - the autumn 2024 schedule (which was updated to end on 6th June in CPDNPQ-2692), or the spring 2025 schedule.
We're seeing errors, because we currently try and pick the autumn 2024 schedule, even for courses that are only in the spring 2025 schedule.

For the 2025 and 2024 cohorts, courses are only ever in one of these schedules - never in both.
For previous cohorts, courses are in both schedules.

For the 2025 cohort, use the spring 2025 schedule, for the 2024 cohort, use the autumn 2024 schedule.
For applications being accepted in the 2021-2023 cohorts - the previous date ranges will still be respected when picking the autumn or spring schedules.

Also `course_group.schedule_for` is only ever called with `schedule_date: Date.current` - so the method `autumn_schedule_2022?` would never get called - so I removed it.

### DIFF
--- a/app/services/course_groups/leadership.rb
+++ b/app/services/course_groups/leadership.rb
@@ -10,10 +10,12 @@ module CourseGroups
     delegate :schedules, to: :course_group
 
     def schedule
-      if autumn_schedule_2022?(schedule_date)
-        schedules.find_by!(cohort:, identifier: "npq-leadership-autumn")
-      elsif autumn_schedule_2024?(schedule_date)
-        schedules.find_by!(cohort:, identifier: "npq-leadership-autumn")
+      if autumn_schedule_2024?(schedule_date)
+        if cohort.start_year == 2025
+          schedules.find_by!(cohort:, identifier: "npq-leadership-spring")
+        elsif cohort.start_year == 2024
+          schedules.find_by!(cohort:, identifier: "npq-leadership-autumn")
+        end
       elsif spring_schedule?(schedule_date)
         schedules.find_by!(cohort:, identifier: "npq-leadership-spring")
       elsif autumn_schedule?(schedule_date)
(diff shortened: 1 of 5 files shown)


app/services/course_groups/leadership.rb
```diff
@@ -10,10 +10,8 @@ class Leadership
     delegate :schedules, to: :course_group
 
     def schedule
-      if autumn_schedule_2022?(schedule_date)
-        schedules.find_by!(cohort:, identifier: "npq-leadership-autumn")
-      elsif autumn_schedule_2024?(schedule_date)
-        schedules.find_by!(cohort:, identifier: "npq-leadership-autumn")
+      if autumn_schedule_2024?(schedule_date)
+        schedules.find_by(cohort:, identifier: "npq-leadership-autumn") || schedules.find_by!(cohort:, identifier: "npq-leadership-spring")
```
alkesh, line 14:
```suggestion
        if cohort.start_year == 2025
          schedules.find_by!(cohort:, identifier: "npq-leadership-spring")
        elsif cohort.start_year == 2024
          schedules.find_by(cohort:, identifier: "npq-leadership-autumn")
        end
```
an alternative implementation is this.

peteryates, line 14: Would this be clearer as a single query?

```suggestion
        schedules.find_by!(cohort:, identifier: %w[npq-leadership-autumn npq-leadership-spring])
```

peteryates: This is clearer still actually and makes it obvious that 2024 and 2025 are different.

peteryates (approved):
This looks sensible to me and is easy enough to follow given the context.

app/services/course_groups/leadership.rb
peteryates: Scratch that actually, your alternative above is more informative.

jebw (approved):
LGTM

PR #2332, opened 2025-03-26
[CPDNPQ-2688] fix user unable to log in

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2688

user unable to log in - we see the error `Email address must be unique` when the user is being saved on login.

Refactor the code responsible for finding/creating a user on login.

Cope with the following scenario:  there is a clashing user in the system, that has the same email as the user logging in, but a different DfE Identity UID.

The error we're seeing in production may not be that last scenario - the log entry doesn't show any clashing user, but the error is still `Email address must be unique`.
Hopefully, this refactor will either fix the problem, or make it easier to see what the issue is.


slawosz (approved):
I like the approach with using end to end tests in service 👍  Good feature specs coverage too.

app/services/users/find_or_create_from_provider_data.rb
```diff
@@ -0,0 +1,68 @@
...
+        Rails.logger.info("[GAI] User not found using UID, UID=#{provider_data.uid}, using email to find user")
+        check_if_supplied_uid_matches_archived_account # CPDNPQ-2647
+        user = User.find_or_initialize_by(email: provider_data.info.email, archived_at: nil)
+        user.assign_attributes(provider: provider_data.provider, uid: provider_data.uid)
+      end
+
+      assign_provider_data(user, provider_data)
+      user.assign_attributes(feature_flag_id: feature_flag_id)
+
+      user.tap(&:save)
```
jebw, line 28: This is the only part I'm not sure about, I know we inherited it like it but feels likely to result in a silent fail

I wonder if this would be better as `user.tap(&:save!)`

jebw (commented):
Looks good but wonder about using this as an opportunity to swap to `.save!` to avoid the silent fail?

app/services/users/find_or_create_from_provider_data.rb
alkesh: This service is only called in one place, in `OmniauthController`.
Right after calling this service, the controller does it's own save, and sends the error to sentry - this is useful to see when unexpected errors are occurring.
The user is then just redirected to the `failed_sign_in_path`.

PR #2333, opened 2025-03-26
[CDPNPQ-2487] Set caching headers

jebw:
### Context

Ticket: [CPDNPQ-2487](https://dfedigital.atlassian.net/browse/CPDNPQ-2487)

We should not be using `no-cache`, the ITHC report suggests one of `public`, `private` or `no-store`. I've gone with `no-store`.

1. Ensure pages are flagged uncacheable
2. Added some missing specs to check cache headers in a few places

[CPDNPQ-2487]: https://dfedigital.atlassian.net/browse/CPDNPQ-2487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/requests/npq_separation/admin_controller_spec.rb
```diff
@@ -3,9 +3,11 @@
 RSpec.describe NpqSeparation::AdminController, type: :request do
   include Helpers::NPQSeparationAdminLogin
 
+  subject(:admin_response) { get(npq_separation_admin_path) && response }
```
slawosz, line 6: I personally would name it `admin_request`, just feels better for me.

slawosz (approved):

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
+  before_action :set_cache_headers
```
alkesh, line 4: is it worth adding an entry to `spec/controllers/application_controller_spec.rb` for completeness and symmetry?
I can see this line is already covered by the other request and controller specs, so it's not a coverage issue.

alkesh (approved):

spec/requests/npq_separation/admin_controller_spec.rb
jebw: Yeah, thats what I've usually been naming this. 

The reason I decided that was the wrong name was it doesn't actually return a page request object, it returns a page response object - it makes a request, but this is equivalent to a `let` and you'd name a `let` for what it holds, not how it got it, eg `let(:user)` instead of `let(:create_factory_user)`

So its either `make_request` (ie named for what its doing) or `admin_response` (ie named for what it is) ?

app/controllers/application_controller.rb
jebw: Tbh I'd not realised there was an `spec/controllers/application_controller_spec` - yep, will add it

PR #2341, opened 2025-03-28
All young offender and hospital schools applications should be in review

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2715


jebw (approved):
LGTM

PR #2354, opened 2025-04-01
[BAU] Fix visualisation github action not to use ancient ubuntu release

jebw:
### Context

Ticket: BAU

Visualisation step has failed because of GH Actions ubuntu 20.04 brownout. We shouldn't be using version 20.04

1. Use Ubuntu 24.04


alkesh (approved):

alkesh:
`restore_snapshot_database.yml` also uses 20.04

PR #2356, opened 2025-04-02
[CPDNPQ-2698] Add admin UI to manage delivery partnerships

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2698

Ability to set the lead providers and cohorts for a delivery partner

n.b. I had to remove the ID from the delivery partners index because otherwise the table wraps and gets super messy — reflecting on this, not having IDs here is probably not a huge loss since the table is paginated anyway.

Also in the design mockups the term "course provider" is used instead of "lead provider", so I've stuck with the latter for now since "course provider" isn't yet used anywhere else. Presumably that'll be a later update.

I also tweaked the action labels: "Assign lead providers" over "Add course provider" for the action link because you can also remove them, and "Update name" over "Update" to clarify what that action does.

### DIFF
--- a/app/views/npq_separation/admin/delivery_partnerships/edit.html.erb
+++ b/app/views/npq_separation/admin/delivery_partnerships/edit.html.erb
@@ -29,7 +29,7 @@
   </div>
 <% end %>
 
-<script type="text/javascript">
+<%= nonced_javascript_tag do %>
   // uncheck all cohorts when a lead provider is deselected
   document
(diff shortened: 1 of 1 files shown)


jebw (changes requested):
Looks great - only minor comment is the inline JS tag

jebw (approved):
LGTM :+1:

PR #2358, opened 2025-04-02
[CPDNPQ-2754] Handle nil admin in OTP helper

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2754

Before this change, if you attempt to sign in to a non-production admin console with an email address that doesn't match an admin account, you'll get the 500 error screen and an exception will be recorded/alerted via Sentry.

After this change, this scenario will behave the same as production:

Screenshot 2025-04-02 at 13 27 27


jebw (approved):
LGTM

PR #2364, opened 2025-04-03
2747 - remove users sections from feature flags index and show pages

EmmaFrith:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2747

In the new feature flags section, we are currently trying to show how many actors (i.e. users) are feature-flagged for different features:

image
image

However, this does not work in an intuitive way.

Every time a feature-flagged user signs in to the NPQ service, they’re recorded as another actor - making it seem like there are more actors than there are. 

What’s more, the actors section is only relevant for the Registration Open feature flag - so it’s confusing to have the actors section for all flags. 

We want to prevent confusion by removing the actors information from these pages.

Instead, we can link users to the ‘Closed registration users’ page, which is a better source of truth to understand who is feature flagged.

image

**User story**
As a DFE admin 
I need to know who can register late for an NPQ

- Remove the Users column from the ‘New feature flags’ index page.
- Remove the Users section from the feature flags show pages.
- Link users to the ‘Closed registration users’ page from relevant places.
- Remove now-redundant code from controller.

### DIFF
--- a/app/controllers/admin/features_controller.rb
+++ b/app/controllers/admin/features_controller.rb
@@ -6,9 +6,7 @@ class Admin::FeaturesController < SuperAdminController
 
   def show
     @feature = params[:id]
-    if Feature::FEATURE_FLAG_KEYS.include?(@feature)
-      @users = User.with_feature_flag_enabled(@feature)
-    else
+    unless Feature::FEATURE_FLAG_KEYS.include?(@feature)
       redirect_back fallback_location: admin_features_path
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -63,12 +63,6 @@ class User < ApplicationRecord
     Users::FindOrCreateFromProviderData.new(provider_data: provider_data, feature_flag_id: feature_flag_id).call
   end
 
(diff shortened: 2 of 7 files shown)


jebw (changes requested):
This looks great :raised_hands:  

This change does remove the need for the `User#with_feature_flag_enabled?` method. The other day we'd decided this method is a little confusing since it may not be correct if a user has not logged in since registration was closed?

With that in mind I think that method can also be removed together with its the specs.

That in turn means theres no need for `spec/suppport/shared_contexts/with_activerecord_flipper.rb` so that can also be removed.

EmmaFrith:
> This looks great 🙌
> 
> This change does remove the need for the `User#with_feature_flag_enabled?` method. The other day we'd decided this method is a little confusing since it may not be correct if a user has not logged in since registration was closed?
> 
> With that in mind I think that method can also be removed together with its the specs.
> 
> That in turn means theres no need for `spec/suppport/shared_contexts/with_activerecord_flipper.rb` so that can also be removed.

Thanks @jebw! That's quite a lot of code I forgot about 🙈

jebw (approved):
LGTM :rocket:

PR #2369, opened 2025-04-04
[CPDNPQ-2434] Remove redundant feature flags

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2434

> There are some redundant feature flags.
> This includes the senco course. This defaults to off in both review apps and staging. It no longer needs to be feature flagged and has led to confusion when staging got rebuilt so its best to just remove the feature flag making the senco course always available.

Since the purpose of `Feature::FEATURE_FLAG_KEYS` is to define the current set of feature flag keys, we may as well make this authoritative and automatically delete any other Flipper features that aren't referenced in that array. Also removed conditional paths throughout the app referencing defunct feature flags.

**⚠️ Important caveat:** `Feature.initialize_feature_flags` is run in the seeds process, but due to [.github/actions/deploy-environment-to-aks/action.yml#L68](https://github.com/DFE-Digital/npq-registration/blob/main/.github/actions/deploy-environment-to-aks/action.yml#L68), it seems that this will not occur automatically on deployment to production, **therefore after this is merged, `rake feature_flags:initialize` (or `Feature.initialize_feature_flags` in a console) probably needs to be run.** It might make sense to consider automating this as a next step, we probably always want production flags in sync with the deployed code.

### DIFF
--- a/app/views/accounts/_funding_details.html.erb
+++ b/app/views/accounts/_funding_details.html.erb
@@ -27,7 +27,7 @@
 
         <% scholarship_funding_row(application) %>
 
-        <% if Flipper.enabled?(:targeted_support_funding) || application.created_at < APRIL_2024_CUTOFF_DATE %>
+        <% if application.created_at < APRIL_2024_CUTOFF_DATE %>
           <div class="govuk-summary-list__row">
             <dt class="govuk-summary-list__key">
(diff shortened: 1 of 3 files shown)


jebw (approved):
LGTM

PR #2376, opened 2025-04-07
[CPDNPQ-2156] Admin - Add Adjustments

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2156

The ability to add adjustments.

There are separate tickets for:
* the changes to the statement page
* editing/deleting adjustments


app/services/statements/query.rb
```diff
@@ -6,7 +6,7 @@ class Query
     attr_reader :scope
 
     def initialize(lead_provider: :ignore, cohort_start_years: :ignore, updated_since: :ignore, state: :ignore, output_fee: true)
-      @scope = Statement.distinct.includes(:cohort)
+      @scope = Statement.distinct.includes(:lead_provider, :cohort)
```
alkesh, line 9: I was getting an eager-loading warning on the statements page without this

spec/forms/admin/adjustments/add_another_adjustment_form_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Admin::Adjustments::AddAnotherAdjustmentForm, type: :model do
+  include Rails.application.routes.url_helpers
```
jebw, line 6: Less objection here but still not a fan, how about

```ruby
let(:urls) { Rails.application.routes.url_helpers }

it { is_expected.to eq urls.some_statement_path }
```

app/forms/admin/adjustments/add_another_adjustment_form.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    validates_inclusion_of :add_another, in: %w[yes no]
+
+    def redirect_to
+      if add_another == "yes"
+        new_npq_separation_admin_finance_statement_adjustment_path(statement)
+      else
+        npq_separation_admin_finance_statement_path(statement)
+      end
+    end
```
jebw, line 20: Normally its the controller that handles the redirect location, and deciding by asking the model object (form in this case) for relevant information, ie `form.adding_another?`

app/forms/admin/adjustments/create_adjustment_form.rb
```diff
@@ -0,0 +1,54 @@
...
+    attribute :description
+    attribute :amount, :integer
+    attribute :adjustment
+
+    validate :adjustment_valid
+    validate :statement_open
+
+    def initialize(*)
+      super
+      session[:created_adjustment_ids] ||= []
```
jebw, line 19: This shouldn't need know about the session - its a model not a controller and this breaks separation of concerns - surely it can just get initialized with the adjustment_ids themselves and the controller be responsible for passing in the ids (which it can choose to hold in the session)

and have `#save` update the attribute with list of adjustment ids? Controller again can update whereever it holds it, eg session

app/models/adjustment.rb
```diff
@@ -0,0 +1,8 @@
+class Adjustment < ApplicationRecord
+  belongs_to :statement
+
+  validates :description, presence: true
+  validates :amount, presence: true
+  validates :amount, numericality: { other_than: 0 }, if: :amount
+  validates :amount, format: { with: /\A-?\d*\z/ }, if: -> { amount.present? }
```
jebw, line 7: typecasting will convert this to the appropriate number (or a 0 if its a word) so could this also just be `if: :amount` ?

if the user types something invalid wont it just convert it to 0 and it'll hit the non-zero validation on line 6?

spec/forms/admin/adjustments/create_adjustment_form_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    context "when the adjustment is valid" do
+      let(:description) { "Adjustment description" }
+      let(:amount) { 100 }
+
+      it "saves the adjustment" do
+        expect { form.save }.to change(statement.adjustments, :count).by(1)
+      end
+
+      it "adds the adjustment ID to the session" do
+        form.save # rubocop:disable Rails/SaveBang
```
jebw, line 23: Why not just use `save!` - if its invalid it'll just error out the line? and the user can still see the spec thats failing?

app/forms/admin/adjustments/create_adjustment_form.rb
jebw: Thinking some more on this overnight, ultimately the model is just being handed a hash (from its perspective) to persist into. It just happens to be a session, so a more minimal change would be to rename this to something like `store` instead of `session` 

I still think its better to have the controller putting stuff into the session, instead of the model but other models do back onto other stores (eg AR -> postgres)

spec/features/npq_separation/admin/finance/adjustments_spec.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Adjustments", type: :feature do
```
jebw, line 5: The spec isn't checking that the page is inaccessible when not logged in - thats crucial to catch any problems from future refactors around auth

jebw (changes requested):

app/models/adjustment.rb
alkesh: I think the attribute is only typecast when it is saved.
```
> a = Adjustment.new(description:'test', amount:'test')
> a.valid?
=> false
> a.errors.messages
=> => {:statement=>["must exist"], :amount=>["You can only enter numeric values in the adjustment amount field"]}
```

spec/forms/admin/adjustments/add_another_adjustment_form_spec.rb
alkesh: actually, moving the URLs out into the controller means I don't need this include any more

spec/forms/admin/adjustments/create_adjustment_form_spec.rb
alkesh: `save` was a method I'd implemented on the form.
I've renamed it to `save_form` to make it clear it's not a standard activerecord class.

app/forms/admin/adjustments/create_adjustment_form.rb
alkesh: I've gone for the first option - just maintaining an array of created_adjustment_ids in the form.
The controller is now responsible for storing it in the session, as well as clearing it.

app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
```diff
@@ -20,14 +23,19 @@ def create
...
-      redirect_to @add_another_form.redirect_to
+      if @add_another_form.adding_another_adjustment?
+        redirect_to new_npq_separation_admin_finance_statement_adjustment_path(@statement)
+      else
+        session[:created_adjustment_ids] = nil
+        redirect_to npq_separation_admin_finance_statement_path(@statement)
+      end
     else
       render :index, status: :unprocessable_entity
     end
```
jebw, line 41: Super minor but this could be 'inverted' to remove a layer of nesting and simplify readability

```ruby
if form.invalid?
  render :index, status: :unprocessable_entity
elsif form.adding_another?
  # redirect to add another screen
else
  # redirect to statement page
end
```

jebw (approved):
LGTM :raised_hands:

jebw (approved):
LGTM

PR #2391, opened 2025-04-11
[CPDNPQ-2788] add validation to check the delivery partners exist

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2788

currently, if an ID is specified that refers to a non-existent delivery partner, no error is raised, the delivery partner is just set to nil.

Add validation to check there is a delivery partner for the specified ID


jebw (approved):
LGTM

PR #2399, opened 2025-04-15
[CPDNPQ-2460] Delay removal of postgres on review app teardown

jebw:
### Context

Ticket: [CPDNPQ-2460](https://dfedigital.atlassian.net/browse/CPDNPQ-2460)

Currently Delayed::Job will sleep for a 5 second delay in between polling for new jobs. If it receives a signal to exit, it will continue to sleep until the sleep duration has passed only then does it exit.

This leads to a scenario in review apps where the worker application continues sleeping whilst the postgres container gets destroyed. I'd initially tried marking the worker app as explicitly depending upon postgres _but_ k8s reports back to terraform as having removed an application before the container has actually exited so that doesn't solve the problem.

1. The worker container explicitly depends upon postgres, via a 'wait' resource
2. The wait resource, a 15 second delay, is introduced between worker removal and postgres removal
3. The wait resource is set to 'count' zero for environments using azure resources instead of k8s containers

This only impacts the **review** environment 

[CPDNPQ-2460]: https://dfedigital.atlassian.net/browse/CPDNPQ-2460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,7 @@
 # https://gds-way.cloudapps.digital/manuals/programming-languages/editorconfig
 root = true
 
-[*.{css,erb,js,json,rb,scss,sh,yml}]
+[*.{css,erb,js,json,rb,scss,sh,tf,yml}]
 indent_size = 2
 indent_style = space
 charset = utf-8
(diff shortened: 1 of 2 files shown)


RMcVelia (approved):
looks fine to me!

PR #2402, opened 2025-04-15
Add a new row to the participant details table to show whether a trn …

EmmaFrith:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2758

The new admin participants #show screen does not include the TRN auto validated field which is present in the legacy admin users section. 

Make it clear whether a TRN is automatically or manually validated:

Screenshot 2025-04-23 at 16 01 05

### DIFF
--- a/spec/features/npq_separation/admin/users_spec.rb
+++ b/spec/features/npq_separation/admin/users_spec.rb
@@ -57,7 +57,7 @@ RSpec.feature "User administration", type: :feature do
         expect(summary_list).to have_summary_item("Email", user.email)
         expect(summary_list).to have_summary_item("Name", user.full_name)
         expect(summary_list).to have_summary_item("TRN", user.trn)
-        expect(summary_list).to have_summary_item("TRN validated", "No")
+        expect(summary_list).to have_summary_item("TRN status", "TRN not verified")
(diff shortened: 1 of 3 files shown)


Dom-Collins:
Jake has mentioned the language 'verified' rather than 'validated' tends to be used when discussing the status of TRNs. How about if we have this approach of displaying 3 possible options of either:

**Verified** (automatically)
**Verified** (manually)
**Not verified**

'Verified' would have a green tag component.
'Not verified' would have a red tag component.

WDYT?

EmmaFrith:
I think that's a great idea @Dom-Collins. Just playing around with it - do you mean like this?

| Scenario number    | Scenario example |
| -------- | ------- |
| 1  | Screenshot 2025-04-16 at 17 09 13    |
| 2 | Screenshot 2025-04-16 at 17 08 19   |
| 3    | Screenshot 2025-04-16 at 17 11 59    |

spec/features/npq_separation/admin/users_spec.rb
alkesh, line 1: we're going to need to add specs for when the TRN is verified

alkesh (approved):

jebw (approved):
LGTM :raised_hands:

PR #2403, opened 2025-04-17
Remove `plpgsql` from `azure_extensions`

peteryates:
### Context

This is causing errors when deploying as 'plpgsql' is being treated like an invalid value

@RMcVelia said this has happened before and Microsoft don't always make it clear when it's added/removed

### DIFF
--- a/terraform/application/database.tf
+++ b/terraform/application/database.tf
@@ -36,7 +36,7 @@ module "postgres" {
   azure_enable_monitoring        = var.enable_monitoring
   azure_enable_backup_storage    = var.enable_postgres_backup_storage
   server_version                 = "14"
-  azure_extensions               = ["btree_gin", "citext", "plpgsql", "pg_trgm"]
+  azure_extensions               = ["btree_gin", "citext", "pg_trgm"]
   azure_enable_high_availability = var.postgres_enable_high_availability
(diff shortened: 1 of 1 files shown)


terraform/application/database.tf
```diff
@@ -36,7 +36,7 @@ module "postgres" {
   azure_enable_monitoring        = var.enable_monitoring
   azure_enable_backup_storage    = var.enable_postgres_backup_storage
   server_version                 = "14"
-  azure_extensions               = ["btree_gin", "citext", "plpgsql", "pg_trgm"]
+  azure_extensions               = ["btree_gin", "citext", "pg_trgm"]
```
alkesh, line 39: do we also need to remove the line from `db/schema.rb`? - it currently contains `enable_extension "plpgsql"`

alkesh: we also have `plpgsql` below in the `postgres-snapshot`

peteryates: The current assumption is that Microsoft have trusted `plpgsql` so it doesn't need to be whitelisted and added manually, so this _should_ work. But we'll soon find out 🙈

alkesh (approved):

terraform/application/database.tf
peteryates: Ah totally missed it down there, well spotted!

PR #2413, opened 2025-04-23
2761 Institution autocomplete bug

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2761

Further discussion in the ticket.

The changes here:

- Adds missing i18n keys for the institution_identifier format validation, aligning these with the message that does exist for `choose_private_childcare_provider` — n.b. it’s hard to come up with a better message here because the error isn’t actually in the user’s control
- Enables the Sentry JavaScript SDK snippet in environments where it’s enabled server-side, so that if/when the autocomplete component next throws an error, we can collect more information on what’s going on

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -69,4 +69,11 @@ module ApplicationHelper
 
     govuk_tag(text: lead_provider_approval_status.humanize, colour:)
   end
+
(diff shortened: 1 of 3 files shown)


jebw (approved):
LGTM

PR #2421, opened 2025-04-24
[CPDNPQ-2807] Handle badly-formed participant_id parameter

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2807

Currently, if you send wonky `participant_id` parameters to an API endpoint that accepts them, you'll get a HTTP 500 Internal Server Error response (and an exception will be recorded by Sentry — example in the ticket).

This PR validates that the parameters are in the expected form (a comma-delimited list of UUIDs), and responds with 400 Bad Request and an informative error message if not.


jebw (approved):
LGTM :raised_hands:

PR #2433, opened 2025-04-29
Add change-delivery-partner documentation - v3 only

slawosz:
Missing fields documentation - for `declaration` response and `change-delivery-partner` endpoint:
image
image

### DIFF
--- a/spec/swagger_schemas/models/participant_declaration.rb
+++ b/spec/swagger_schemas/models/participant_declaration.rb
@@ -149,6 +149,13 @@ PARTICIPANT_DECLARATION = {
     nullable: true,
     example: "524df095-f9bf-4f9d-ba4c-772545a99e60",
   }
+  h[:v3][:properties][:attributes][:properties][:delivery_partner_name] = {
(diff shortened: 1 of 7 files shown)


spec/swagger_schemas/models/participant_declaration.rb
```diff
@@ -141,6 +141,36 @@
...
   }
+  h[:v3][:properties][:attributes][:properties][:delivery_partner_id] = {
+    description: "The delivery partner ID",
+    type: :string,
+    format: :uuid,
+    required: true,
+    nullable: true,
+    example: "524df095-f9bf-4f9d-ba4c-772545a99e60",
+  }
+  h[:v3][:properties][:attributes][:properties][:secondary_delivery_partner_id] = {
```
alkesh, line 152: is it worth ordering these new fields so they match the `DeclarationSerializer` so they are in the same order the API returns?
so:
delivery_partner_id
delivery_partner_name
secondary_delivery_partner_id
secondary_delivery_partner_name

alkesh (approved):

PR #2446, opened 2025-05-01
[CPDNPQ-2627] bulk operation - preparatory refactor

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2627

Pure refactor - ready for adding the new bulk operation - no change to functionality in this PR
* bulk operations controllers inherit from `BulkOperations:: BaseController`
* removed redundant word Bulk from Job classes
* all file reading moved into BulkOperation model
* row count updating moved into BulkOperation model


app/controllers/npq_separation/admin/bulk_operations/revert_applications_to_pending_controller.rb
```diff
@@ -1,44 +1,13 @@
...
-      @bulk_operation.update!(started_at: Time.zone.now, ran_by_admin_id: current_admin.id)
-      BulkOperation::BulkChangeApplicationsToPendingJob.perform_later(bulk_operation_id: @bulk_operation.id)
-      redirect_to :npq_separation_admin_bulk_operations_revert_applications_to_pending_index
-    end
-
-    def show
-      # empty method, because rubocop will complain in the before_action otherwise
-    end
-
+  class RevertApplicationsToPendingController < NpqSeparation::Admin::BulkOperations::BaseController
```
jebw, line 2: Should be enough to inherit from `BaseController` - you're already in the `BulkOperations` namespace?

app/models/bulk_operation.rb
```diff
@@ -1,34 +1,54 @@
...
 
-  def check_format(string)
-    CSV.parse(string) do |row|
-      if row.size > 1
-        errors.add(:file, :invalid)
-        break
-      end
-    end
+  def attached_file
+    attachment_changes["file"].attachable
```
jebw, line 35: Is there a scenario where `attachment_changes["file"]` might be `nil` ? If so can this be changed to `&.attachable`

jebw: I notice lower down in `row_count` it is checking for whether `attachment_changes["file"]` is nil?

app/services/bulk_operation/bulk_revert_applications_to_pending.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class BulkOperation::BulkRevertApplicationsToPending
```
jebw, line 3: If we're already renaming this class, lets drop the superfluous `Bulk` prefix - its already under the `BulkOperations` namespace so I don't think we need the prefix to clarify

jebw (approved):
LGTM with some small amends in the comments

Thanks for splitting this out, it greatly simplified the code review

PR #2450, opened 2025-05-02
[CPDNPQ-2767] admin console - edit and delete adjustments

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2767

Adding the ability to edit and delete adjustments.

There are 3 different flows to consider:
* editing and deleting adjustments whilst creating them - the adjustments index page only shows the adjustments created since leaving the statement page
* editing and deleting adjustments by clicking on the 'Change or remove' link on the statement page - the adjustments index page shows all adjustments
* creating adjustments by clicking the 'Change or remove` link on the statement page, and then choosing the 'Yes' radio button - this flow means all the adjustments are shown on the adjustments index page, and also the cancel links go to back to the adjustments index page

### DIFF
--- a/app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
+++ b/app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
@@ -2,6 +2,8 @@
 
 class NpqSeparation::Admin::Finance::Statements::AdjustmentsController < NpqSeparation::AdminController
   before_action :set_statement
+  before_action :find_adjustment, only: %i[edit update delete destroy]
+  before_action :set_show_all_adjustments
 
   def new
     @adjustment = @statement.adjustments.new
@@ -12,7 +14,7 @@ class NpqSeparation::Admin::Finance::Statements::AdjustmentsController < NpqSepa
       adjustment_params.merge(created_adjustment_ids: session[:created_adjustment_ids], statement: @statement),
     )
 
-    if @create_adjustment_form.save_form
+    if @create_adjustment_form.save_adjustment
       session[:created_adjustment_ids] = @create_adjustment_form.created_adjustment_ids
       redirect_to npq_separation_admin_finance_statement_adjustments_path(@statement)
     else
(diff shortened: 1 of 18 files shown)


app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
```diff
@@ -2,19 +2,27 @@
...
+  before_action :find_adjustment, only: %i[edit update delete destroy]
+  before_action :set_show_all_adjustments, except: %i[create add_another]
 
   def new
     @adjustment = @statement.adjustments.new
+    @cancel_url = if @show_all_adjustments
+                    npq_separation_admin_finance_statement_adjustments_path(@statement, show_all_adjustments: @show_all_adjustments)
+                  else
+                    npq_separation_admin_finance_statement_path(@statement)
+                  end
```
jebw, line 14: Could just add a private method in the controller and mark it using `helper_method :cancel_url` ?

app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
```diff
@@ -2,19 +2,27 @@
...
   def create
     @create_adjustment_form = Admin::Adjustments::CreateAdjustmentForm.new(
       adjustment_params.merge(created_adjustment_ids: session[:created_adjustment_ids], statement: @statement),
     )
 
-    if @create_adjustment_form.save_form
+    if @create_adjustment_form.save_adjustment
       session[:created_adjustment_ids] = @create_adjustment_form.created_adjustment_ids
-      redirect_to npq_separation_admin_finance_statement_adjustments_path(@statement)
+      show_all_adjustments = params[:adjustment][:show_all_adjustments] == "true"
```
jebw, line 24: This line is also used below? Could it be moved to its own `show_all_adjustments?` method?

spec/components/npq_separation/admin/adjustments_table_component_spec.rb
```diff
@@ -1,14 +1,17 @@
 require "rails_helper"
 
 RSpec.describe NpqSeparation::Admin::AdjustmentsTableComponent, type: :component do
-  subject { render_inline described_class.new(adjustments: adjustments, show_total: show_total) }
+  include Rails.application.routes.url_helpers
```
jebw, line 4: Rather than pull all the url_helpers (and there are many) into the spec class, how about

```ruby
let(:urls) { Rails.application.routes.url_helpers }
```

then just referencing `urls.some_path` lower down

spec/forms/admin/adjustments/create_adjustment_form_spec.rb
```diff
@@ -3,31 +3,28 @@
 require "rails_helper"
 
 RSpec.describe Admin::Adjustments::CreateAdjustmentForm, type: :model do
-  subject(:form) { described_class.new(created_adjustment_ids:, statement:, description:, amount:) }
+  let(:form) { described_class.new(created_adjustment_ids:, statement:, description:, amount:) }
```
jebw, line 6: I see no advantage to this

the existing version means you can add a short `it { is_expected...... }` spec, the second doesn't ?

jebw (approved):
Looks great. I've a couple of minor suggestions only if they make sense, the only item I'm not keen on is pulling all of url_helpers into classes?

app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
alkesh: maybe - the values of @cancel_url are different for the `new` and `edit` methods - so the helper method would end up being `cancel_new_url`?
I'm not sure it's worth creating a helper method when it's only used once.

app/controllers/npq_separation/admin/finance/statements/adjustments_controller.rb
alkesh: not really - the two are different - different form names:
line 24: `show_all_adjustments = params[:adjustment][:show_all_adjustments] == "true"`
line 64: `show_all_adjustments = params[:add_another_form][:show_all_adjustments] == "true"`

spec/components/npq_separation/admin/adjustments_table_component_spec.rb
alkesh: 👍🏽  ah yes - I remember this from another PR 😅

spec/forms/admin/adjustments/create_adjustment_form_spec.rb
alkesh: no - for each of the describes below, the subject is set correctly to `form.save_adjustment`.
what I should do, is actually use `it { is_expected...... }` in those specs - will fix.

PR #2453, opened 2025-05-06
Npq 2323 use delete review app GitHub action

pasitconsulting:
### Context

[What has been changed?]

failures:


.github/workflows/destroy_review_app.yml
```diff
@@ -3,72 +3,59 @@ name: Delete review
...
-
-          if [ -n "$state_file_status" ]; then
-            echo "TF_STATE_EXISTS=true" >> $GITHUB_ENV
+          if [ ${{ github.event_name }} == 'workflow_dispatch' ]; then
+            PR_NUMBER=${{ github.event.inputs.pr_number }}
+          else
+            PR_NUMBER=${{ github.event.pull_request.number }}
           fi
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+          echo "ENVIRONMENT=pr-${PR_NUMBER})" >> $GITHUB_ENV
```
jebw, line 37: `ENVIRONMENT` feels very generic for a env var name with chance of clashes, etc? How about something something like `AZURE_ENVIRONMENT` ?

jebw (commented):
Looks good, can you populate the PR template? Couple of queries inline as well, thanks

PR #2457, opened 2025-05-07
Add missing specs

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2744

Adding missing specs to the `User` model, and removing a few unused methods.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -245,4 +245,116 @@ RSpec.describe User do
       end
     end
   end
+
+  describe ".find_by_get_an_identity_id" do
+    let(:uid) { SecureRandom.uuid }
+    let!(:user) { create(:user, :with_get_an_identity_id, uid:) }
+
+    it "returns the user with matching uid from the with_get_an_identity_id scope" do
+      expect(User.find_by_get_an_identity_id(uid)).to eq(user)
+    end
+
+    it "returns nil if no user matches the uid" do
+      expect(User.find_by_get_an_identity_id("nonexistent-uid")).to be_nil
+    end
+  end
+
+  describe "#get_an_identity_user" do
+    let(:user) { create(:user, :with_get_an_identity_id, uid:) }
+    let(:uid) { SecureRandom.uuid }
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -245,4 +245,116 @@
       end
     end
   end
+
+  describe ".find_by_get_an_identity_id" do
+    let(:uid) { "foobar" }
```
jebw, line 250:
```ruby
SecureRandom.uuid
```

spec/models/user_spec.rb
```diff
@@ -245,4 +245,116 @@
...
+    let!(:user) { create(:user, :with_get_an_identity_id, uid:) }
+
+    it "returns the user with matching uid from the with_get_an_identity_id scope" do
+      expect(User.find_by_get_an_identity_id(uid)).to eq(user)
+    end
+
+    it "returns nil if no user matches the uid" do
+      expect(User.find_by_get_an_identity_id("nonexistent-uid")).to be_nil
```
jebw, line 258:
```ruby
User.find_by_get_an_identity_id(nonexistant_uid)
```

jebw (commented):
Looks great, only real point is replacing the stubbed GAI uuid's (which are "foobar") with valid ones, ie using `SecureRandom.uuid` ?

spec/models/user_spec.rb
```diff
@@ -245,4 +245,116 @@
       end
     end
   end
+
+  describe ".find_by_get_an_identity_id" do
+    let(:uid) { "foobar" }
+    let!(:user) { create(:user, :with_get_an_identity_id, uid:) }
```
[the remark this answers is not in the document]
slawosz: Otherwise the finder wont work.

spec/models/user_spec.rb
slawosz: If the variable is not used anywhere, I personally prefer to use it inline. However, happy to move it to let - but then is there a some sort of guidance, or maybe we could configure rubocop to catch this kind of things?

jebw (approved):
LGTM

PR #2458, opened 2025-05-07
[CPDNPQ-2809] Fix registration closed banner link

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2809

The banner (service name in the header) link currently links to /registration/closed, which has different content to /registration_closed — the ticket specifies we prefer the latter.

This PR is a quick fix for that. I removed the conditional `service_url` assignment because `RegistrationWizardController` already handles this in `before_action :registration_closed`.

n.b. I looked at removing the `/registration/closed` RegistrationWizard step altogether, but it's tangled into various spots so will take a bit more effort than this ticket allows for.


EmmaFrith (approved):
Tested and works as expected

PR #2463, opened 2025-05-08
[CPDNPQ-2718] Detail unicode in login Sentry exceptions

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2718

We have Sentry exceptions ([example](https://dfe-teacher-services.sentry.io/issues/5998221231/?alert_rule_id=14507801&alert_type=issue&environment=production&notification_uuid=21b42aee-69ba-44a5-9ad3-73eb957b1328&project=5817541)) suggesting that after signing in with DfE Identity, users who are returned with multibyte characters in their email addresses cannot be persisted, so they're unable to use our service despite having successfully authenticated.

Initially this was hypothesised to be a collation problem, but multibyte values can be persisted to the database with no problems, the validation error is caused by the `notify_email` validation on `User#email`.

We had supposed that problematic email addresses used IDN hostnames, but investigating a specific user's UID with the Identity team revealed no visible unicode characters in any of the email addresses they'd used, so this PR follows on from #2459 by including actual unicode codepoints in the Sentry notification, so that we can understand the cause if we do encounter this again. (We can't send the full email address due to PII policy.)

### DIFF
--- a/app/controllers/omniauth_controller.rb
+++ b/app/controllers/omniauth_controller.rb
@@ -23,7 +23,6 @@ class OmniauthController < Devise::OmniauthCallbacksController
       # TODO: need a feature test for the error scenario:
       # - email has already been taken - but somehow user_id_with_clashing_email is nil
       user_id_with_clashing_email = User.find_by(email: provider_data.info.email)&.id
-      multibyte_email = (provider_data.info.email.chars.count != provider_data.info.email.bytes.count)
 
       send_error_to_sentry(
         "Could not persist user after omniauth callback",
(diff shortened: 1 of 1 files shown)


app/controllers/omniauth_controller.rb
```diff
@@ -70,6 +69,14 @@ def failure
 
 private
 
+  def multibyte_email_characters(email)
+    email.each_char.filter_map.with_index do |c, i|
+      next unless c.bytesize > 1
+
+      "Character #{i} = U+" + c.ord.to_s(16).upcase
```
alkesh, line 76: maybe worth a comment on how to convert from the outputted hex value in sentry, to the initial unicode string:
hex_value_string.to_i(16).chr(Encoding::UTF_8)

alkesh (approved):

PR #2464, opened 2025-05-08
Increase Flipper limit actors

slawosz:
### Context

We have hit the default flipper limit of 100 on the closed registration route:
https://dfe-teacher-services.sentry.io/issues/6593675525/?referrer=slack&notification_uuid=d42a8537-a82a-4f3b-877a-094ce3abc124&environment=production&alert_rule_id=14507801&alert_type=issue

Increase the limit to 1000

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -118,4 +118,5 @@ Rails.application.configure do
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
   config.session_store :active_record_store, key: "_npq_registration_session", secure: true, expire_after: 2.weeks
+  config.flipper.actor_limit = 1000
 end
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -118,4 +118,5 @@
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
   config.session_store :active_record_store, key: "_npq_registration_session", secure: true, expire_after: 2.weeks
+  config.flipper.actor_limit = 1000
```
alkesh, line 121:
```suggestion
  # increase the actor limit from the default of 100
  # this limits the number of users that can use the late registration route
  config.flipper.actor_limit = 1000
```
worth adding a comment here explaining why we're increasing the limit

PR #2475, opened 2025-05-12
[CPDNPQ-2721] send DfE::Analytics custom events for API requests

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2721

switch from bespoke writing to bigquery, to using DfE::Analytics custom events

New custom event type `persist_api_request`.

before changes:
API requests written using bigquery client directly, to `npq_api_requests_production` table in `npq_api_requests` dataset.

after changes:
API requests written using DfE::Analytics custom events, to `events` table in `npq_events_production` dataset, with namespace `npq`.
Also - `Crons::ScheduleUploadAnalyticsReportJob` has been removed, as it's not used, and is duplicate data - there will be a future ticket to remove the `Report` model and `Exporters::AnalyticsReport`.


spec/requests/dfe_analytics_spec.rb
```diff
@@ -3,13 +3,49 @@
 require "rails_helper"
 
 RSpec.describe "DfE Analytics", type: :request do
-  before { Flipper.enable(Feature::DFE_ANALYTICS_ENABLED) }
+  let(:current_lead_provider) { create(:lead_provider) }
 
-  it "does send DFE Analytics web request events" do
-    expect { get root_path }.to have_sent_analytics_event_types(:web_request)
+  context "when DfE Analytics is enabled" do
+    before { Flipper.enable(Feature::DFE_ANALYTICS_ENABLED) }
```
jebw, line 9: Is it possible to either

a) Just stub this, ie either 

```ruby
before { allow(Feature).to receive(:dfe_analytics_enabled?).and_return(true) }
```

or possibly

```ruby
before do
  allow(Flipper).to receive(:enabled).and_call_original
  allow(Flipper).to receive(:enabled).with(Feature::DFE_ANALYTICS_ENABLED).and_return(true)
end
```

b) or failing that, put it back to how it was after in an `after` block - otherwise this is state which will persist across tests leading to flaky behaviour

spec/requests/dfe_analytics_spec.rb
```diff
@@ -3,13 +3,49 @@
...
+
+    it "does not send DfE Analytics web request events for healthcheck" do
+      expect { get "/healthcheck" }.not_to have_sent_analytics_event_types(:web_request)
+    end
   end
 
-  it "does not send DFE Analytics web request events for healthcheck" do
-    expect { get "/healthcheck" }.not_to have_sent_analytics_event_types(:web_request)
+  context "when DfE Analytics is disabled" do
+    before { Flipper.disable(Feature::DFE_ANALYTICS_ENABLED) }
```
jebw, line 37: Same commend as above - just trying to head off side effects from tests modifying state

jebw (approved):
Looks awesome - approved - can the Flipper flags be replaced with stubbing (or at least put back to how it was after the test)

spec/requests/dfe_analytics_spec.rb
alkesh: I don't think this is necessary - since v1.2.0, Flipper uses a memory adapter in the test environment (https://www.flippercloud.io/docs/testing), so each test is run with a fresh instance of Flipper and all features disabled.

PR #2488, opened 2025-05-15
2581 workplaces view with search

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2581

Redesign the school view and add search capability. Rename `Schools` to `Workplaces`.

### Search using the school name
image

### Search using the school `URN`

image


app/helpers/application_helper.rb
```diff
@@ -76,4 +76,8 @@ def sentry_javascript_tag
 
     javascript_include_tag "https://js.sentry-cdn.com/#{dsn}.min.js", crossorigin: "anonymous"
   end
+
+  def join_with_commas(*args)
+    args.select(&:present?).join(", ")
+  end
```
jebw, line 82: `#to_sentence` ?

app/views/npq_separation/admin/schools/index.html.erb
```diff
@@ -1,18 +1,49 @@
...
+<p class="govuk-body">Search for a workplace to find its:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>ID</li>
+  <li>Unique Reference Number (URN)</li>
+  <li>UK Provider Reference Number (UKPRN)</li>
+  <li>local authority</li>
+  <li>address</li>
+</ul>
+
+<div style="background-color: #f3f2f1; padding: 20px 15px; margin-bottom: 30px;">
```
jebw, line 11: Inline CSS will fail CSP checks, can this be moved to a class - I'm not clear on why this form needs different styling to the other filter forms though? Is it possible to re-use the class from those forms?

app/views/npq_separation/admin/schools/index.html.erb
```diff
@@ -1,18 +1,49 @@
...
+  <%= form_with method: :get do |f| %>
+    <%= f.govuk_text_field(
+          :q,
+          value: params[:q],
+          hint: { text: "Search by entering the workplace name or its Unique Reference Number (URN)" },
+          label: { text: "Find a workplace", tag: "h1", size: "m" },
```
jebw, line 17: Why H1, there's already an H1 on the page? I'd expect an `m` should be a h3 (or possible h2) ?

jebw (commented):
Looks good, only nit is the inline css - I'm busy pulling that out of other parts of the codebase at the same time.

Looks like it should be possible to use the `admin-search-box` css class?

app/helpers/application_helper.rb
slawosz: Not really:
```ruby
["foo", "", "baz"].to_sentence(last_word_connector: ', ')
# => "foo, , baz"
```

jebw (approved):
LGTM

PR #2499, opened 2025-05-16
[CPDNPQ-2564] user overview page design update

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2564

New design for user overview page - based on this [prototype](https://npq-prototype.herokuapp.com/admin-v3/registration-1/overview).

New design
Also - made "TRN verified" consistent across the admin console (replaced instances of "TRN validated").


spec/views/npq_separation/admin/users/_application.html.erb_spec.rb
alkesh, line 1: Removing this spec, because the original reason this was added was for showing the School URN and School UKPRN when a school was not present - we now do not show either of these fields.
The remaining fields are covered by the feature spec.

app/views/npq_separation/admin/users/_application.html.erb
```diff
@@ -1,51 +1,38 @@
...
+        row.with_value(text: application.lead_provider.name)
+      end
+
+      sl.with_row do |row|
+        row.with_key(text: "Eligible for funding")
+        row.with_value(text: boolean_tag(application.eligible_for_funding))
+      end
+
+      sl.with_row do |row|
+        row.with_key(text: "Provider Approval Status")
```
EmmaFrith, line 23: On the prototype this looks like it should be: `Provider approval status` (lower case)

app/views/npq_separation/admin/users/_application.html.erb
```diff
@@ -1,51 +1,38 @@
...
-  end
-%>
+<%- title = "#{application.course.name} registration" -%>
+<%- actions = [govuk_link_to("View full registration", npq_separation_admin_application_path(application))] -%>
+
+<div class="admin-application-card">
+  <%=
+    govuk_summary_list(card: { title:, actions: }) do |sl|
+      sl.with_row do |row|
+        row.with_key(text: "NPQ Course")
```
EmmaFrith, line 8: On the prototype this looks like it should be `NPQ course`(lower case)

app/views/npq_separation/admin/users/_application.html.erb
```diff
@@ -1,51 +1,38 @@
...
+        row.with_value(text: boolean_tag(application.eligible_for_funding))
+      end
+
+      sl.with_row do |row|
+        row.with_key(text: "Provider Approval Status")
+        row.with_value(text: lead_provider_approval_status_badge(application.lead_provider_approval_status))
+      end
+
+      sl.with_row do |row|
+        row.with_key(text: "Funded Place")
```
EmmaFrith, line 28: ditto looks like it should be `Funded place` from the prototype

app/views/npq_separation/admin/users/_application.html.erb
```diff
@@ -1,51 +1,38 @@
...
+        row.with_value(text: boolean_red_green_nil_tag(application.funded_place))
+      end
+
+      sl.with_row do |row|
+        row.with_key(text: "Registration submission date")
+        row.with_value(text: application.created_at.to_fs(:govuk_short))
+      end
+    end
```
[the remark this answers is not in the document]
EmmaFrith: The ticket says 'Funding review status' doesn't need to be included yet but is 'Latest declaration' missing?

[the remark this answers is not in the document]
alkesh: "Latest declaration" has moved into its own ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2862

app/assets/stylesheets/admin.scss
```diff
@@ -177,3 +177,9 @@
 .bulk-operation-action-button-group {
   margin-bottom: 0;
 }
+
+.admin-application-card {
+  h2.govuk-summary-card__title {
+    font-size: 1.5rem;
+  }
+}
```
jebw, line 185: Whats the motivation for not following the design system styling?

jebw: Could this do something like `@include`, `@extend` - can't remember which - one of the override classes from the design system?

https://design-system.service.gov.uk/styles/font-override-classes/

jebw (approved):
Approved - code looks great

I'm not really following why we need a different heading size for these particular application cards? The standard size would consistent with the rest of the site?

alkesh:
> I'm not really following why we need a different heading size for these particular application cards? The standard size would consistent with the rest of the site?

I'm following the prototype - and I do think it looks better.

PR #2500, opened 2025-05-16
Move 'Admins' to new admin console

EmmaFrith:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2769

We're gradually moving sections from the old admin console to the new admin console, so that we can eventually phase out the old console. 

This PR moves the 'Admins' functionality to the new console. This includes:

- moving the code/content to the relevant folders/ namespaces
- updating the controllers so that they inherit from NpqSeparation::AdminController and use the 'require_super_admin' before_action
- removing the 'Admins' link from the old navbar and adding it to the new navbar, and ensuring this only shows for super admins
- updating link paths in the views and helpers 

| Before    | After |
| -------- | ------- |
| Screenshot 2025-05-16 at 15 01 26  | Screenshot 2025-05-16 at 15 03 31    |


app/controllers/npq_separation/admin/super_admins_controller.rb
```diff
@@ -1,13 +1,12 @@
-class Admin::SuperAdminsController < SuperAdminController
+class NpqSeparation::Admin::SuperAdminsController < NpqSeparation::AdminController
+  before_action :require_super_admin
```
jebw, line 2: Can you add a line of whitespace after this line before the `update` line :+1:

app/controllers/npq_separation/admin/super_admins_controller.rb
```diff
@@ -1,13 +1,12 @@
...
     @admin = Admin.find(params[:id])
-
     if @admin.update(super_admin: true)
       flash[:success] = t(".success", email: @admin.email)
     else
       flash[:error] = t(".failure", email: @admin.email)
     end
-
-    redirect_back(fallback_location: admin_admins_path)
+    redirect_back(fallback_location: npq_separation_admin_admins_path)
```
jebw, line 10: Can you add whitespace line before this one

jebw (approved):
Looks pretty much perfect :raised_hands:

I've approved but can you drop the unneeded pagy line and add the missing whitespace before merging, thanks

PR #2526, opened 2025-05-20
[CPDNPQ-2818] Admin side navigation

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2818

Currently, side navigation is defined globally in `app/components/npq_separation/navigation_structures/admin_navigation_structure.rb`. We want side navigation to:

a) have headings
b) use variables set by the controller

#### Approach: 

a) Although the ticket suggests importing the MoJ "Side navigation" component to get headings, we already have an implementation of the [X-GOVUK prototype sub navigation component](https://x-govuk.github.io/govuk-prototype-components/sub-navigation/#sub-navigation-preview) providing the existing side navigation, so this PR extends that component to allow setting and displaying the heading

b) Rather than introduce more coupling between the current request and the navigation structure, we can keep the global structure as-is, and support extra controller-dependent content in the rare cases that it's needed by just rendering another instance of the component inside `content_for :side_navigation` (which also makes reusing such content across actions/controllers easy via a partial or helper).

Although this is perhaps a bit verbose, it gets the job done cleanly, with minimal ✨ magic ✨, and it's tailored to the view. (Eventually dropping the `NpqSeparation` namespace will tidy this a fair bit too!)

---

I did explore implementing the component block style, something like:

```ruby
render NpqSeparation::SubNavigationComponent.new(request.path) do |nav|
  nav.with_heading text: @application.user.full_name, visible: true
  nav.with_node({
    name: "#{@application.course.name}, #{@application.created_at.to_date.to_fs(:govuk)}",
    href: npq_separation_admin_application_path(@application),
    prefix: "/npq-separation/admin/applications"
  })
  # etc etc
end
```

But it just didn't seem worth the effort right now. This or some other terser abstraction may become obviously worthwhile later if we end up using a few of these, but OTOH I can only see this used in one place in the prototype at the moment. 

---

#### Implementation

This PR also implements the new side navigation on the admin "Applications" screens, as shown in the ticket's prototype link (although the 2nd and 3rd application-level navigation destinations don't yet exist, so it's just the one for now):

image


alkesh:
do we want to style the side navigation, so that it matches the prototype?

jebw (commented):
This looks good but the 'special casing' the first node in the list feels a bit like the magic you were keen to avoid

Nodes logically are 'under' a particular heading so how about adding a `nodes:` attribute to the `Heading` class

Then you'd just have an array of objects which are some mixture of `Heading` and/or `Node`

That would allow for a side nav multiple headings, each with their own nodes under them. 

The sidenav on the page below feels like its more inline with the design for prototypes applications page

https://govuk-components.netlify.app/components/tabs/

jebw (approved):

PR #2536, opened 2025-05-22
Add workplace search for all workplace models

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2582


app/services/admin_service/workplace_search.rb
```diff
@@ -0,0 +1,110 @@
...
+        return records
+      end
+
+      # If we reached this point it means that current scope is shorter than offset.
+      extra_offset += scope.count
+      i += 1
+    end
+
+    [] # Fallback when no records could be found
+  end
```
jebw, line 109: The above code seems to work well but is a little hard hard to follow - could the outer while loop could be replaced with `each_index`, ? Or alternatively the `j` and the `i` be given names

If I'm understanding the above code correctly, then I think the following does the same thing?

```ruby
def count_for(scope)
  @scope_counts[scope] ||= scope.count
end

def find_scopes
  scopes.each.with_object([]) do |scope, records|
    if count_for(scope) > @limit + @offset
      records += scope.limit(@limit).offset(@offset)
      return records
    elsif count_for(scope) < @offset
      @offset -= count_for(scope)
    else
      records += scope.offset(@offset)
      limit -= records.length
    end
  end
end
```

app/services/admin_service/workplace_search.rb
```diff
@@ -0,0 +1,110 @@
...
+    [
+      schools_scope,
+      private_childcare_provider_scope,
+      local_authority_scope,
+    ]
+  end
+
+  # Return properly paginated scopes
+  def find_scopes
+    # Extra offset is to compensate for the previous scouts count.
```
jebw, line 69: I'm not clear on what is meant by 'scouts count' thats not a term I'm familiar with

app/services/admin_service/workplace_search.rb
```diff
@@ -0,0 +1,110 @@
...
+        return scope.offset(@offset - extra_offset).limit(@limit)
+      # case where we need to display more than one different scope in one page
+      elsif scope.count + extra_offset > @offset && scope.count + extra_offset < @offset + @limit
+        remaining_records_count = (scope.count + extra_offset) % @limit
+
+        records = scope.offset(@offset - extra_offset).limit(remaining_records_count).to_a
+        j = i
+        # we need to fill current page with scopes
+        while (next_scope = scopes[j + 1]) && records.count < @limit
+          extra_records = @limit - records.count
```
jebw, line 95: `extra_records` implies  it holds records when its actually a counter for the remaining limit if I'm reading this correctly? Can it be named clearer

jebw (commented):
The code seems to work well in my initial testing and the algorithm implemented solves the challenging of merging the results from 3 tables. The code is a little hard to follow though - whilst extensively commented I suspect some refactoring would reduce the need for some of the comments?

If my reading of the algorithm is correct then I think my suggested code does the same thing?

spec/requests/npq_separation/admin/schools_controller_spec.rb
```diff
@@ -13,21 +13,4 @@
...
-    end
-
-    it { is_expected.to have_http_status(:ok) }
-
-    context "when the school cannot be found", :exceptions_app do
-      let(:school_id) { -1 }
-
-      it { is_expected.to have_http_status(:not_found) }
-    end
-  end
```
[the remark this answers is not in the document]
slawosz: We don't use show view anymore. Deleted different files as well.

jebw (approved):
Looks great, nice bit of logic! 

There's a couple of really minor tweaks, if those could get done before merging that'd be great thanks

PR #2539, opened 2025-05-23
[CPDNPQ-2734] add link to allow notes to be changed from any application

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2734

All notes to be added/edited from any application

### DIFF
--- a/app/controllers/npq_separation/admin/applications/notes_controller.rb
+++ b/app/controllers/npq_separation/admin/applications/notes_controller.rb
@@ -6,12 +6,14 @@ module NpqSeparation
       class NotesController < NpqSeparation::AdminController
         before_action :set_application
 
-        def edit; end
+        def edit
+          @return_path = request.referer || npq_separation_admin_application_path(@application)
+        end
 
         def update
           if @application.update(notes_params)
             flash[:success] = "Notes updated."
-            redirect_to npq_separation_admin_application_review_path(@application)
+            redirect_to return_path_param
           else
             render :edit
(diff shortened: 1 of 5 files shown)


app/controllers/npq_separation/admin/applications/notes_controller.rb
```diff
@@ -6,12 +6,14 @@ module Applications
...
+          @return_path = request.referer || npq_separation_admin_application_path(@application)
+        end
 
         def update
           if @application.update(notes_params)
             flash[:success] = "Notes updated."
-            redirect_to npq_separation_admin_application_review_path(@application)
+            redirect_to return_path_param
           else
             render :edit
```
EmmaFrith, line 18: Should we show an error message here to explain why the update failed?

jebw: @EmmaFrith so in the above code

if it succeeds then it will do the redirect back to `return_path_param` but if it fails it will show the 'edit' screen again

The data object being used for the form on the edit screen will have validation errors set on it, so reshowing that screen will now automatically show those error messages above the form fields

EmmaFrith: Thanks @jebw :) so is that what this bit is about <%= f.govuk_error_summary %> in the edit.html.erb file?

jebw (approved):
LGTM :+1:

PR #2552, opened 2025-05-29
Delete users from legacy admin console

EmmaFrith:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2757

**Context** 

We're removing unneeded sections from the old admin console. 

This PR is for removing the 'Users' section of the old admin console. 

'Users' is now covered by the 'Participants' section on the new console. 

Screenshot 2025-05-29 at 12 47 13


jebw (approved):
LGTM

PR #2565, opened 2025-06-02
Remove unsynced applications from legacy admin console

EmmaFrith:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2757

We're removing unneeded sections from the old admin console.

This PR is for removing the 'Unsynced applications' section of the old admin console.


jebw (approved):
LGTM :raised_hands:

PR #2583, opened 2025-06-05
[CPDNPQ-2901] use latest application when creating declaration

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2901

use latest application (based on `created_at`) when creating a declaration, instead of the oldest.


jebw (changes requested):
I think the code for checking declarations also needs changing so that if there are existing declarations against another older application, those do not block declaration against the newer application

Needs some specs adding for that scenario as well I think

jebw (approved):
LGTM :+1:

PR #2592, opened 2025-06-09
[CPDNPQ-2904] course provider pages

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2904

Instead of showing a list of statements for each course provider,
show delivery partners for each course provider / cohort.
Updated side nav to match prototype.


jebw (approved):
LGTM but the extra nesting on the div seems unneeded since that side column is only for the nav anyway?

PR #2604, opened 2025-06-10
[CPDNPQ-2939] move declarations

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2939

Rake task to move UCL declarations - from Jan 2026 to June 2025


spec/services/one_off/migrate_declarations_between_statements_spec.rb
```diff
@@ -293,6 +293,7 @@
 
         expect(logged_output).to eq([
           "Migrating declarations from #{from_statement.year}-#{from_statement.month} to #{to_statement.year}-#{to_statement.month} for 1 providers",
+          "Migrating 1 declarations, with IDs:\n#{declaration.ecf_id}",
```
jebw: Looks like this is causing the specs to fail?

PR #2612, opened 2025-06-12
[CPDNPQ-2868] admin console consistent cohort date format

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2868

Admin console cohort displayed as: `2025 to 2026` instead of the old `2025/26` format.


jebw (commented):
If we're dropping the 'non-full' version everywhere - can we not make format_cohort follow the 'full' format and drop the `_full` helper which is only used in 1 location?

PR #2625, opened 2025-06-16
[CPDNPQ-2398] make ApplicationController default to requiring authentication

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2398

* `ApplicationController` raises an error if used directly
* `LoggedInController` has the `before_action :authenticate_user!`
* controllers can inherit from `PublicPagesController` if they are supposed to be public
* make `API::V1::GetAnIdentity::WebhookMessagesController` inherit from `ActionController::API` instead of `ApplicationController`


app/controllers/omniauth_controller.rb
```diff
@@ -1,5 +1,6 @@
 class OmniauthController < Devise::OmniauthCallbacksController
   skip_before_action :verify_authenticity_token, only: %i[tra_openid_connect]
+  skip_before_action :authenticate_user!, only: %i[tra_openid_connect failure]
```
alkesh, line 3: this is because `Devise::OmniauthCallbacksController` inherits from `ApplicationController`

app/controllers/errors_controller.rb
```diff
@@ -1,4 +1,4 @@
-class ErrorsController < ApplicationController
+class ErrorsController < PublicPagesController
```
jebw, line 1: I think this one should probably inherit from `ActionController::Base` since its an errors controller we don't think we would ever want any of the hooks, or other changes from ApplicationController running

spec/controllers/application_controller_spec.rb
```diff
@@ -37,5 +31,11 @@ def index
 
       expect(Sentry).to have_received(:set_user).with(id: user.id)
     end
+
+    it "sets caching headers" do
+      get :index
+
+      expect(response.headers).to include "Cache-Control" => "no-store"
+    end
```
jebw, line 39: This spec isnt specific to whether they are signed in or not, so I think its fine outside of the contexts?

jebw (dismissed):
So ideally I'd like the developer to be explicitly choosing what to inherit from to determine behaviour? Reading this, I think if a member of the public is signed in, then an admin screen inheriting from `ApplicationController` would 'pass' the `authenticate_user!` step. Inheriting from ApplicationController should just raise an exception

I think this probably needs 4 scenerios

```ruby
class ApplicationController
  before_action :authenticate_user!

   def authenticate_user!; raise "some error"; end
end

class PublicPagessController < ApplicationController
  def authenticate_user!; end
end

class LoggedInController < ApplicationController
  def authenticate_user!
    # check that is currently in ApplicationController
  end
end

class AdminController < ApplicationController
  skip_before_action :authenticate_user!

  # do admin auth checks
end
```

spec/controllers/application_controller_spec.rb
alkesh: ah yes - I'm going to duplicate this test for both signed in, and not signed in - to ensure we have the correct ordering of before actions.

app/controllers/errors_controller.rb
alkesh: it looks like this one needs to be PublicPagesController, because we currently still show the shared header on the errors page, and this relies on having `current_admin` and `current_user` from `ApplicationController`.

app/controllers/private_childcare_providers_controller.rb
```diff
@@ -1,4 +1,4 @@
-class PrivateChildcareProvidersController < ApplicationController
```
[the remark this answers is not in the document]
alkesh: I've made this inherit from `PublicPagesController` now, so it's the same behaviour as before.

PR #2655, opened 2025-06-24
[CPDNPQ-3036] Add participant_id currency validation

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3036

When a lead provider attempts to use a retired `participant_id` (i.e. a user who has been archived & had their applications moved to a different user), they will get an error response that says the `participant_id` does not exist. This is confusing and can lead to support queries.

In this scenario it would be helpful to more specifically inform providers that the `participant_id` has changed, so that they can understand how to rectify the situation.

#### Notes

- I've only applied this to the current V3 version of the API, especially given older versions should be deprecated in the near future.

- I believe this covers all API endpoints that take a single `participant_id` as a parameter, but I haven't included endpoints that accept multiple IDs due to complexity in adding this error into an otherwise valid response. We could look at this in future but it seems like the existing behaviour for these multiple-ID endpoints is probably acceptable, at least for now. The issues we've had reported around this have been related to single-ID endpoints such as creating a declaration.

- Technically any authenticated lead provider can see the participant ID change response, even if the ID belongs to a different provider, but the risk here seems nominal since they'd need to know the old ID in the first place, and they won't be able to do anything with the new ID anyway. With this in mind it didn't seem worth the extra effort to conceal this based on lead provider.

### DIFF
--- a/app/services/declarations/create.rb
+++ b/app/services/declarations/create.rb
@@ -15,7 +15,7 @@ module Declarations
     attribute :secondary_delivery_partner_id
 
     validates :lead_provider, presence: true
-    validates :participant_id, presence: true, participant_id_currency: true
+    validates :participant_id, presence: true, participant_id_change: true
     validates :participant, participant_presence: true, participant_not_withdrawn: true
     validates :course_identifier, course_for_participant: true
     validates :declaration_date, declaration_date: true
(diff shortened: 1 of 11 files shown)


jebw (approved):
LGTM

PR #2661, opened 2025-06-26
[CPDNPQ-3038] Resolve docker build annotation

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3038

The goal is to remove the Docker build annotations in GitHub actions e.g.:
image

As per [Docker docs](https://docs.docker.com/reference/build-checks/json-args-recommended/#workarounds), variable expansion and command chaining used in the existing `CMD` aren't available in the exec form , so instead we can explicitly declare the shell wrapper to silence the annotation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,4 +77,4 @@ ENV PORT=8080
 EXPOSE ${PORT}
 
 SHELL ["/bin/sh", "-c"]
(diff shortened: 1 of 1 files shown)


jebw (approved):
LGTM

PR #2670, opened 2025-06-30
Configure containers to run as non-root user

shaheislamdfe:
### Context

Ticket: https://trello.com/c/ylAGk14G/2375-security-aks-containers-run-as-non-root-user-npq

Run the container as non root as a security mitigation


jebw (changes requested):
Looks good and will be a nice security upgrade but I think its probably marking more areas as writable than necessary so we can lock that down a bit tighter

PR #2678, opened 2025-07-01
[CPDNPQ-2665] test DfE Identity sign in using captured auth hashes

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2665

To ensure we can test our DfE Identity auth flow works, based on what we actually get back from DfE Identity.

2 specs to generate fixtures using real DfE Identity responses.
2 specs to test our sign in using these fixtures, for a user with a TRN and without a TRN.


spec/features/journeys/happy_paths/dfe_sign_in_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+# 4. update `email_for_account_with_trn` to the email address for the account with a TRN
+# 5. update `email_for_account_without_trn` to the email address for the account without a TRN
+# 6. make sure your local server is not running on port 3000
+# 7. run the first 2 specs - be sure to follow the prompts from the command-line, and do not enter your code into the browser
+RSpec.feature "DfE Sign In", :with_default_schedules, type: :feature do
+  include Helpers::JourneyAssertionHelper
+  include Helpers::JourneyStepHelper
+
+  before do
+    WebMock.disable_net_connect!(allow_localhost: true, allow: URI(ENV["TRA_OIDC_DOMAIN"]).host)
```
jebw, line 19: This is only needed when regenerating fixtures, and needs to be 'put back' after the spec runs to avoid mutating the value for the rest of the test suite

Does it work if you use mocking of this somehow, or failing that using an `after` block

PR #2681, opened 2025-07-02
fix flaky spec - specify cohort, so that factory does not attempt to update current cohort

alkesh:
### Context

occasional failure of spec: `spec/services/applications/query_spec.rb:362`

specify previous cohort, otherwise the cohort factory will attempt to update the existing current cohort, which causes the validation failure.


jebw (approved):
LGTM :raised_hands:

PR #2717, opened 2025-07-15
[CPDNPQ-3009] Admin ability to create declarations on behalf of Lead Providers

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3009

Based on the other bulk operations, this new one lets admins create declarations on behalf of providers who no longer have API access. Uses the same `Declarations::Create` service as the API to actually create the declarations.

n.b. the design work appears to have based the input CSV columns on the API parameters, so I took the liberty of adding a `lead_provider_name` column, since in the API it's implicit based on the API token, but obviously no API token is used here.

### DIFF
--- a/app/views/npq_separation/admin/bulk_operations/submit_declarations/index.html.erb
+++ b/app/views/npq_separation/admin/bulk_operations/submit_declarations/index.html.erb
@@ -28,7 +28,7 @@
     For example:
     <pre>
       participant_id,declaration_type,declaration_date,course_identifier,delivery_partner_id,lead_provider_name
-      1234567890,declaration_type,2021-01-01,course_identifier,1234567890,"Lead provider name"
+      1234567890,declaration_type,2021-01-01T00:00:00Z,course_identifier,1234567890,"Lead provider name"
     </pre>
   </p>
 


app/models/bulk_operation/submit_declarations.rb
```diff
@@ -0,0 +1,60 @@
...
+  HEADERS = true
+  FILE_HEADERS = %w[participant_id declaration_type declaration_date course_identifier delivery_partner_id lead_provider_name].freeze
+
+  def run!
+    result = {}
+    ActiveRecord::Base.transaction do
+      result = csv_rows.each_with_index.to_h do |row, index|
+        row_number = index + 1
+        [row_number, process_csv_row(row)]
+      end
```
jebw, line 11: I may be missing something but `each_with_index.to_h` looks odd ?

app/models/bulk_operation/submit_declarations.rb
```diff
@@ -0,0 +1,60 @@
...
+      delivery_partner_id: row["delivery_partner_id"],
+    )
+
+    if service.create_declaration
+      "Declaration created successfully"
+    else
+      service.errors.full_messages.join(", ")
+    end
+  rescue StandardError => e
+    "Error: #{e.message}"
```
jebw, line 58: I think if the users getting something more than a validation error we probably want to just 500 and let us get notified via sentry

Otherwise if they don't tell us we'll never know about it and in general I'm pretty cautious about exposing unknown exception messages to the user in case there's something leaking in the exception messages which shouldn't be exposed.

jebw (changes requested):
Looks good - minor comment around the error handling and query around the `each_with_index.to_h` stuff though from the specs that implies this works just fine and its something I'm misreading

PR #2722, opened 2025-07-16
Configure containers to run as non-root user

shaheislamdfe:
### Context

[What has been changed?]

failures:


Dockerfile
```diff
@@ -64,10 +64,23 @@ RUN apk add --update --no-cache libpq tzdata yaml ${EXTRA_PACKAGES} && \
...
 
+# Change ownership of the app directory to the non-root user
+RUN chown -R appuser:appgroup /app
+
+# Make everything read-only (with execute for directories)
+RUN chmod -R a-w /app && \
+    find /app -type d -exec chmod +x {} \;
+
+# Grant write access only to directories that need it
+RUN chmod -R u+w /app/tmp /app/public/api/docs /app/log
```
RMcVelia, line 82: As we run a chown on everything below /app, the chmod commands aren't going to give us any security improvement i.e. if you connect to the pod you can change it. The ask here is can we restrict the directories we need to chown instead of recursively chown /app e.g. just chown /app/log, /app/tmp, ...

Dockerfile
```diff
@@ -64,10 +64,23 @@ RUN apk add --update --no-cache libpq tzdata yaml ${EXTRA_PACKAGES} && \
...
+
 # Copy files generated in the builder image
 COPY --from=builder /app /app
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 
+# Change ownership only for directories that need write access
+RUN chown -R appuser:appgroup /app/tmp /app/public/api/docs /app/log
+
+# Make everything read-only (with execute for directories)
+RUN chmod -R a-w /app && \
```
RMcVelia, line 78: We don't need to run any of the chmods. The default umask means the appuser has the right permissions. Plus we want to keep this as simple as possible.

PR #2724, opened 2025-07-16
Add new row for schedule identifier to the application show page

EmmaFrith:
Add schedule identifier row to the funding eligibility section on the application show page.

Admins need this info when investigating queries and bugs.

Screenshot 2025-07-16 at 13 55 14

ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3071

### DIFF
--- a/spec/features/npq_separation/admin/applications_spec.rb
+++ b/spec/features/npq_separation/admin/applications_spec.rb
@@ -179,7 +179,7 @@ RSpec.feature "Listing and viewing applications", type: :feature do
       expect(summary_list).to have_summary_item("Funded place", "")
       expect(summary_list).to have_summary_item("Status code", application.funding_eligiblity_status_code.humanize)
       expect(summary_list).to have_summary_item("Schedule cohort", application.cohort.start_year)
-      expect(summary_list).to have_summary_item("Schedule identifier", application.schedule_id)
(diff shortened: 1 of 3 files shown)


spec/features/npq_separation/admin/applications_spec.rb
```diff
@@ -179,6 +179,7 @@
       expect(summary_list).to have_summary_item("Funded place", "")
       expect(summary_list).to have_summary_item("Status code", application.funding_eligiblity_status_code.humanize)
       expect(summary_list).to have_summary_item("Schedule cohort", application.cohort.start_year)
+      expect(summary_list).to have_summary_item("Schedule identifier", application.schedule&.identifier || "-")
```
alkesh, line 182: I think this should be 2 different specs - one for when the schedule is nil, and one for when it isn't

jebw: @alkesh I think if we're testing permutations of views at the feature level we'll end up the road to a very slow test suite

I think this is where view specs really benefit - they are just 'view' unit tests and so very fast and allow checking many permutations easily

@EmmaFrith The actual spec should be explicitly testing for something, ie 'not this or that', it should be 'check this=something' if that makes sense?

alkesh (approved):

jebw (approved):
LGTM :raised_hands:

PR #2748, opened 2025-07-25
[CPDNPQ-3099] change application#inside_catchment? to check school for 2023 cohort and earlier

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3099

- check the school URN for applications in the 2023 or earlier cohort
- make the seeds more like production - `teacher_catchment` will be `nil` for the 2023 or earlier cohorts

### DIFF
--- a/spec/factories/schools.rb
+++ b/spec/factories/schools.rb
@@ -1,5 +1,5 @@
 FactoryBot.define do
-  sequence(:urn) { |n| sprintf("1%05d", n % 1_000_000) }
+  sequence(:urn) { |n| sprintf("1%05d", n % 100_000) }
   sequence(:ukprn) { |n| sprintf("%08d", n % 100_000_000) }
 
   factory :school do


spec/factories/schools.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
-  sequence(:urn) { |n| sprintf("%06d", n % 1_000_000) }
+  sequence(:urn) { |n| sprintf("1%05d", n % 1_000_000) }
```
jebw, line 2: now needs to loop at 100k not 1 million

app/models/application.rb
```diff
@@ -137,7 +137,7 @@ def inside_uk_catchment?
   end
 
   def inside_catchment?
-    %w[england].include?(teacher_catchment)
+    %w[england].include?(teacher_catchment) || (cohort.start_year < 2024 && !!school&.urn&.starts_with?("1"))
```
jebw, line 140: I think considering this is a 'fix' for declarations against the legacy applications I'd rather this changed the check in Declaration rather than changing the definition of inside_catchment? on Application

jebw (commented):
I think this change is better on the Declaration?

jebw (approved):
LGTM

PR #2755, opened 2025-07-30
[CPDNPQ-2944] load latest private childcare providers

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2944

* import latest files
* update headers to match files
* cope with files being ISO-8859-1 encoding
* cope with unexpected unicode dash character
* cope with non-breaking spaces in postcode


app/services/importers/import_private_childcare_providers.rb
```diff
@@ -99,6 +101,14 @@ def persist_csv_row(csv_row, row_number)
       @import_errors[row_number_for_errors] << e.message
     end
 
+    def translate_unicode_characters(string)
+      string.tr("\u0096", "\u2013")
+    end
+
+    def strip_whitespace(string)
+      string.gsub(/\A\p{Space}+|\p{Space}+\z/, "")
```
jebw, line 109: is this doing the same as `#strip`?

PR #2763, opened 2025-08-01
[CPDNPQ-3034] bugfix - cope with versions with no object_changes

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3034

application history doesn't work for some applications

cope with application.versions that don't have any object_changes


jebw (approved):
LGTM :raised_hands:

PR #2767, opened 2025-08-04
Prevent line break between minus sign and pound symbol in currency formatting

EmmaFrith:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3097

We do not want minus numbers to wrap over two lines. Resolving this by using a non-breaking hyphen.

# Before

image

# After

Screenshot 2025-08-04 at 13 58 28


spec/components/npq_separation/admin/adjustments_table_component_spec.rb
```diff
@@ -39,14 +39,14 @@
...
-      it { is_expected.not_to have_css "tbody td", text: "£600" }
+      it { is_expected.not_to have_css "tbody td", text: "£0.00" }
     end
 
     context "when show_total is true" do
       let(:show_total) { true }
 
       it { is_expected.to have_css "tbody th", text: t(".total") }
-      it { is_expected.to have_css "tbody th", text: "£600" }
+      it { is_expected.to have_css "tbody th", text: "£0.00" }
```
alkesh, line 49: can we have the total adding up to something more than zero?
just to test the total is actually working, and it's not just doing `nil.to_i`

PR #2773, opened 2025-08-05
Ensure that training status cannot be set when provider approval status is pending

EmmaFrith:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3119

When the provider approvial status is in pending, i.e. it has not yet been set:
- do not show change link next to 'Training status' - people should not be able to change the training status in this scenario
- add a dash next to 'Training status' - we use a dash throughout the admin console when a status is not yet set

| Before    | After |
| -------- | ------- |
| Screenshot 2025-08-05 at 17 05 58  | Screenshot 2025-08-05 at 17 04 06    |

Account for the scenario where you could still reach this page with a better error message and consistent use of the dash in the inset text:

| Before    | After |
| -------- | ------- |
| Screenshot 2025-08-05 at 17 08 23 | Screenshot 2025-08-05 at 17 16 12     |

### DIFF
--- a/app/services/applications/change_training_status.rb
+++ b/app/services/applications/change_training_status.rb
@@ -93,6 +93,8 @@ module Applications
     end
 
     def do_not_change_status_of_pending_application
+      return if errors.any?
+
       if application.lead_provider_approval_status == "pending"
         errors.add(:training_status, :pending_lead_provider_approval_status)
       end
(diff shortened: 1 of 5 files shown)


jebw (approved):
LGTM :raised_hands:

PR #2778, opened 2025-08-06
[CPDNPQ-3133] TAD request for SENCO

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3133

a new exporter: `TadSencoDataRequest` because this request is different to the previous requests.

### DIFF
--- a/app/services/exporters/tad_senco_data_request.rb
+++ b/app/services/exporters/tad_senco_data_request.rb
@@ -3,13 +3,10 @@ class Exporters::TadSencoDataRequest
     @cohort = cohort
     @schedules = schedules
     @file = file
-    @csv = []
-    @csv << csv_headers
   end
 
   def call
-    generate_data
-    save_data_to_csv
+    generate_csv
   end
 
   def applications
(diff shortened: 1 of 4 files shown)


app/services/exporters/tad_senco_data_request.rb
```diff
@@ -0,0 +1,95 @@
...
+        application.lead_provider_approval_status,
+        application.training_status,
+        application.headteacher_status,
+        application.cohort.start_year,
+      ]
+    end
+  end
+
+  def started?(application)
+    application.declarations.where(declaration_type: "started").any?
```
jebw, line 58: AFAIK this will not use the eager-loaded declarations and instead will do its own query - would be better iterating the loaded declarations

```ruby
!!applications.declarations.find { |d| d.declaration_type == "started" }
```

jebw (changes requested):
Just spotted the repeated database query - I think that does need changing and is less 'syntactic preference'

I've now tested this and it does generate a query each time `#started?` is called

jebw (approved):
LGTM :raised_hands:

PR #2781, opened 2025-08-08
Remove unnecessary side navs from pages

EmmaFrith:
We have side navs on some pages where we don't need them. We also have some pages where there are 2 separate navs on the same page - making the page cluttered and confusing. 

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3125 
(This PR removes the side nav in more places than mentioned in the ticket - but this has been agreed with the designers.)

### Remove the 'Statements' and 'Cohorts' sidenav from the 'Finance' page

The top nav can already be used to access statements and cohorts. What's more, when you click through to view a statement, you end up having 2 side navs - which isn't conventional. The filter also doesn't need to be on the show page as there's nothing to filter. It's already on the index page.

| Before  | After |
| Screenshot 2025-08-08 at 17 08 34      | Screenshot 2025-08-08 at 17 09 16     |
| Screenshot 2025-08-08 at 17 02 51      | Screenshot 2025-08-11 at 12 43 27     |

### Remove the providers side nav

1. Remove the providers list side nav. Having a side nav here means when you click through to a provider, you end up with 2 sidenavs - which isn't conventional. It's also unnecessary because there's already an item on the top nav for 'Providers'.

2. Use a caption for the provider name rather than a H1, so that the H1 be be reserved for the cohort name and therefore match the nav items. This also avoids having heading levels stacked on top of each other with no content in between.

| Before  | After |
| Screenshot 2025-08-08 at 17 10 37      | Screenshot 2025-08-08 at 17 11 59     |
| Screenshot 2025-08-08 at 17 12 38      | Screenshot 2025-08-18 at 11 59 19     |

### Remove the 'All applications' and 'In review' sidenav from the 'Users' index page

'All applications' and 'In review' is related to the 'Applications' page, not the 'Users' page.

| Before  | After |
| Screenshot 2025-08-08 at 17 32 41    | Screenshot 2025-08-08 at 17 30 56    |

### Edits to the user show page

1. Remove the user name and 'Overview' form the side nav. 
This doesn't serve a purpose as it only ever says overview here. 

2. Put the user info into a summary card
This is more accessible and in line with the gov design system.

3. Remove 'All applications' and 'In review' from the side nav.
These are related to applications not users.

| Before  | After |
| Screenshot 2025-08-12 at 16 54 39    | Screenshot 2025-08-12 at 16 52 14    |

### Edits to the applications/applications in review show pages

1. Remove the 'All applications' and 'In review' sidenav 
Having 2 separate nav bars on a page is confusing.

2. Make content more consistent and descriptive 
Make the H1s and related nav items match. Make it clearer that you're on an application page rather than a user page.

5. Remove the user info from application show page
Add a link to the user's page instead to avoid duplication. 

| Before  | After |
| Screenshot 2025-08-11 at 17 34 43     |  Screenshot 2025-08-18 at 10 49 03  |

### Remove user links from the applications index page

It wasn't clear which link would take you to the application show page.

| Before  | After |
| Screenshot 2025-08-12 at 17 09 19    |  Screenshot 2025-08-12 at 17 11 04  |


siddiqah-ii:
All of these look good! Just a few things I noticed now:

For the provider details page [see screenshot] we've removed the name of the provider on the side as the header already covers this. I've made sure to update the prototype to reflect this:https://npq-prototype.herokuapp.com/admin-v3/ambition

Pages that need a back button
finance statement page
providers details page
user details page
overview page

image

spec/components/npq_separation/navigation_structures/admin_navigation_structure_spec.rb
```diff
@@ -50,35 +50,4 @@
...
-
-    describe "Finance" do
-      subject { instance.sub_structure("Finance") }
-
-      it "the first entry is Statements" do
-        expect(subject.first.name).to eql("Statements")
-        expect(subject.first.href).to eql("/npq-separation/admin/finance/statements")
-      end
-    end
-  end
```
jebw, line 83: This functionality still exists but isn't testable with the side navdata removed. I'd prefer we didn't have untested functionality so I'd suggest adding a new spec file which creates a test structure to test the different behaviours, eg

`spec/components/npq_separation/navigation_structures/navigation_structure_spec.rb`

```ruby
require "rails_helper"

RSpec.describe NpqSeparation::NavigationStructure do
  subject(:navigation) { test_structure.new }

  let :test_structure do
    Class.new(described_class) do
    private
      def structure
        {
          Node.new(
            name: "First",
            href: "/first",
            prefix: "/first",
          ) => [
            Node.new(
              name: "Subnav",
              href: "/subnav",
              prefix: "/subnav",
            )
          ],
          Node.new(
            name: "Second",
            href: "/first",
            prefix: "/first",
          ) => [],
        }
      end
    end
  end

  describe "#primary_structure" do
    # specs for primary structure go here
  end

  describe "#sub_structure" do
    # specs for sub structure go here
  end
end
```

jebw (commented):
Looks good, but I'd like to see the substructure code getting replacement specs added

I've a secondary concern with the removal of the side 'form' on the statements - we've just added that so if its a useful to the contract team then removing it again shortly after could cause frustration. I think that particular change needs checking with the users?

app/views/npq_separation/admin/applications/reviews/_application_side_navigation.html.erb
```diff
@@ -1,16 +1,14 @@
 <% content_for :side_navigation do %>
-  <hr class="govuk-section-break govuk-section-break--m">
-
   <%=
     render NpqSeparation::SubNavigationComponent.new(
       request.path,
       heading: {
-        text: @application.user.full_name,
+        text: "#{@application.user.full_name}, #{@application.course.name}, #{@application.created_at.to_date.to_fs(:govuk)}",
```
jebw, line 6: This looks very long for something being shown in a narrow side nav?

jebw: Looking at the version it is replacing - by concatentating all the information together on a single row which will wrap in an inconsistent fashion I think we make this harder to read - not easier

Since the name is just duplicating the h1 title to the side of it - can it be something like the Date followed by a line break, followed by the Course name on the line below (possibly in a slightly smaller font to reduce the likelyhood of text wrapping)

app/views/npq_separation/admin/applications/reviews/index.html.erb
```diff
@@ -125,7 +146,7 @@
           row.with_cell(text: boolean_red_green_tag(application.eligible_for_funding))
           row.with_cell(text: lead_provider_approval_status_badge(application.lead_provider_approval_status))
           row.with_cell(text: application.notes)
-          row.with_cell(text: govuk_link_to("View", npq_separation_admin_application_review_path(application.id)))
+          row.with_cell(text: govuk_link_to("View application", npq_separation_admin_application_review_path(application.id)))
```
jebw, line 149: Whilst this does make this clearer, it is providing context the user already has. Given a list of applications, with a link to view - a default expectation would be to view the application that show is showing.

This has the disadvantage that it also takes up significantly more horizontal space - the screen grabs are all using 'short demon' names for both user and school - I think with longer names, which they are likely to be in reality, then this will result in `View application` potentially wrapping.

Whilst

app/views/npq_separation/admin/applications/index.html.erb
```diff
@@ -1,3 +1,24 @@
...
+            name: "All applications",
+            href: npq_separation_admin_applications_path,
+            prefix: /\/npq-separation\/admin\/applications(?!\/reviews)$/,
+            ),
+          NpqSeparation::NavigationStructure::Node.new(
+            name: "In review",
+            href: npq_separation_admin_application_reviews_path,
+            prefix: /\/npq-separation\/admin\/applications\/reviews$/,
+            ),
+        ]
```
jebw, line 17: Can this be moved to a partial and be reused for the screen with the same code below, thanks

jebw (commented):
Code changes look good - the only technical comment i have is can the duplicate sidenav code in applications and applications in review be shared (using a partial)

There are a couple of UX queries I have though

1) The 'View application' is arguably clearer but is providing context the user already has and comes at a cost of likely resulting in the words _View_ and _application_ ending up on different lines which will be _more_ confusing

2) The Sidenav title when viewing an individual application is 3 lots of information concatentating onto a single sentence which is so long it then wraps over three lines

This feels more complex then what it is replacing. The Individuals name seems redundant, which trims it slightly but splitting the other via a line break or using different formatting would improve readability?

app/views/npq_separation/admin/applications/reviews/index.html.erb
EmmaFrith: Yea I don't have a strong opinion on this - happy to remove 'application'.

PR #2790, opened 2025-08-14
[CPDNPQ-3143] Exclude withdrawn applications when allowing a lead provider to accept

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3143

when looking for other accepted applications with same course and cohort - don't include withdrawn applications.

### DIFF
--- a/app/services/applications/accept.rb
+++ b/app/services/applications/accept.rb
@@ -77,14 +77,12 @@ module Applications
     def other_accepted_applications_with_same_course_and_cohort
       return if application.blank?
 
-      not_withdrawn_training_statuses = Application.training_statuses.except(:withdrawn).keys.append(nil)
-
       @other_accepted_applications_with_same_course_and_cohort ||= Application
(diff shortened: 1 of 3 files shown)


app/services/applications/accept.rb
```diff
@@ -77,11 +77,14 @@ def reject_other_applications_in_same_cohort!
     def other_accepted_applications_with_same_course_and_cohort
       return if application.blank?
 
+      not_withdrawn_training_statuses = Application.training_statuses.except(:withdrawn).keys.append(nil)
```
jebw, line 80: Is there a reason not to use

```ruby
.where.not(training_status: nil)
```

alternatively can this be moved to a scope on the `Application` model so its available for use in other scenarios

jebw (commented):
Looks good - I'm wondering whether the exclusion of training status can be simpler?

jebw (approved):
LGTM

PR #2796, opened 2025-08-15
fix flaky spec

alkesh:
### Context

occasional failure in funding_eligibility_spec:

`rspec ./spec/services/funding_eligibility_spec.rb[1:15:38:2:5:1:2]` # FundingEligibility when institution is a School when school is LA nursery when not LA disadvantaged nursery when user has selected the early_years_leadership course behaves like funding eligibility returns funding_eligiblity_status_code not_entitled_ey_institution

The URN was previously just a number, and occasionally it would match a URN from LA_DISADVANTAGED_NURSERIES.
This PR changes the factory so URNs and UKPRNs start with 'TEST', so they will never match a URN from the PP50 lists.


jebw (approved):
LGTM

PR #2811, opened 2025-08-20
[CPDNPQ-3025] enable GAI syncing - update message format and fix admin console

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3025

* the message format of the webhook message has changed
* we did not used to use the preferred name (we do when the user first logs in) - this has been fixed
* the admin console webhook messages pages had the wrong path names
* make GAI syncing and login code use same logic when updating TRN/trn_verified/trn_lookup_status - this logic is now in the `User` model
* add missing specs for webhook messages pages on admin console


jebw (approved):
LGTM

PR #2833, opened 2025-08-28
[CPDNPQ-3028] Follow-up 2: fix childminder eligibility

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3028

Following the last testing session, the childminder eligibility list was removed since the only qualification for eligibility in the upcoming cohort is Ofsted registration.

However, while the `private_childcare_providers` table had been updated from the latest CSV, E2E testing revealed a URN that's not in that latest CSV incorrectly returning a positive funding result, so it seems that the update task does not delete/disable old records and therefore we cannot use this table to check eligibility in the current cohort.

It's not immediately obvious that we can safely delete/disable old `PrivateChildcareProvider` records, so instead, this PR simply restores the eligibility list using the URN column of the latest CSV. I've checked against the E2E testing criteria and it appears to now return the desired results (although this will be confirmed by another round of BA testing.)


jebw (approved):
LGTM

PR #2844, opened 2025-09-02
[CPDNPQ-3192] Fix display of timestamps in admin

jebw:
### Context

Ticket: [CPDNPQ-3192](https://dfedigital.atlassian.net/browse/CPDNPQ-3192)

We are currently showing the timestamps an hour out in the admin console. This is because we are configured to show timestamps in UTC, not BST

1. Swap to using a UK local timezone which should then display correctly in both summer and winter

failures:


app/controllers/application_controller.rb
```diff
@@ -70,4 +71,9 @@ def clear_null_user_sessions
       redirect_to root_path
     end
   end
+
+  def set_time_zone(&block)
```
alkesh, line 75: worth adding a comment here why we're not using `config.time_zone`.

PR #2881, opened 2025-09-17
[CPDNPQ-2997] Application#reason_for_rejection

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2997

There are several reasons that an application can end up in the `rejected` state, which can make it hard for Lead Providers and administrators to understand. Recording and exposing this information should make the service more efficient.

n.b. after chatting with Slawosz and Alkesh, it seems that we can't add swagger documentation that depends on feature flags, so this is a conscious omission from the Jira acceptance criteria.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -308,6 +308,10 @@ en:
           attributes:
             add_another:
               inclusion: Select if you need to add another adjustment
+        admin/delivery_partners/continue_form:
+          attributes:
+            continue:
+              inclusion: Select ‘Yes’ or ‘No’ to continue
         applications/change_training_status:
           attributes:
             training_status:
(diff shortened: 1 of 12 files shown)


app/services/applications/accept.rb
```diff
@@ -71,7 +71,10 @@ def accept_application!
     def reject_other_applications_in_same_cohort!
       return if other_applications_in_same_cohort.blank?
 
-      other_applications_in_same_cohort.update!(lead_provider_approval_status: "rejected")
+      other_applications_in_same_cohort.update!(
+        lead_provider_approval_status: "rejected",
+        reason_for_rejection: Application.reason_for_rejections[:application_accepted_by_other_provider],
```
alkesh, line 76: is this the correct reason here?
the `other_applications_in_same_cohort` could be for the same provider.

app/services/applications/reject.rb
```diff
@@ -4,15 +4,17 @@ class Reject
     include ActiveModel::Attributes
 
     attribute :application
+    attribute :reason_for_rejection
 
     validates :application, presence: true
+    validates :reason_for_rejection, presence: true
```
alkesh, line 10: should this presence validation be only if the feature flag is enabled?

config/locales/en.yml
```diff
@@ -633,6 +633,8 @@ en:
           attributes:
             application:
               <<: *application
+            reason_for_rejection:
+              <<: *application
```
alkesh, line 637: is this correct? seems odd for `reason_for_rejection` to have all the application attributes.

PR #2886, opened 2025-09-18
fix visualisation

alkesh:
### Context

Generations the Registration Wizard Flow is failing on deployment

* remove spec that tests if PNG is created (it wasn't actually testing this anyway, it was testing if the .dot file was created)
* fix regex that looks for next_steps - so it matches `:stepname`, but not (for example) `Application.employment_types[:lead_mentor_for_accredited_itt_provider`
* don't try and test using graphviz PNG generation in rspec, but change the github workflow `build_flow_visualisation.yml` to run on pull requests, but not actually commit the diagram

example of what happens when the visualisation image generation fails: https://github.com/DFE-Digital/npq-registration/actions/runs/17833408241/job/50704183815?pr=2886

### DIFF
--- a/app/services/registration_wizard_visualiser.rb
+++ b/app/services/registration_wizard_visualiser.rb
@@ -115,8 +115,11 @@ private
 
     Rails.logger.debug("Generating #{output_graph_filename}")
     generate_graph_command = "dot -Tpng #{output_digraph_filename} -o #{output_graph_filename}"
+
     Rails.logger.debug(generate_graph_command)
     system(generate_graph_command)
+
+    raise "Graph generation failed" unless $CHILD_STATUS.exitstatus.zero?
   end
 
   def save_file(name, content)


.github/workflows/build_flow_visualisation.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Build visualisation
         run: |
-          dot -v
+          dot -V
```
alkesh, line 32: I'm not sure what this was trying to achieve with `dot -v` - this just hangs (-v is for verbose, -V is for version).
`dot -V` just prints out the version, so useful to see if graphviz is installed.

.github/workflows/rspec.yml
```diff
@@ -91,6 +91,9 @@ jobs:
         ci_node_index: [0, 1, 2, 3, 4, 5]
 
     steps:
+      - name: Install Graphviz
+        run: sudo apt install graphviz
```
alkesh, line 95: adds 1m38s to the build time

PR #2889, opened 2025-09-18
Add optional application_id attribute to V3 api declaration create request

slawosz:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3144

In some cases there could be 2 or more applications for given course. We would like to let LP choose an application in such case.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -601,6 +601,8 @@ en:
               missing_contract_for_cohort_and_course: You cannot submit a declaration for this participant as you do not have a contract for the cohort and course. Contact the DfE for assistance.
             application:
               application_schedule_missing: The application is missing a schedule.
+            application_id:
+              application_not_found: "Your update cannot be made as an accepted application cannot be found for the given '#/application_id'."
             delivery_partner_id:
               not_found: The property '#/delivery_partner_id' does not exist
             secondary_delivery_partner_id:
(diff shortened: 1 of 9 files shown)


config/locales/en.yml
```diff
@@ -601,6 +601,8 @@ en:
               missing_contract_for_cohort_and_course: You cannot submit a declaration for this participant as you do not have a contract for the cohort and course. Contact the DfE for assistance.
             application:
               application_schedule_missing: The application is missing a schedule.
+            application_id:
+              application_not_found: "Your update cannot be made as an accepted application cannot be found for the given '#/application_id'."
```
jebw, line 605: This error can be briefer I think - "An accepted application cannot be found for the given '#/application_id'." ?

public/api/docs/v3/swagger.yaml
```diff
@@ -214,7 +214,7 @@ paths:
                         teacher_catchment_iso_country_code: GBR
                         itt_provider: University of Southampton
                         lead_mentor: true
-                        funded_place:
+                        funded_place:
```
jebw, line 217: Odd how this is adding the whitespace on the end of lines - I think these whitespace only changes can be skipped

jebw (commented):
LGTM - some minor syntactic nits and I think we could do with an extra spec - looks great :+1:

jebw (approved):
LGTM

PR #2896, opened 2025-09-22
[CPDNPQ-2998] Lead Provider to have visibility of declarations if applicants have transferred

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2998

- When applicants transfer Lead Providers, the new Lead Provider should be able to see existing declarations made by the previous Lead Provider
- The previous Lead Provider should continue to only see their own declarations
- No change to write permissions

Changing the query seems the most expedient route for this, since it's used in the declaration endpoints, and doesn't affect any write permissions. With this change, we grant visibility to the current `Application` owner, as well as the `Declaration` owner.

### DIFF
--- a/app/services/feature.rb
+++ b/app/services/feature.rb
@@ -17,6 +17,7 @@ class Feature
     MAINTENANCE_BANNER,
     DFE_ANALYTICS_ENABLED,
     LP_SELF_SERVE,
+    LP_TRANSFERRED_DECLARATIONS_VISIBILITY,
   ].freeze
 
   class << self


slawosz (approved):
Minor comment, LGTM

jebw (changes requested):
I've realised this would also allow the LP to update another LPs declarations rather then just see them.

That is not correct so we'll need to special case how this is handled in Declarations::Query for the GET endpoints

jebw (commented):
My concern with this approach is potential for a future developer misunderstanding the `Declaration::Query` behaviour - it was too easy to not see the broader impact of `Declarations::Query` pulling in declarations from other LPs - I think default behaviour should be the original behaviour since that is more inline with what would be expected.

I think this more expansive behviour needs to be an 'opt in' - so an additional named param on `Declarations::Query` telling it to pull in the extra declarations (this can then just get fed the feature flag result).

jebw (approved):
LGTM 🙌

PR #2899, opened 2025-09-23
[CPDNPQ-2910] raise RoutingError when no template found

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-2910

- catch the `ActionView::MissingTemplate` for the `API::GuidanceController#show` and raise a `ActionController::RoutingError`, which results in a 404 response.
- make sure error pages for `/api/guidance/` and `/api/docs/` are in HTML not JSON.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -30,6 +30,6 @@ class ErrorsController < PublicPagesController
 private
 
   def set_format
-    request.format = :json if request.original_fullpath =~ /^\/api\/v\d+/ && request.format.html?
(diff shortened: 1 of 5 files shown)


app/controllers/errors_controller.rb
```diff
@@ -30,6 +30,6 @@ def unprocessable_entity
 private
 
   def set_format
-    request.format = :json if request.original_fullpath.start_with?("/api/") && request.format.html?
+    request.format = :json if request.original_fullpath =~ /^\/api\/v\d+/ && request.format.html?
```
jebw, line 33: This would negatively impact APIs not with a version number (ie qualifications api)

How about just overriding this method to be empty in the guidance controller?

PR #2919, opened 2025-09-30
[CPDNPQ-3058] Admin console - update eligibility lists

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3058

- page with tabs to allow new eligibility list files to be uploaded
- one new table (`eligibility_lists`) to store all the entries, with a `type` using Rails STI
- when checking eligibility, the `School` and `PrivateChildcareProvider` models will use the the table if any entries have been uploaded, otherwise the current in-memory hashes will be used
- a future commit can remove the current in-memory hashes, once all the lists have been uploaded in production via this feature.

### DIFF
--- a/app/controllers/npq_separation/admin/eligibility_lists_controller.rb
+++ b/app/controllers/npq_separation/admin/eligibility_lists_controller.rb
@@ -6,8 +6,12 @@ class NpqSeparation::Admin::EligibilityListsController < NpqSeparation::AdminCon
   def create
     entries_loaded = @service.call
 
-    flash[:success] = "Eligibility list updated - #{entries_loaded} entries loaded" unless @service.errors.any?
-    redirect_to npq_separation_admin_eligibility_lists_path
+    if @service.errors.any?
+      render :show
+    else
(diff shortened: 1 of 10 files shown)


jebw (commented):
Really nice! I think you're going to curse me though 😉 

So thinking this through - it would be nice to add `has_many` relationships to School and PrivateChildcareProvider for accessing the related records

```ruby
class School
  has_many :list_entries, 
            -> { where(identifier_type: :urn) },
            class_name: "EligibilityList",
            foreign_key: :identifier,
            primary_key: :urn
end
```

That would be eager loadable for example if needed in the future?

I'd not understood you were planning to use STI when we'd talked this through - considering you are using STI, this leads to a second thought - do we need the `identifier_type` column? The same behaviour could be achieved with `type` hierarchy ?

The downside of this is there's nothing explicit in the database describing what the identifier refers to?

```
-> EligibilityList::Entry
    -> EligibilityList::Urn
        -> EligibilityList::Pp50School
        -> EligibilityList::RiseSchool
    -> EligibilityList::Ukprn
        -> EligibilityList::Childminder
```

thoughts?

PR #2921, opened 2025-09-30
[CPDNPQ-3149] allow adjustments to have decimal places

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3149

* change datatype in database
* change datatype in form attributes
* update textfield to have no suffix (was previously `.00`)


db/migrate/20250930132614_change_adjustments_amount_precision.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeAdjustmentsAmountPrecision < ActiveRecord::Migration[7.2]
+  def up
+    safety_assured do # this will lock the table whilst updating the column
```
alkesh, line 3: should be fine locking the adjustments table - we have relatively few, and it only affects admin pages.

PR #2922, opened 2025-09-30
[CPDNPQ-3250] Record preferred name

rwrrll:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3250

>DfE Identity provides us with 2 names for users - their full name and their preferred name
>
>Our DfE Id integration is currently choosing the ‘preferred name’ if available to record as the users name. Some users are putting only their first name as the preferred name though - meaning we don’t record their full name.
>
>We should changes the Users table to have columns for both and the DfE Identity integration to record both fields.

n.b. Additional implementation notes re: acceptance critieria in Jira comments.

WRT analytics: I've kept this in the blocklist for this PR to avoid merge issues with in-progress CPDNPQ-3197. Depending on review/merge timing, the new `preferred_name` field can be included in that PR, or a quick follow-up.

> [!TIP]
> It may be easier to review the two commits separately, the latter deals with fixing unrelated tests broken by the former

### DIFF
--- a/app/services/applications/search.rb
+++ b/app/services/applications/search.rb
@@ -18,6 +18,7 @@ module Applications
       scope.where(ecf_id: query)
         .or(scope.where(declarations: { ecf_id: query }))
         .or(scope.where("users.full_name ilike ?", "%#{query}%"))
+        .or(scope.where("users.preferred_name ilike ?", "%#{query}%"))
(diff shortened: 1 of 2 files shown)


jebw (approved):
LGTM 🙌

PR #2942, opened 2025-10-06
Remove `iconv` as it's no longer necessary

peteryates:
Basic encoding and decoding [is now supported directly by Ruby](https://docs.ruby-lang.org/en/master/encodings_rdoc.html). We're only using Iconv in this one place for dealing with GIAS downloads, so it makes sense to remove the unnecessary dependency.

Testing manually:

```ruby
[1] pry(main)> data = CSV.parse(File.read('edubasealldata20251003-sample.csv'), headers: true)
# CSV::InvalidEncodingError: Invalid byte sequence in UTF-8 in line 4. (CSV::InvalidEncodingError)
# from /home/peter/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/csv-3.3.5/lib/csv/parser.rb:894:in 'CSV::Parser#build_scanner'


[2] pry(main)> data = CSV.parse(File.read('edubasealldata20251003-sample.csv').encode(Encoding::UTF_8, Encoding::ISO8859_1), headers: true)
# => #<CSV::Table mode:col_or_row row_count:500>
```


jebw (approved):
LGTM 🙌 

Good spot, I'd not realised we were using Iconv - kinda amazed we ever were - pretty much all reasons to use it went away with ruby 2.0 from years ago

PR #2948, opened 2025-10-07
[CPDNPQ-3058] admin upload eligibility lists only

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-3058

Aplitting this ticket into two:

1. allow eligibility lists to be uploaded into the database, but do not use the new lists - old ones will be used.
2. switch to using the new database eligibility lists, and the code referring to the old lists will be removed.

This PR is the first step only.

- page with tabs to allow new eligibility list files to be uploaded
- one new table (eligibility_list_entries) to store all the entries, with a type using Rails STI


spec/services/eligibility_lists/update_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+    end
+
+    context "when the file has no data rows" do
+      let(:file) { tempfile_with_bom("#{EligibilityList::Pp50School::IDENTIFIER_HEADER}\n") }
+
+      it "is not valid" do
+        expect(subject).to have_error(:file, :empty)
+      end
+    end
```
jebw, line 34: Can we add a test in for a file being uploaded which is not parseable as CSV

db/migrate/20250925143706_add_eligibility_lists.rb
```diff
@@ -0,0 +1,14 @@
+class AddEligibilityLists < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
```
jebw, line 2: Since this is adding an empty table, I think we can adding the index in a blocking fashion (since it will be nanoseconds) which means we don't need to disable DDL transactions

jebw (changes requested):
Looks good - could do with a check in the service spec for if the users uploads something like an excel file

For DB transactions - unless its a table with a many thousands of rows I'd prefer we kept DDL transactions and added the index inline

jebw (approved):
LGTM 🙌

PR #3061, opened 2025-11-27
[BAU] Check feature flag before sending custom version events

jebw:
### Context

Ticket: BAU

Currently we're trying to send custom events when versions are created even if the feature flag is off for dfe analytics

1. Check feature flag before sending event

### DIFF
--- a/lib/paper_trail_extensions/version.rb
+++ b/lib/paper_trail_extensions/version.rb
@@ -1,7 +1,9 @@
 module PaperTrailExtensions
   module Version
     def self.included(base)
-      base.after_commit :send_to_dfe_analytics, on: :create
+      base.after_commit :send_to_dfe_analytics,
+                        on: :create,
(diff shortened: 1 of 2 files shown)


app/jobs/stream_versions_to_big_query_job.rb
```diff
@@ -1,9 +1,13 @@
 class StreamVersionsToBigQueryJob < ApplicationJob
+  delegate :dfe_analytics_enabled?, to: Feature
+
   self.log_arguments = false
 
   queue_as :low_priority
 
   def perform(user_name, data)
+    return unless dfe_analytics_enabled?
```
jebw: I've done this check within the Job but I can see an argument for doing it in the module and never creating the job? Thoughts

alkesh (approved):

alkesh (approved):

PR #3280, opened 2026-04-09
[NPQ-3620] update course start date form with new options

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3620

* update options on `course-start-date` page: now answering 'yes' or 'no' to starting in Autumn 2026
* new ineligible messaging if it's because the user chose 'no' (i.e. the Spring 2026 cohort)
* update specs to work with new choice
* update funding eligibility checker to return unfunded if the cohort is unfunded
* remove 'Course start' section on account page, because this was using a hardcoded single value for the course start date
* update seeds to include an unfunded cohort

### DIFF
--- a/app/forms/questionnaires/course_start_date.rb
+++ b/app/forms/questionnaires/course_start_date.rb
@@ -1,23 +1,17 @@
 module Questionnaires
   class CourseStartDate < Base
-    class Form < QuestionTypes::RadioButtonGroup
-      include ApplicationHelper
-      def type
-        "radio_button_group"
-      end
-
-      def question_text
-        "Do you want to start a course in #{application_course_start_date}?"
-      end
-    end
-
     include ApplicationHelper
 
     QUESTION_NAME = :course_start_date
 
+    OPTIONS = {
+      "2026a" => "Spring 2026",
+      "2026b" => "Autumn 2026",
+    }.freeze
+
     attribute QUESTION_NAME
 
-    validates QUESTION_NAME, presence: true, inclusion: { in: %w[yes no] }
+    validates QUESTION_NAME, presence: true, inclusion: { in: OPTIONS.keys }
 
     def self.permitted_params
       [QUESTION_NAME]
@@ -25,19 +19,15 @@ module Questionnaires
 
     def questions
       [
-        Form.new(
+        QuestionTypes::RadioButtonGroup.new(
           name: :course_start_date,
           options:,
-          style_options: { legend: { size: "m", tag: "h2" } },
         ),
       ]
     end
 
     def options
-      [
-        build_option_struct(value: "yes", link_errors: true, hint: I18n.t("helpers.hint.registration_wizard.course_start_date_hint")),
-        build_option_struct(value: "no"),
-      ]
+      OPTIONS.map { |value, label| build_option_struct(value:, label:) }
     end
 
     def requirements_met?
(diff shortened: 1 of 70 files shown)


app/forms/questionnaires/course_start_date.rb
```diff
@@ -1,59 +1,41 @@
...
-
-    include Helpers::Institution
     include ApplicationHelper
 
-    QUESTION_NAME = :course_start_date
+    QUESTION_NAME = :course_start_cohort
+
+    OPTIONS = {
+      "2026a" => "Spring 2026",
+      "2026b" => "Autumn 2026",
```
alkesh, line 9: hardcoded values. Is that ok?

app/services/handle_submission_for_store.rb
```diff
@@ -41,16 +41,21 @@ def call
...
+        **({ funded_place: false } unless cohort.funded?),
       )
       enqueue_send_application_submission_email_job(application)
     end
   end
 
 private
 
+  def cohort
+    Cohort.find_by(identifier: store["course_start_cohort"])
```
spqr-praetoria: Unlikely but can get a silent nil here

```suggestion
     Cohort.find_by(identifier: store["course_start_cohort"]) || Cohort.current
```

db/seeds/base/add_applications.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 all_courses = Course.all.to_a
-all_cohorts = Cohort.all.to_a
+all_cohorts = Cohort.where.not(funding: "unfunded").to_a
```
spqr-praetoria:
```suggestion
all_cohorts_without_funding = Cohort.where.not(funding: "unfunded").to_a
```

app/forms/questionnaires/course_start_date.rb
spqr-praetoria: Hmm, they seem simple enough, but means if we don't have exact values for say FundingEligibility. We will get a null value.

A test spec for this would help, that way if OPTIONS gets updated, 


```suggestion
   # spec/forms/questionnaires/course_start_date_spec.rb                                                                              
  describe "OPTIONS" do                                            
    it "only references cohorts that exist" do
      described_class::OPTIONS.each_key do |identifier|                                                                              
        expect(Cohort.find_by(identifier:)).to be_present,                                                                           
          "OPTIONS key #{identifier.inspect} has no matching Cohort record"                                                          
      end                                                                                                                            
    end                                                            
  end 
```

app/services/handle_submission_for_store.rb
jebw: I'd suggest changing this to be a `find_by!` - should only be a problem if something odd is going on and if that is the case we should just fail out rather than assuming a particular cohort

db/seeds/base/add_applications.rb
jebw: Rails' `enum` adds scopes automatically you could use?

```ruby
Cohort.funded.or(Cohort.capped)
```

app/forms/questionnaires/course_start_date.rb
jebw: I think thats ok for now - there's specific content which will be needed on the form (see prototype) - that content can be a follow up ticket though if necessary

app/services/handle_submission_for_store.rb
```diff
@@ -41,16 +41,21 @@ def call
         on_submission_trn: store["trn"],
         teacher_catchment_country:,
         teacher_catchment_iso_country_code:,
-        cohort: Cohort.current,
+        cohort:,
         lead_provider_approval_status: Application.lead_provider_approval_statuses[:pending],
         review_status: funding_eligibility_service.subject_to_review? ? "needs_review" : nil,
+        **({ funded_place: false } unless cohort.funded?),
```
jebw: This line is a bit complicated to read - how about setting `funded_place` from a method and that can have the appropriate logic within it

jebw (commented):
Looks great, couple of minor code suggestions

spqr-praetoria (approved):
Looks great

PR #3284, opened 2026-04-14
[NPQ-3661] - blanking the email addresses when later used

spqr-praetoria:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3661

- Data migration to allow email to be NULL and add a constraint to only allow null addreses if both archived_email and archived_at are set
- Sign in code for TeacherAuth is changed to handle when it finds an user account with an existing email address it moves the email of that existing account to the archived_email field together with setting archived_at
- Sends a message to sentry to notify that an email address has been NULLed reporting the user ecf_id field
- Tweak the API Docs to mark the email field as being NULLable

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -25,10 +25,11 @@ class User < ApplicationRecord
 
   validates :full_name, presence: true
 
+  validates :email, presence: true, unless: :archived?
   validates :email,
-            presence: true,
-            uniqueness: true,
-            notify_email: true
+            uniqueness: { allow_nil: true },
+            notify_email: true,
+            if: :email?
 
   validates :uid, uniqueness: { allow_blank: true }
   validates :ecf_id, uniqueness: { case_sensitive: false }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -25,10 +25,11 @@ class User < ApplicationRecord
 
   validates :full_name, presence: true
 
+  validates :email, presence: true, unless: :archived?
   validates :email,
-            presence: true,
             uniqueness: true,
-            notify_email: true
+            notify_email: true,
+            if: :email?
 
   validates :uid, uniqueness: { allow_blank: true }
   validates :ecf_id, uniqueness: { case_sensitive: false }
--- a/app/services/users/blank_email_for_reuse.rb
+++ b/app/services/users/blank_email_for_reuse.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Users
+  class BlankEmailForReuse
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :user
+
+    def call
+      ApplicationRecord.transaction do
+        user.update!(
+          archived_email: user.email,
+          archived_at: Time.zone.now,
+          email: nil,
+          uid: nil,
+          provider: nil,
+        )
+      end
+
+      Sentry.capture_message(
+        "Blanked email on the user due to reuse when used by a later participant",
+        level: :info,
+        extra: { ecf_id: user.ecf_id },
+      )
+    end
+  end
+end
--- a/app/services/users/find_or_create_from_teacher_auth.rb
+++ b/app/services/users/find_or_create_from_teacher_auth.rb
@@ -26,10 +26,12 @@ module Users
       user_matched_using_uid = User.find_by(provider: Omniauth::Strategies::TeacherAuth::NAME, uid:)
 
       if user_matched_using_uid
+        blank_clashing_email_user(except: user_matched_using_uid)
         user_matched_using_uid.update!(email: email, trn:, trn_verified: true, trn_auto_verified: true, full_name:, feature_flag_id:, previous_names:)
         return user_matched_using_uid
       end
 
+      blank_clashing_email_user
       create_user_with_provider_data
     end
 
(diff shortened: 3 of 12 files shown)


app/services/users/blank_email_for_reuse.rb
alkesh, line 1: would it be better to use `Users::Archiver` and update that with an option `blank_email: true`,
or maybe add a method on that service called `archive_and_blank_email!`,
as this service is still archiving a user, and it makes sense to have that logic in one place.

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -39,6 +41,16 @@ def matching_users
       @matching_users ||= User.where(trn:, trn_verified: true, archived_at: nil).order(updated_at: :desc).all.to_a
     end
 
+    def blank_clashing_email_user(except: nil)
+      scope = User.where(email:).where(archived_at: nil)
```
jebw: I don't think this needs to check for `archived_at: nil` - otherwise if we have an archived user whose `email` field is still populated with a clashing address then we'll still not be able to create the new user

db/migrate/20260413174148_allow_null_email_on_users.rb
```diff
@@ -0,0 +1,12 @@
+class AllowNullEmailOnUsers < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :users, :email, true
+
+    add_check_constraint(
```
peteryates, line 5: 👌

db/migrate/20260413174148_allow_null_email_on_users.rb
```diff
@@ -0,0 +1,12 @@
+class AllowNullEmailOnUsers < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :users, :email, true
```
alkesh, line 3: I think you need to add:
```
change_column_default :users, :email, from: "", to: nil
```
otherwise the check constraint won't prevent creating an archived user with no email, because the email will be an empty string, and not actually null.

spqr-praetoria: check-constraint

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -39,6 +41,16 @@ def matching_users
...
+      scope = scope.where.not(id: except.id) if except
```
alkesh, line 80: this line is not covered by any spec - is it needed?

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -39,6 +41,16 @@ def matching_users
...
+      clashing = scope.first
+      return unless clashing
+      return if clashing.trn.present? && clashing.trn == trn
```
alkesh: this line is not covered by any spec - is it needed?

app/models/user.rb
```diff
@@ -25,10 +25,11 @@ class User < ApplicationRecord
...
   validates :full_name, presence: true
 
+  validates :email, presence: true, unless: :archived?
   validates :email,
-            presence: true,
-            uniqueness: true,
-            notify_email: true
+            uniqueness: { allow_nil: true },
+            notify_email: true,
+            if: :email?
```
alkesh, line 32: this line is not covered by any spec - is it needed?

app/models/user.rb
```diff
@@ -25,10 +25,11 @@ class User < ApplicationRecord
 
   validates :full_name, presence: true
 
+  validates :email, presence: true, unless: :archived?
   validates :email,
-            presence: true,
-            uniqueness: true,
-            notify_email: true
+            uniqueness: { allow_nil: true },
```
alkesh: the spec still passes with this line as `uniqueness: true`

app/models/user.rb
```diff
@@ -25,10 +25,11 @@ class User < ApplicationRecord
 
   validates :full_name, presence: true
 
+  validates :email, presence: true, unless: :archived?
```
alkesh, line 28: the spec still passes without the `unless: :archived?`

alkesh (changes requested):

app/services/users/find_or_create_from_teacher_auth.rb
spqr-praetoria: Yes, it's needed, it's used above, so we do not find the same user_matched_using_uid on line 29 abobe as part of a guard clause.

I have added a spec to cover this now.

spec/swagger_schemas/models/participant.rb
```diff
@@ -29,9 +29,9 @@
             example: "Isabelle MacDonald",
           },
           email: {
-            description: "The email address registered for this NPQ participant",
+            description: "The email address registered for this NPQ participant. May be null if the participant's email has been archived due to reuse by a different participant.",
```
alkesh: I wonder if we should be updating this description at all - that description is confusing:
- the participant's email is not archived, it's the participant that is archived
- "reuse by a different participant" may be confusing, as it's really the same participant, but internally we have clashing database rows.
It's probably enough to just have the `nullable: true`.

alkesh (approved):

PR #3285, opened 2026-04-14
[NPQ-3548] email notifications when application status and course outcome changes

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3548

send email notifications to the user when:
* application is accepted
* application is rejected
* application is deferred
* application is withdrawn
* application is resumed
* course is completed - passed
* course is completed - failed

### DIFF
--- a/app/mailers/application_deferred_mailer.rb
+++ b/app/mailers/application_deferred_mailer.rb
@@ -1,0 +1,14 @@
+class ApplicationDeferredMailer < ApplicationMailer
+  TEMPLATE_ID = "6e29f6f2-5c9d-4106-baa7-7fd9aeed63d6".freeze
+
+  def application_deferred_mail(to:, full_name:, provider_name:, course_name:, ecf_id:)
+    template_mail(TEMPLATE_ID,
+                  to:,
+                  personalisation: {
+                    full_name:,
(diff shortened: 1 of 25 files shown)


jebw (approved):
LGTM - only query is just to confirm the email notification when a participant goes from Deferred to Active

This would seem to make sense to me but worth checking

app/mailers/application_deferred_mailer.rb
```diff
@@ -0,0 +1,14 @@
+class ApplicationDeferredMailer < ApplicationMailer
+  TEMPLATE_ID = "6e29f6f2-5c9d-4106-baa7-7fd9aeed63d6".freeze
```
spqr-praetoria, line 2: Just a question, is this ok or should it be an env variable

spqr-praetoria (commented):
Main question on TEMPLATE_ID

spqr-praetoria (approved):

app/mailers/application_deferred_mailer.rb
alkesh: that should be a constant - it's only applicable to this mailer, and it matches the pattern we've used for the other mailers.

PR #3293, opened 2026-04-20
[BAU] switch to versioned_environment

alkesh:
### Context

BAU

switch from `environment` to `versioned_environment` for all rake tasks that write to the database (or have the potential to write to the database).


jebw (approved):
LGTM

PR #3294, opened 2026-04-20
[NPQ-3669] course cohort provider datamodel

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3669
(1st PR - 2nd one is here: https://github.com/DFE-Digital/npq-registration/pull/3296)

* datamodel for `CourseCohort` and `CourseCohortProvider`
* rake task to load `CourseCohortProvider`s


config/analytics.yml
```diff
@@ -291,3 +291,15 @@ shared:
...
+    - course_cohort_id
+    - lead_provider_id
+    - created_at
+    - updated_at
+  :course_cohorts:
+    - id
+    - course_id
+    - cohort_id
+    - created_at
+    - updated_at
```
jebw, line 305: Can you check with data they are ready to receive this - if not we can just block it for now since they weren't receiving it anyway when it was hard coded

jebw (approved):
LGTM 🙌

config/analytics.yml
alkesh: 👍🏽 I sent these tables to Theo last week, and got an acknowledgement reply

PR #3296, opened 2026-04-21
[NPQ-3669] switch to using course cohort providers

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3669
2nd PR (1st is here: https://github.com/DFE-Digital/npq-registration/pull/3294)

* switch to using database `CourseCohort`s and `CourseCohortProivder`s instead of the hardcoded mapping in `LeadProvider`
* remove hardcoded mapping in `LeadProvider`
* update DB seeds so that cohorts up to 2025 will have the current (default) mappings
* cohort 2026a will have the new unfunded mappings, and 2026b will have the full default mappings
* new `:with_cohorts` tag and associated context to allow specs to load the default set of `CourseCohortProivder`s
* updated the journey step helper `choose_a_school` to work for when a school has previously been selected with no JS


app/forms/questionnaires/choose_your_npq.rb
```diff
@@ -129,7 +129,7 @@ def lead_provider_valid?
     end
 
     def valid_providers
-      LeadProvider.for(course:)
+      LeadProvider.for(course:, cohort: Cohort.find_by!(identifier: wizard.query_store.course_start_cohort))
```
jebw: Maybe split this into 2 lines - would read a lot simpler as

```ruby

cohort = ....
LeadProvider.for(course:, cohort:)
```

spec/features/journeys/happy_paths/unfunded_cohort_registration_journey_spec.rb
```diff
@@ -1,13 +1,18 @@
...
   include Helpers::JourneyStepHelper
   include ApplicationHelper
 
-  # include_context "retrieve latest application data"
   include_context "Stub Get An Identity Omniauth Responses"
 
+  before do
+    # create course cohort providers for the unfunded spring 2026a cohort
+    file_name = "db/seeds/data/unfunded_spring_2026a_course_cohort_providers.csv"
+    CourseCohortProviders::Updater.new(cohort: Cohort.find_by(identifier: "2026a"), course_to_provider_csv: file_name, dry_run: false).call
```
jebw: Better adding factories and using that - you can just create the data you require for the spec and avoids importing all the data needed for a 'full mapping' for production

spec/support/with_cohorts.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    # create an unfunded cohort for the next cohort
+    FactoryBot.create(:cohort, :unfunded, start_year: next_cohort_year)
+
+    # create a capped 'b' cohort for the next cohort
+    FactoryBot.create(:cohort, start_year: next_cohort_year, suffix: "b")
+
+    # create course cohort providers for the next capped cohort
+    file_name = "db/seeds/data/default_course_cohort_providers.csv"
+    CourseCohortProviders::Updater.new(cohort: Cohort.find_by(identifier: "#{next_cohort_year}b"), course_to_provider_csv: file_name, dry_run: false).call
```
jebw: This will do 80 odd INSERTs for every spec using this shared context - if possible I think it should be calling factories directly to create a 'few' designated courses for a couple of LPs in the specified cohort? eg

```
FactoryBot.create :cohort, start_year: 2024, lead_providers:, courses: 
```

spec/support/with_cohorts.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  after(:context) do
+    Cohort.destroy_all
+  end
```
jebw, line 26: This should be unneeded since the transaction will be rolled back afterwards anyway

jebw (commented):
Looks great but I'm cautious about the volume of data it creates in the shared context?

spec/support/with_cohorts.rb
alkesh: it was needed previously, because I was using `before(:context)` - but I've now switched it to a standard before example hook, with a small subset of factory created course cohort providers, so it's no longer needed.

jebw (approved):
LGTM 🙌

PR #3304, opened 2026-04-24
[NPQ-3667] allow unverified GAI account to be converted to a teacher auth account

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3667

* if a user logs in with teacher auth, and we find a matching account that is DfE Identity, with the same email and TRN, but the TRN is unverified - we will update the provider and UID on the matched account, and allow the user to log in.
* The TRN will be marked as verified
* enable teacher auth for review apps


jebw (approved):
LGTM

PR #3312, opened 2026-04-29
[NPQ-3687] Initial minimal wizard

jebw:
### Context

Ticket: [NPQ-3687](https://dfedigital.atlassian.net/browse/NPQ-3687)

Whilst our existing wizard works fairly well, we have various oddities  
* these stem from two different understandings of the route through the wizard (forwards and backwards)
* presenting, and then saving, data which is 'out of scope', ie user went down a branch and then back tracked
* Odd bugs stemming from attempting to redirect a user to an appropriate location within the wizard without reference to what data they've already entered.

1. Initial implementation of a replacement registration wizard - this is very minimal at this stage and isn't exposed via a controller yet but can co-exist alongside the existing wizard

Note: I've tried to strip out anything extraneous or untestable from this initial set of changes but I know the routing logic for check_answers is in there. This isn't really testable till later steps are added so when I'll add specs for that in the follow up PRs

Note 2: The SonarScanner warnings on the `require` lines are technnically correct but this is a conscious choice to not load dependencies until actually required as part of visualisation generation 

Note 3: I need to add an ADR, will do that an revise PR

[NPQ-3687]: https://dfedigital.atlassian.net/browse/NPQ-3687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/rails_helper.rb
```diff
@@ -43,7 +43,7 @@
   config.default_max_wait_time = 10 # without this, the default is 2
 end
 
-require "capybara-screenshot/rspec"
+require "capybara-screenshot/rspec" unless ENV["NO_CAPYBARA_SCREENSHOTS"] == "true"
```
alkesh, line 46: might be worth documenting this env var in `docs/specs_and_linting.md`

app/forms/questionnaires/base.rb
```diff
@@ -1,16 +1,16 @@
...
     include ActiveModel::Validations::Callbacks
     include Questionnaires::FlowHelper
 
-    attr_accessor :wizard
-
     def self.permitted_params
       []
     end
 
+    def to_s = self.class.model_name.element
```
alkesh, line 11:
```suggestion
    def to_s = self.class.model_name.element
    
```
is it worth having a blank line separating these methods? I'm not hugely fussed either way really.

spec/wizards/registration/wizard_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      it { is_expected.to have_flow_path(%i[start course_start_date check_answers]) }
+      it { is_expected.to have_valid_path(%i[start]) }
+    end
+
+    context "when yes at course_start_date" do
+      let(:state) { { started: true, course_start_date: "yes" } }
+
+      it { is_expected.to have_flow_path(%i[start course_start_date check_answers]) }
+      it { is_expected.to have_valid_path(%i[start course_start_date check_answers]) }
```
alkesh: I think this PR will clash with my PR: https://github.com/DFE-Digital/npq-registration/pull/3280
It's probably worth me waiting for this PR to be merged first - wdyt?

app/forms/questionnaires/base.rb
jebw: My take (and I know rubocop follows the same pattern for specs) tends to be if a method/codeblock is a oneliner - then pack them together, if they become multiline then always separate them out?

alkesh:
looks good - I think it just needs the ADR you mentioned

alkesh (approved):

PR #3317, opened 2026-05-05
[NPQ-3490] - store long lived auth token

spqr-praetoria:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3490

- Store long lived tokens
- Allow users to quietly sign back in after at a later date

### DIFF
--- a/app/jobs/refresh_user_token_job.rb
+++ b/app/jobs/refresh_user_token_job.rb
@@ -1,0 +1,14 @@
+class RefreshUserTokenJob < ApplicationJob
+  queue_as :default
+
+  def perform(user_id)
+    user = User.find_by(id: user_id)
+    return unless user&.needs_token_refresh?
+
+    token_record = user.oauth_token
+    new_token = TeacherAuth::RefreshAccessToken.call(refresh_token: token_record.token)
+    token_record.update!(token: new_token, token_updated_at: Time.current)
+  rescue StandardError => e
+    Sentry.capture_exception(e, extra: { user_id: })
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -18,6 +18,7 @@ class User < ApplicationRecord
 
   has_paper_trail meta: { note: :version_note }, ignore: %i[raw_tra_provider_data updated_at feature_flag_id]
 
+  has_many :oauth_tokens, dependent: :destroy
   has_many :applications, dependent: :destroy
   has_many :declarations, through: :applications
   has_many :participant_id_changes, -> { order("created_at desc") }
@@ -42,6 +43,20 @@ class User < ApplicationRecord
     where.not(uid: nil)
          .where(provider: Omniauth::Strategies::TraOpenidConnect::NAME)
   }
+
+  scope :needing_token_refresh, lambda {
+    where(trn: nil)
+      .joins(:oauth_tokens)
+      .where(oauth_tokens: { token_type: "refresh_token", token_updated_at: ...7.days.ago })
+  }
+
+  def oauth_token
+    oauth_tokens.refresh_token.first
+  end
+
+  def needs_token_refresh?
+    trn.blank? && oauth_token.present?
+  end
 
   EMAIL_UPDATES_STATES = %i[senco other_npq].freeze
   EMAIL_UPDATES_ALL_STATES = [:empty] + EMAIL_UPDATES_STATES
--- a/app/services/applications/reject.rb
+++ b/app/services/applications/reject.rb
@@ -14,7 +14,10 @@ module Applications
     def reject
       return false unless valid?
 
-      application.update!(lead_provider_approval_status: "rejected", reason_for_rejection:)
+      ApplicationRecord.transaction do
+        application.update!(lead_provider_approval_status: "rejected", reason_for_rejection:)
+        clear_token_if_no_remaining_applications
+      end
       application.reload
 
       ApplicationRejectedMailer.application_rejected_mail(
@@ -29,6 +32,15 @@ module Applications
     end
 
   private
+
+    def clear_token_if_no_remaining_applications
+      user = application.user
+      return if user.trn.present?
+      return if user.oauth_token.blank?
+      return if user.applications.where(lead_provider_approval_status: "pending").where.not(id: application.id).exists?
+
+      user.oauth_token.destroy!
+    end
 
     def not_already_rejected
       return unless application
--- a/app/services/teacher_auth/refresh_access_token.rb
+++ b/app/services/teacher_auth/refresh_access_token.rb
@@ -1,0 +1,31 @@
+module TeacherAuth
+  class RefreshAccessToken
+    include HTTParty
+
+    default_timeout 15.seconds
+    raise_on 400..599
+
+    def self.call(refresh_token:)
+      new(refresh_token:).call
+    end
+
+    def initialize(refresh_token:)
+      @refresh_token = refresh_token
+    end
+
+    def call
+      response = self.class.post(
+        "#{Rails.configuration.x.teacher_auth.domain.to_s.chomp('/')}/oauth2/token",
+        body: {
+          grant_type: "refresh_token",
+          refresh_token: @refresh_token,
+          client_id: Rails.configuration.x.teacher_auth.client_id,
+          client_secret: Rails.configuration.x.teacher_auth.client_secret,
+        },
+        headers: { "Content-Type" => "application/x-www-form-urlencoded" },
+      )
+
+      response.parsed_response["refresh_token"]
+    end
+  end
+end
--- a/app/services/users/find_or_create_from_teacher_auth.rb
+++ b/app/services/users/find_or_create_from_teacher_auth.rb
@@ -3,6 +3,7 @@
 module Users
   class FindOrCreateFromTeacherAuth
     def initialize(provider_data:, feature_flag_id:)
+      @provider_data = provider_data
       @access_token = provider_data.credentials.token
       @uid = provider_data.uid
       @trn = provider_data.extra.raw_info.trn
@@ -12,20 +13,24 @@ module Users
       @date_of_birth = Date.parse(provider_data.extra.raw_info.verified_date_of_birth, "%Y-%m-%d")
     end
 
-    attr_reader :access_token, :uid, :trn, :email, :full_name, :date_of_birth, :feature_flag_id
+    attr_reader :provider_data, :access_token, :uid, :trn, :email, :full_name, :date_of_birth, :feature_flag_id
 
     def call
       user_matched_using_trn = verified_trn_matching_users.first
 
       if user_matched_using_trn
-        user_matched_using_trn.update!(
-          uid:,
-          provider: Omniauth::Strategies::TeacherAuth::NAME,
-          email:,
-          full_name:,
-          feature_flag_id:,
-          previous_names:,
-        )
+        ApplicationRecord.transaction do
+          user_matched_using_trn.update!(
+            uid:,
+            provider: Omniauth::Strategies::TeacherAuth::NAME,
+            email:,
+            full_name:,
+            feature_flag_id:,
+            previous_names:,
+          )
+          persist_token(user_matched_using_trn, provider_data)
+        end
+
         merge_and_archive_other_users(user_matched_using_trn, verified_trn_matching_users[1..])
         return user_matched_using_trn
       end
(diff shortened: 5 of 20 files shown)


spec/services/exporters/contracts_spec.rb
```diff
@@ -71,11 +71,11 @@
     subject { described_class.new(cohort: older_cohort).call }
 
     let(:older_cohort) { create(:cohort, start_year: 2022) }
-    let(:older_statement_output_fee_true) { create(:statement, cohort: older_cohort, lead_provider:, output_fee: true, for_date: 3.months.ago) }
-    let(:older_statement_output_fee_false) { create(:statement, cohort: older_cohort, lead_provider:, output_fee: false, for_date: 2.months.ago) }
-    let(:latest_statement_output_fee_true) { create(:statement, cohort: older_cohort, lead_provider:, output_fee: true, for_date: 1.month.ago) }
-    let(:latest_statement_output_fee_false) { create(:statement, cohort: older_cohort, lead_provider:, output_fee: false, for_date: 0.months.ago) }
-    let(:future_statement_output_fee_true) { create(:statement, cohort: older_cohort, lead_provider:, output_fee: true, for_date: 1.month.from_now) }
```
jebw, line 78: This shouldn't be here - this reverts Alkesh' recent fix for a flaky spec

spec/services/users/find_or_create_from_provider_data_spec.rb
```diff
@@ -327,6 +329,64 @@
...
+      let(:provider_data_refresh_token) { nil }
+
+      it "does not set the refresh token" do
+        expect(subject.user_refresh_token).to be_nil
+      end
+
+      it "does not enqueue a RefreshUserTokenJob" do
+        expect { subject }.not_to have_enqueued_job(RefreshUserTokenJob)
+      end
+    end
```
jebw: Can you add a spec here to check sentry was alerted that we didn't get a refresh token back for some reason

jebw: Think you'll also need to change it to alert sentry

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -34,37 +41,77 @@ def call
...
+        schedule_refresh(unverified_trn_matching_user) if new_refresh_token_assigned
         return unverified_trn_matching_user
       end
 
       blank_clashing_email_user
-      create_user_with_provider_data
+
+      new_refresh_token_assigned = false
+      user = nil
+      Rails.logger.info("[HELLO4] Create user with provider data")
```
jebw: ????

app/services/teacher_auth/refresh_access_token.rb
```diff
@@ -0,0 +1,31 @@
...
+        body: {
+          grant_type: "refresh_token",
+          refresh_token: @refresh_token,
+          client_id: Rails.configuration.x.teacher_auth.client_id,
+          client_secret: Rails.configuration.x.teacher_auth.client_secret,
+        },
+        headers: { "Content-Type" => "application/x-www-form-urlencoded" },
+      )
+
+      response.parsed_response["refresh_token"] || @refresh_token
```
jebw: This makes no sense - if the API returned a success resposnse, then `@refresh_token` is now used up and cannot be resused again so returning it will lead to the application incorrectly updating with an expired token

app/services/applications/reject.rb
```diff
@@ -14,7 +14,10 @@ class Reject
     def reject
       return false unless valid?
 
-      application.update!(lead_provider_approval_status: "rejected", reason_for_rejection:)
+      ApplicationRecord.transaction do
+        application.update!(lead_provider_approval_status: "rejected", reason_for_rejection:)
+        clear_refresh_token_if_orphaned
```
jebw: ~Orphaned isn't a term we're using so may lead to confusion for other developers - how about `clear_refresh_token_if_no_remaining_applications!`~

app/services/applications/reject.rb
```diff
@@ -30,6 +33,15 @@ def reject
 
   private
 
+    def clear_refresh_token_if_orphaned
+      user = application.user
+      return if user.trn.present?
```
jebw, line 38: ~If TRN is present then we don't need the refresh token any more so can clear it~

app/services/applications/reject.rb
```diff
@@ -30,6 +33,15 @@ def reject
...
+      return if user.user_refresh_token.blank?
```
jebw: `.nil?` would be better here since you're checking for absence of a record

app/services/applications/reject.rb
```diff
@@ -30,6 +33,15 @@ def reject
...
+      return if user.applications.where(lead_provider_approval_status: "pending").where.not(id: application.id).exists?
```
jebw, line 40: ~Logically, it cannot be the id of this application because it got its status changed in the line prior to calling this method?~

~`return if user.applications.pending_lead_provider_approval_status.any?`~

app/models/user.rb
```diff
@@ -43,6 +44,16 @@ class User < ApplicationRecord
          .where(provider: Omniauth::Strategies::TraOpenidConnect::NAME)
   }
 
+  scope :needing_token_refresh, lambda {
+    where(trn: nil)
+      .joins(:user_refresh_token)
+      .where(user_refresh_tokens: { refresh_token_updated_at: ...7.days.ago })
+  }
```
jebw, line 51: ~This relates to tokens so I think belongs in that model - we'd also talked about refreshing on a shorter time frame so can this be 72.hours.ago, and can the value itself be held on a constant to make it trivial for future tweaking if needed~

app/models/user.rb
```diff
@@ -43,6 +44,16 @@ class User < ApplicationRecord
...
+
+  def needs_token_refresh?
+    trn.blank? && user_refresh_token.present?
```
jebw: This has difference logic to the scope - it should presumably be including the same age related check

jebw: I think it probably makes more sense on the token class

app/services/teacher_auth/refresh_access_token.rb
```diff
@@ -0,0 +1,31 @@
+module TeacherAuth
+  class RefreshAccessToken
+    include HTTParty
+
+    default_timeout 5.seconds
```
jebw: Given this will be getting called from a Job I think we can give this longer to respond - say `15.seconds`

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -12,20 +13,26 @@ def initialize(provider_data:, feature_flag_id:)
...
+            provider: Omniauth::Strategies::TeacherAuth::NAME,
+            email:,
+            full_name:,
+            feature_flag_id:,
+            previous_names:,
+          )
+          new_refresh_token_assigned = persist_refresh_token(user_matched_using_trn, provider_data)
+        end
+
+        schedule_refresh(user_matched_using_trn) if new_refresh_token_assigned
```
jebw: This shouldn't be needed if we're using an hourly job

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -34,37 +41,77 @@ def call
...
+        user.user_refresh_token.destroy!
+        false
+      else
+        false
+      end
+    end
+
+    def schedule_refresh(user)
+      RefreshUserTokenJob.set(wait: 7.days).perform_later(user.id)
+    end
```
jebw: Cron job supercedes this

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -3,6 +3,7 @@
 module Users
   class FindOrCreateFromTeacherAuth
```
jebw, line 4: ~The `call` method in this class are getting very complicated and its because there are multiple places all trying to do the same 'after sorting user out'  action of persisting the token.~

~I can see 2 approaches~

* ~Move the `persist the token` out to the omniauth controller (with some helper code on the token class) - this logic is orthogonal to signing the user in.~

```ruby
@user = Users::FindOrCreateFromTeacherAuth
if @user
  @user.set_refresh_token(provider_data.credentials&.fetch(:token)) if @user.trn.blank?
  # ....
end
```

3. ~Or alternatively, add a level of indirection within this class - move the existing `#call` method to be a differently named private method, eg `sign_in_user` or similar since that always returns a user class. Then add a new `#call` method with something like the following~
```ruby
def call
  sign_in_user(...).tap do |user|
    set_refresh_token_on_user(user) if user && user.trn.blank?
  end
end
```

db/migrate/20260505115737_create_user_refresh_tokens.rb
```diff
@@ -0,0 +1,12 @@
+class CreateUserRefreshTokens < ActiveRecord::Migration[8.0]
+  def change
+    create_table :user_refresh_tokens do |t|
+      t.references :user, null: false, foreign_key: true, index: { unique: true }
+      t.text :refresh_token, null: false
+      t.datetime :refresh_token_updated_at, null: false
+      t.timestamps
+    end
```
jebw: Could do with a chat around the schema on this

jebw (dismissed):
Looks good - I've a few comments on changes - can you review them then ping me when you're ready for a catch up, thanks

app/jobs/refresh_user_token_job.rb
```diff
@@ -0,0 +1,15 @@
...
+  def perform(user_id)
+    user = User.find_by(id: user_id)
+    return unless user&.needs_token_refresh?
+
+    token_record = user.user_refresh_token
+    new_refresh_token = TeacherAuth::RefreshAccessToken.call(refresh_token: token_record.refresh_token)
+    token_record.update!(refresh_token: new_refresh_token, refresh_token_updated_at: Time.current)
+  rescue StandardError => e
+    Sentry.capture_exception(e, extra: { user_id: })
+    raise
```
jebw: This shouldn't re-raise otherwise you'll end up with a thundering herd problem when you chain jobs

1) 9am - Enqueue job kicks of 1000 api jobs
2) they fail because endpoint is unavailable
3) those 1000 jobs reschedule themselves
4) and keep rescheduling them selves with increasing backoff
5) meanwhile at 10am - Enqueue job kicks off the same 1000 api jobs for a second time
6) and so the loop continues

jebw (commented):
See responses in earlier approval (I'd meant to just comment rather than approve so I've dismissed that one)

spec/services/exporters/contracts_spec.rb
spqr-praetoria: Will update

app/services/users/find_or_create_from_teacher_auth.rb
spqr-praetoria: Removed leftover logging

app/models/omniauth_token.rb
```diff
@@ -0,0 +1,14 @@
+class OmniauthToken < ApplicationRecord
+  belongs_to :user
+
+  enum :token_type, {
+    refresh_token: "refresh_token",
+    access_token: "access_token",
```
jebw: We're not persisting `access_tokens`, so we don't need this in the enum

app/models/omniauth_token.rb
```diff
@@ -0,0 +1,14 @@
...
+  }
+
+  encrypts :refresh_token
+
+  validates :refresh_token, presence: true
+  validates :refresh_token_updated_at, presence: true
```
jebw: This was made into a generic omniauth_tokens table so it could handle other types in the future if necessary (by just adding to the enum definition)

With that in mind can these be renamed to `token` and `token_updated_at`

db/migrate/20260505115737_create_omniauth_tokens.rb
```diff
@@ -0,0 +1,15 @@
+class CreateOmniauthTokens < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :omniauth_token_types, %w[refresh_token access_token]
```
jebw: Can you take access token out of the enum list since we're not persisting those

db/migrate/20260505115737_create_omniauth_tokens.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    create_table :omniauth_tokens do |t|
+      t.references :user, null: false, foreign_key: true, index: { unique: true }
```
jebw: The uniqueness index needs to cover both `user_id` and `token_type` otherwise a user can only store a token of one type

db/migrate/20260505115737_create_omniauth_tokens.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.enum :token_type, enum_type: "omniauth_token_types", default: "refresh_token", null: false
+      t.text :refresh_token, null: false
+      t.datetime :refresh_token_updated_at, null: false
```
jebw: Since we're generalising this table to allow for other future token types, can the columns just be called `token` and ` token_updated_at`

app/services/users/find_or_create_from_provider_data.rb
```diff
@@ -25,8 +25,8 @@ def call
 
       assign_provider_data(user, provider_data)
       user.assign_attributes(feature_flag_id: feature_flag_id)
-
-      user.tap(&:save)
```
jebw, line 29: This shouldn't be changing

jebw (commented):
Very close, few remaining tweaks still needed then we can get this merged

spec/services/users/find_or_create_from_provider_data_spec.rb
```diff
@@ -13,6 +13,7 @@
   let(:provider_data_first_name) { "John" }
   let(:provider_data_last_name) { "Doe" }
   let(:provider_data_trn_lookup_status) { "Found" }
+  let(:provider_data_refresh_token) { SecureRandom.hex(32) }
```
jebw: This shouldn't be in here - we wont be getting refresh tokens from GAI and aren't using GAI

spec/services/users/find_or_create_from_provider_data_spec.rb
```diff
@@ -29,6 +30,7 @@
       }),
       credentials: OpenStruct.new({
         "token" => SecureRandom.uuid,
+        "refresh_token" => provider_data_refresh_token,
```
jebw: likewise

spec/services/exporters/contracts_spec.rb
jebw: This needs to go back to the code which is on `main` - there's no need for a change to our TeacherAuth integration to require specs changes on an admin feature relating to contracts

jebw (commented):
Getting there - there are spec changes to unrelated code
* contracts exporter
* GAI service spec (`find_or_create_from_provider_data`)

jebw (approved):
LGTM 🚀 🙌

PR #3320, opened 2026-05-06
[NPQ-3718] Tweaked dependabot.yml

jebw:
### Context

Ticket: [NPQ-3718](https://dfedigital.atlassian.net/browse/NPQ-3718)

1. Check for ruby deps daily at 7am -> means they are always ready at the start of the day, I'm presuming crontab is off've UTC
2. Check JS deps weekly, there is constant churn with JS and it just creates noise for deps we primarily use for building assets - the actually assets we include client side have low rates of churn
3. Added cooldown period for javascript version updates
4. Only raise PRs for patch version changes with Docker images and `rails` gems - we're only getting notified of changes to the latest version rather than the patch updates we wish to know about

[NPQ-3718]: https://dfedigital.atlassian.net/browse/NPQ-3718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,17 +4,31 @@ updates:
   directory: "/"
   schedule:
     interval: daily
-  open-pull-requests-limit: 10
+    time: "07:00"
+  allow:
+    - dependency-name: "rails"
+      update-types:
+      - version-update:semver-patch
+  open-pull-requests-limit: 20
 - package-ecosystem: npm
   directory: "/"
+  cooldown:
+    default-days: 5
   schedule:
-    interval: daily
-  open-pull-requests-limit: 10
+    interval: weekly
+    time: "07:00"
+    day: tuesday
+  open-pull-requests-limit: 20
 - package-ecosystem: docker
   directory: "/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,17 +4,31 @@ updates:
   directory: "/"
   schedule:
     interval: daily
-  open-pull-requests-limit: 10
+    time: "07:00"
+  allow:
+    - dependency-name: "rails"
+      update-types:
+      - version-update:semver-patch
```
alkesh, line 11: should we include `version-update:semver-minor` here as well? so that we allow minor _and_ patch updates?

alkesh (approved):

.github/dependabot.yml
alkesh: ah - I see - we're only wanting dependabot PRs for patch upgrades,
because minor version upgrades we'll have separate tickets for.

PR #3333, opened 2026-05-13
[NPQ-3729] account page fix for pre-2026 applications

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3729

cope with pre-2026 cohorts that do not have an entry in `Questionnaires::CourseStartDate::OPTIONS`

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -39,7 +39,8 @@ module ApplicationHelper
   end
 
   def application_cohort_description(application)
-    Questionnaires::CourseStartDate::OPTIONS[application.cohort.identifier][:cohort_description]
+    cohort_option = Questionnaires::CourseStartDate::OPTIONS[application.cohort.identifier]
+    cohort_option ? cohort_option[:cohort_description] : application.cohort.start_year
   end
 
   def show_otp_code_in_ui(current_env, admin)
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -39,7 +39,8 @@ def rejected?(application)
   end
 
   def application_cohort_description(application)
-    Questionnaires::CourseStartDate::OPTIONS[application.cohort.identifier][:cohort_description]
+    cohort_option = Questionnaires::CourseStartDate::OPTIONS[application.cohort.identifier]
+    cohort_option ? cohort_option[:cohort_description] : application.cohort.name
```
jebw: I can see a couple of issues with this

1. It will bitrot over time - as the wizard gets changed, this will break
2. Pre-2026 we don't want participants knowing about the different cohorts

I think for now, change `application.cohort.name` to `application.cohort.year`

But can you also create a ticket to revisit this screen and dump it in something like Sprint 142 (ie after we've gone live) - I think we want to add a 'public description' type field to cohorts so we can have multiple internal cohorts showing the same string to the user but thats a bigger change

jebw (commented):
Looks good, only change is to not expose `Cohort#name` since thats would highlight the multiple 2025 cohorts

jebw (approved):

PR #3339, opened 2026-05-15
[NPQ-3735] change position and descriptions of courses

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3735

* update position and descriptions of courses - needs `rake courses:update` to be run
* changed ECF IDs of courses, so they match production
* made the names of the courses (only shown in the admin console) consistent with what the user sees


jebw (approved):
LGTM

PR #3346, opened 2026-05-19
[NPQ-3734] new start now behaviour and sign in button

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3734

* there's now only one Start now button, and it uses teacher auth
* start now button goes directly to journey start, even if the user has applications
* sign in button goes to the user's previous applications
* removed `teacher_reference_number` and `dont_have_teacher_reference_number` wizard steps
* renamed `get_an_identity_callback` step to `login_callback`
* late registration and closed registration still uses DfE Identity login
* enable teacher auth on staging
* do not persist date_of_birth any more for teacher auth login - we don't need this any more


app/assets/stylesheets/application.sass.scss
```diff
@@ -69,3 +69,13 @@ $admin-template-max-width: 1200px;
...
+
+.sign-in-button {
+  form {
+    display: inline;
+  }
+  button {
+    vertical-align: baseline;
+    margin-bottom: govuk-spacing(1);
+  }
+}
```
jebw: This is really not GDS compliant

I thought the conclusion was a a Sign In button below the paragraph rather than in the middle of it?

app/controllers/omniauth_controller.rb
```diff
@@ -121,14 +121,14 @@ def after_sign_in_path_for(user)
       return new_email_update_path
     end
 
-    return account_path if user.applications.any?
+    return account_path unless request.env["omniauth.params"]["start_now"] == "true"
```
jebw: Can the check be moved to its own appropriately named private method, eg `user_starting_registration?`

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -75,6 +70,16 @@ def unverified_trn_matching_user
         User.find_by(provider: Omniauth::Strategies::TraOpenidConnect::NAME, trn:, trn_verified: false, email:, archived_at: nil)
     end
 
+    def always_updated_attributes
+      {
+        date_of_birth:,
```
jebw: Appreciate this is existing code but I'd not realised we were persisting `date_of_birth` - we shouldn't be - we don't really have a use for it, we don't share it with providers and have stopped exposing it in the UI - linking to TRS instead.

Can `data_of_birth` be dropped, thanks

app/views/registration_wizard/change_your_course_or_provider.html.erb
```diff
@@ -17,10 +17,6 @@
 
     <p class="govuk-body">If you have already started your course, contact your provider to make any changes.</p>
 
-    <% if Feature.trn_required? && current_user.trn.blank? %>
-      <%= govuk_start_button(text: "Submit a new registration", href: registration_wizard_show_path(:teacher_reference_number)) %>
-    <% else %>
-      <%= govuk_start_button(text: "Submit a new registration", href: registration_wizard_show_path("course-start-date")) %>
-    <% end %>
+    <%= govuk_start_button(text: "Submit a new registration", href: registration_wizard_show_path("course-start-date")) %>
```
jebw: If we're keeping the `first_questionnaire_step` method - lets use it here (but symbolized)

app/views/registration_wizard/start.html.erb
```diff
@@ -66,19 +66,30 @@
...
-            button_text: "Start now with Teacher Auth"
+            button_text: "Start now"
           )
         %>
+
+      <h2 class="govuk-heading-m">Sign in</h2>
+
+      <span class="govuk-body sign-in-button">
+        You can also
+        <%= govuk_button_to("sign in", user_teacher_auth_omniauth_authorize_path, secondary: true, class: "sign-in-button", form_class: "sign-in-button") %>
```
jebw: See comment in CSS above but I strongly think we shouldn't have a button inline in a paragraph

config/initializers/devise.rb
```diff
@@ -13,7 +13,7 @@
   oidc_domain = ENV["TRA_OIDC_DOMAIN"].presence
 
   config.omniauth :tra_openid_connect,
-                  allow_authorize_params: %i[request_email_updates],
+                  allow_authorize_params: %i[request_email_updates start_now],
```
jebw: Given we're dropping the ability to start via GAI is this needed?

jebw: Should this be on the TeacherAuth section below?

spec/services/users/find_or_create_from_teacher_auth_spec.rb
```diff
@@ -68,7 +68,12 @@
     context "when user's details have updated" do
       it "updates the user" do
         subject
-        expect(user.reload).to have_attributes(email:, full_name: verified_name.join(" "))
+        expect(user.reload).to have_attributes(
+          email:,
+          full_name: verified_name.join(" "),
+          date_of_birth: Date.parse("1990-01-01"),
```
jebw: Can we change TA to not be persisting date of birth - its PII we don't need to hold, don't expose and don't share with anyone else

spec/support/shared_contexts/stub_teacher_auth_oauth.rb
```diff
@@ -45,12 +45,35 @@
...
+      "uid" => user_uid,
+    }
+  end
+
   let(:user_attributes_from_stubbed_callback_response) do
     {
+      "active_alert" => false,
+      "archived_at" => nil,
+      "archived_email" => nil,
+      "date_of_birth" => user_date_of_birth,
```
jebw: Yeah, can we change TA login's to not persist this field

jebw (commented):
Awesome work - big ol' PR - looks great, couple of minor comments and queries around the inline button (which just seems wrong to me?) and the DOB stuff - happy for the DOB stuff to be ignored though and handled in a separate PR if thats more involved to change?

config/initializers/devise.rb
alkesh: actually, it seems to work without any `allow_authorize_params` set.
I'll set it anyway on teacher_auth, just in case it's something that's only restricted on production environments.

app/views/registration_wizard/start.html.erb
```diff
@@ -66,19 +66,30 @@
...
           )
         %>
+
+      <h2 class="govuk-heading-m">Sign in</h2>
+
+      <p class="govuk-body">
+        You can also sign in
+        through GOV.UK One Login to view a previous registration or change your personal details.
+      </p>
+      <%= govuk_button_to("Sign in", user_teacher_auth_omniauth_authorize_path, secondary: true, class: "sign-in-button", form_class: "sign-in-button") %>
```
jebw: `form_class` can go now

config/initializers/devise.rb
```diff
@@ -31,6 +31,7 @@
     teacher_auth_domain = Rails.configuration.x.teacher_auth.domain
 
     config.omniauth :teacher_auth,
+                    allow_authorize_params: %i[request_email_updates start_now],
```
jebw: We're not using TA for email updates and will be moving away from signins for EoI in the future so I don't think `request_email_updates` needs to be in here

jebw (approved):
LGTM - couple of small tweaks 🙌

PR #3347, opened 2026-05-19
[NPQ-3481] Activate dormant TRN requests

jebw:
### Context

Ticket: [NPQ-3481](https://dfedigital.atlassian.net/browse/NPQ-3481)

We need to activate a users request for a TRN when they are accepted onto a course. There is an API provided TRS we can access for this purpose which is authenticated with an access token. To get the access token, we use a refresh token we have stored against the user to fetch a new, valid, set of credentials. We persist the replacement refresh token and and use the new access token to trigger the API.

After successfully activating the API, we _may_ receive a TRN back immediately in the response but may also not receive until later via the webhook.

If the user's application is declined and has no other open applications, we remove the no longer needed tokens.

1. Add a service to refresh the users credentials
2. Add a service to activate the users dormant trn request
3. Add an ActiveJob to trigger fetching a new token and activating the request, and potentially removing the token if it is no longer needed.
4. Schedule the job when a user is accepted onto a course

[NPQ-3481]: https://dfedigital.atlassian.net/browse/NPQ-3481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/jobs/teaching_record_system/allocate_trn_job.rb
+++ b/app/jobs/teaching_record_system/allocate_trn_job.rb
@@ -10,10 +10,10 @@ module TeachingRecordSystem
         return
       end
 
-      new_access, new_refresh = RefreshAccessToken.refresh!(persisted_token.token)
+      new_access, new_refresh = RefreshTokens.refresh!(persisted_token.token)
       persisted_token.update!(token: new_refresh, last_updated_token_at: Time.zone.now)
 
-      new_trn = ActivateTrnAllocation.activate!(new_access)
+      new_trn = ActivateTrnRequest.activate!(new_access)
       if new_trn.present?
         # Received TRN from API response so don't need to wait for webhook to receive TRN
         user.update!(trn: new_trn, trn_verified: true, trn_auto_verified: true)
--- a/app/jobs/teaching_record_system/allocate_trn_job.rb
+++ b/app/jobs/teaching_record_system/allocate_trn_job.rb
@@ -2,16 +2,15 @@ module TeachingRecordSystem
   class AllocateTrnJob < ApplicationJob
     def perform(user_id:)
       user = User.find(user_id)
-      persisted_token = user.oauth_tokens.refresh.first
 
       if user.trn.present?
         # TRN has been allocated already and token no longer needed
-        persisted_token.destroy!
+        user.refresh_token.destroy!
         return
       end
 
-      new_access, new_refresh = RefreshTokens.refresh!(persisted_token.token)
-      persisted_token.update!(token: new_refresh, last_updated_token_at: Time.zone.now)
+      new_access, new_refresh = RefreshTokens.refresh!(user.refresh_token.token)
+      user.refresh_token.store!(new_refresh)
 
       new_trn = ActivateTrnRequest.activate!(new_access)
       if new_trn.present?
@@ -19,7 +18,7 @@ module TeachingRecordSystem
         user.update!(trn: new_trn, trn_verified: true, trn_auto_verified: true)
       end
 
-      persisted_token.destroy!
+      user.refresh_token.destroy!
     end
   end
 end
--- a/app/jobs/teaching_record_system/allocate_trn_job.rb
+++ b/app/jobs/teaching_record_system/allocate_trn_job.rb
@@ -1,0 +1,25 @@
+module TeachingRecordSystem
+  class AllocateTrnJob < ApplicationJob
+    def perform(user_id:)
+      user = User.find(user_id)
+      persisted_token = user.oauth_tokens.refresh.first
+
+      if user.trn.present?
+        # TRN has been allocated already and token no longer needed
+        persisted_token.destroy!
+        return
+      end
+
+      new_access, new_refresh = RefreshAccessToken.refresh!(persisted_token.token)
+      persisted_token.update!(token: new_refresh, last_updated_token_at: Time.zone.now)
+
+      new_trn = ActivateTrnAllocation.activate!(new_access)
+      if new_trn.present?
+        # Received TRN from API response so don't need to wait for webhook to receive TRN
+        user.update!(trn: new_trn, trn_verified: true, trn_auto_verified: true)
+      end
+
+      persisted_token.destroy!
+    end
+  end
+end
--- a/app/services/teaching_record_system/activate_trn_request.rb
+++ b/app/services/teaching_record_system/activate_trn_request.rb
@@ -2,6 +2,7 @@ module TeachingRecordSystem
   class ActivateTrnRequest
     class NoTrnRequestToActivate < RuntimeError; end
 
+    REQUIRED_API_VERSION = "20260416".freeze
     TRN_REQUEST_PATH = "/v3/trn-request".freeze
     ACTIVATE_PATH = "/v3/trn-request/activate".freeze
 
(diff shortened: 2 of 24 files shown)


app/services/teaching_record_system/refresh_tokens.rb
```diff
@@ -0,0 +1,49 @@
...
+        conn.response :raise_error
+        conn.response :logger, Rails.logger if Rails.env.local?
+      end
+    end
+
+    def request_body
+      {
+        grant_type: "refresh_token",
+        refresh_token: @refresh_token,
+        client_id: ENV["TEACHER_AUTH_CLIENT_ID"],
```
alkesh: I get a SonarCloud check failure on this line: "Check failure: Provide a default value or use ENV.fetch() to validate this environment variable access."

app/services/teaching_record_system/refresh_tokens.rb
```diff
@@ -0,0 +1,49 @@
...
+        client_secret: ENV["TEACHER_AUTH_CLIENT_SECRET"],
```
alkesh: I get a SonarCloud check failure on this line: "Check failure: Provide a default value or use ENV.fetch() to validate this environment variable access."

spec/jobs/teaching_record_system/allocate_trn_job_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        expect(TeachingRecordSystem::RefreshTokens).not_to have_received(:refresh!)
+        expect(TeachingRecordSystem::ActivateTrnRequest).not_to have_received(:activate!)
+      end
+    end
+
+    context "with user without TRN" do
+      context "when activate API does return a TRN" do
+        let(:allocated_trn) { "7349349" }
+
+        it "updates user with trn and removes the refresh token" do
```
alkesh:
```suggestion
        it "updates user with TRN and removes the refresh token" do
```

spec/jobs/teaching_record_system/allocate_trn_job_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+          expect { perform_job }
+            .to change { user.reload.trn }.from(nil).to(allocated_trn)
+            .and change(user.oauth_tokens, :count).from(1).to(0)
+        end
+      end
+
+      context "when activate API does not return a TRN" do
+        it "does not change trn but removes the refresh token now activation is triggered" do
```
alkesh:
```suggestion
        it "does not change TRN but removes the refresh token now activation is triggered" do
```

spec/services/teaching_record_system/activate_trn_request_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+        stub_activation
+          .to_return(status: [200, "Success"],
+                     headers:,
+                     body: activation_body.merge("status" => "Pending").to_json)
+      end
+
+      it { is_expected.to be_nil }
+    end
+
+    context "without pending trn request" do
```
alkesh:
```suggestion
    context "without pending TRN request" do
```

spec/services/teaching_record_system/activate_trn_request_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+    context "with valid access token when already activated" do
+      before do
+        stub_activation
+          .to_return(status: [204, "No Content"], headers:, body: activation_body.to_json)
+      end
+
+      it { is_expected.to be_nil }
+    end
+
+    context "with pending trn request" do
```
alkesh:
```suggestion
    context "with pending TRN request" do
```

app/jobs/teaching_record_system/allocate_trn_job.rb
```diff
@@ -0,0 +1,25 @@
...
+        persisted_token.destroy!
+        return
+      end
+
+      new_access, new_refresh = RefreshTokens.refresh!(persisted_token.token)
+      persisted_token.update!(token: new_refresh, last_updated_token_at: Time.zone.now)
+
+      new_trn = ActivateTrnRequest.activate!(new_access)
+      if new_trn.present?
+        # Received TRN from API response so don't need to wait for webhook to receive TRN
```
alkesh, line 17: `new_trn` will also be `nil` if there is no TRN request to activate - does anything need to be done in that scenario?

app/services/teaching_record_system/activate_trn_request.rb
```diff
@@ -0,0 +1,50 @@
...
+      activation.fetch("trn")
+    end
+
+  private
+
+    def trs_api
+      Faraday.new(url: ENV.fetch("TRS_API_URL")) do |conn|
+        conn.request :authorization, "Bearer", @access_token
+        conn.request :json
+        conn.headers["X-Api-Version"] = "20260416"
```
alkesh: may be worth making `20260416` a constant at the top of this file

spec/services/teaching_record_system/activate_trn_request_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+  end
+
+  describe "#trn_request" do
+    subject(:make_api_call) { described_class.new(access_token).trn_request }
+
+    context "with valid access token" do
+      it { is_expected.to eq trn_request_body }
+    end
+
+    context "without dormant trn request" do
```
alkesh:
```suggestion
    context "without dormant TRN request" do
```

app/jobs/teaching_record_system/allocate_trn_job.rb
jebw: It'll raise `NoTrnRequestToActivate` exception if there's not a trn request - it's only nil if there's not a TRN

Does raise the question of whether I've got the right error being rescued in the service class - from memory it was bad request 🤔 will check that one

jebw: API did raise a 400 so swapped the rescue catch a 400 instead

spec/services/teaching_record_system/activate_trn_request_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+    context "with server error" do
+      before do
+        stub_activation
+          .to_return(status: [500, "Internal Server Error"], headers:)
+      end
+
+      it "raises an exception" do
+        expect { make_api_call }.to raise_exception(Faraday::ServerError)
+      end
+    end
```
alkesh, line 165: is there a missing scenario here on `#activate!`?
"when there is no TRN to activate"?
it { is_expected.to be_nil }

spec/services/teaching_record_system/activate_trn_request_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+          .to_return(status: [200, "Success"], headers:, body: activation_body.to_json)
+      end
+
+      it { is_expected.to eq "1234567" }
+    end
+
+    context "with valid access token when already activated" do
+      before do
+        stub_activation
+          .to_return(status: [204, "No Content"], headers:, body: activation_body.to_json)
```
alkesh: no content - should the body be nil here?

jebw: Yep, your right - etag points to the correct content - reality is we don't sent the ETag back to them so shouldn't hit this path but I'll amend

spec/services/teaching_record_system/activate_trn_request_spec.rb
jebw: Turns out this is covered by 118 - 125

alkesh (approved):

PR #3369, opened 2026-06-02
[NPQ-3756] - display change on input

spqr-praetoria:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3756

- Change ordering of display on input field

### DIFF
--- a/app/javascript/institution-picker.js
+++ b/app/javascript/institution-picker.js
@@ -30,24 +30,22 @@ institutionPicker.enhanceSelectElement = (configurationOptions) => {
   configurationOptions.defaultValue = configurationOptions.selectElement.dataset.selectedInstitutionName || ""
   configurationOptions.displayMenu = "overlay"
 
-  configurationOptions.templates = {
-    inputValue: function(object) {
-      if (object === undefined) {
-        return ""
-      } else {
-        return [object.urn, object.name, object.address].filter(n => n).join(' - ')
-      }
-    },
-    suggestion: function(object) {
-      if (object === undefined) {
-        return ""
-      } else {
-        return [object.urn, object.name, object.address].filter(n => n).join(' - ')
-      }
+  const order = configurationOptions.order || ['name', 'address', 'urn']
+
+  const formatObject = function(object) {
+    if (object === undefined) {
+      return ""
+    } else {
+      return order.map(key => object[key]).filter(Boolean).join(' - ')
     }
   }
 
-  const location = configurationOptions.selectElement.dataset.institutionLocation;
+  configurationOptions.templates = {
+    inputValue: formatObject,
+    suggestion: formatObject
+  }
+
+  const location = configurationOptions.selectElement.getAttribute("data-institution-location")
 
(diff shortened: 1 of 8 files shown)


spec/support/helpers/journey_step_helper.rb
```diff
@@ -62,7 +62,7 @@ def choose_a_private_childcare_provider(js:, urn:, name:)
             page.fill_in "private-childcare-provider-picker", with: provider.urn
           end
 
-          expect(page).to have_content("#{provider.urn} - #{provider.name} - #{provider.address_1}, #{provider.town}")
+          expect(page).to have_content("#{provider.name} - #{provider.address_1}, #{provider.town} - #{provider.urn}")
```
alkesh: is this correct? on this step, they are primarily choosing a URN, so it makes sense for it to be the first thing shown.

app/javascript/institution-picker.js
```diff
@@ -35,14 +35,14 @@ institutionPicker.enhanceSelectElement = (configurationOptions) => {
       if (object === undefined) {
         return ""
       } else {
-        return [object.urn, object.name, object.address].filter(n => n).join(' - ')
+        return [object.name, object.address, object.urn].filter(n => n).join(' - ')
```
jebw: My reading of this link is `inputValue` is the hidden value and `suggestion` is the content shown to the user. Based on that I'd expect the `suggestion` to be changing but not the `inputValue` ?

https://www.npmjs.com/package/accessible-autocomplete#user-content-templates-default-undefined

jebw (approved):
I think what the tickets expected is a little ambiguous

Code seems fine _assuming it still works_ in all situations but probably worth getting some feedback from the product review step first?

app/javascript/institution-picker.js
spqr-praetoria: I double-checked the accessible-autocomplete docs. inputValue is actually the text that gets put into the visible input field once a suggestion is selected (not a hidden value), and suggestion is what's rendered in the dropdown list.

spec/features/journeys/happy_paths/choose_a_school_spec.rb
alkesh, line 1: this change also impacts the `choose-a-childcare-provider` step - there should be a test in that feature file too.

app/javascript/institution-picker.js
```diff
@@ -30,23 +30,21 @@ institutionPicker.enhanceSelectElement = (configurationOptions) => {
...
-      } else {
-        return [object.urn, object.name, object.address].filter(n => n).join(' - ')
-      }
+  const order = configurationOptions.order || ['name', 'address', 'urn']
+
+  const formatObject = function(object) {
+    if (object === undefined) {
+      return ""
+    } else {
+      return order.map(key => object[key]).filter(n => n).join(' - ')
```
alkesh:
```suggestion
      return order.map(key => object[key]).filter(Boolean).join(' - ')
```
as per the SonarCloud warning

alkesh (approved):

PR #3371, opened 2026-06-03
[NPQ-3777] Skip check for changing cohort funding model on new records

jebw:
### Context

Ticket: [NPQ-3777](https://dfedigital.atlassian.net/browse/NPQ-3777)

1. Skip check for changing funding model on cohorts when creating new cohorts

_NB: The correct fix is to amend the schema to prevent cohort-less applications and remove any existing one but this addresses the bug for now without getting into identifying and fixing bad data_

[NPQ-3777]: https://dfedigital.atlassian.net/browse/NPQ-3777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -72,6 +72,7 @@ private
   end
 
   def changing_funding_cap_with_dependent_applications
+    return if new_record?
     return unless funding_changed? && Application.where(cohort: self).any?
 
     errors.add(:funding, "Cannot change funding when there are existing applications for this cohort")
--- a/spec/models/cohort_spec.rb
+++ b/spec/models/cohort_spec.rb
@@ -115,6 +115,14 @@ RSpec.describe Cohort, type: :model do
         end
       end
     end
+
+    describe "creating new cohort when there are existing cohort-less applications" do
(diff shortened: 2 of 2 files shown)


app/models/cohort.rb
```diff
@@ -72,6 +72,7 @@ def registration_start_date_matches_start_year
   end
 
   def changing_funding_cap_with_dependent_applications
+    return if new_record?
```
alkesh, line 75: Do we need a comment saying this is a temporary fix, with a link to the ticket that will remove it?
It's that, or we need a spec.

jebw: So I don't think this change is temporary _should_ be running for new records any way - they'll definitely have changed because every attribute is new

I can add a spec for now because the schema allows for simulating the problem today _but_ that spec will need to be removed in a future PR for the schema change because at that point we cannot simulate this scenario

I'll add a spec for now

alkesh (approved):

PR #3389, opened 2026-06-05
[NPQ-3526] - Prevent GAI sign in taking over TeacherAuth accounts

spqr-praetoria:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3526

- Raise an error when a user that has signed in via TeacherAuth when trying to sign into the same account via GetAnIdentity
- Prevents the provider and uuid being updated on the teacher_auth user on sign in.

### Test notes
- Change config.x.teacher_auth.enabled = false in config/application.rb
- Change config/environments/development to have 
```
config.consider_all_requests_local = false
  ```
  
Take a user that you have signed in with TA locally, should have provider set at "teacher_auth"

Go to localhost:3000 and try to sign in. Should fail and take you to the home page with an error

### DIFF
--- a/app/services/users/find_or_create_from_provider_data.rb
+++ b/app/services/users/find_or_create_from_provider_data.rb
@@ -2,6 +2,8 @@
 
 module Users
   class FindOrCreateFromProviderData
+    class TeacherAuthAccountExistsError < StandardError; end
+
     def initialize(provider_data:, feature_flag_id:)
       @provider_data = provider_data
       @feature_flag_id = feature_flag_id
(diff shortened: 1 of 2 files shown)


jebw (changes requested):
In my testing this isn't working as intended - I think I've misunderstood our failure flow which also means I may need to revisit how this works in the morning

jebw (approved):
LGTM

On further review - I had misunderstood how errors were handled but I think this is fine - this is a code path which shouldn't be getting hit in theory and we'll know about it if it is

PR #3400, opened 2026-06-10
[NPQ-3782] - update header to match one login

spqr-praetoria:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3782

- Switch for header for OneLogin and GetAnIdentity
- display of signout change
- Different headers for OneLogin and GetAnIdentity flow

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -15,7 +15,11 @@
 
     <a tabindex="0" href="#main-content" class="govuk-skip-link">Skip to main content</a>
 
-    <%= govuk_header %>
+    <% if current_user&.teacher_auth_provider? %>
+      <%= render partial: "layouts/shared/one_login_header" %>
+    <% else %>
+      <%= govuk_header %>
+    <% end %>
 
     <%=
       govuk_service_navigation(
--- a/app/views/layouts/shared/_one_login_header.html.erb
+++ b/app/views/layouts/shared/_one_login_header.html.erb
@@ -1,0 +1,72 @@
+<% one_login_link = Rails.configuration.x.teacher_auth.onelogin_home_uri %>
+<div class="rebranded-cross-service-header" data-module="one-login-header">
+  <div class="rebranded-one-login-header" data-one-login-header-nav>
+    <div class="rebranded-one-login-header__container govuk-width-container">
+      <div class="rebranded-one-login-header__logo">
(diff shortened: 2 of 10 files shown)


app/views/layouts/shared/_one_login_header.html.erb
```diff
@@ -0,0 +1,72 @@
+<% one_login_link = Rails.configuration.x.teacher_auth.onelogin_home_uri.presence || "https://home.account.gov.uk" %>
```
jebw: I think this would be better to set a default when assigning the config value from the env var in `application.rb` 🙏

app/views/layouts/application.html.erb
```diff
@@ -15,7 +15,11 @@
 
     <a tabindex="0" href="#main-content" class="govuk-skip-link">Skip to main content</a>
 
-    <%= govuk_header %>
+    <% if current_user&.teacher_auth_provider? %>
+      <%= render partial: "layouts/shared/one_login_header" %>
+    <% else %>
+      <%= govuk_header %>
```
jebw, line 21: I'm wary about us having a large block of ERB which is effectively externally managed and will eventually diverge from the upstream templates.

I'm wondering if we can just add the missing parts to the existing page header (conditionalised as necessary)

https://govuk-components.x-govuk.org/components/header/#adding-more-content-to-the-header

looks like you add extra content using a block

And the key part seems to be `[data-module]` needs setting on the page header for the OneLogin JS to kick in

https://github.com/govuk-one-login/service-header/blob/main/src/scripts/service-header.js#L73

That can be assigned using the common options in `govuk-components` gem

* https://govuk-components.x-govuk.org/introduction/common-options/#inset-text-with-custom-html-attributes

jebw (commented):
Looks good and seems to work well but AFAICT we don't need to add the big block of ERB in there? We can extend our existing header - if I'm misunderstanding that or there's other reasons why we can't just extend the existing header then ping me and I'll approve - thanks

FYI reason logout link errors on the review app is nothing to do with this PR, its because we've not got logout configured for review apps yet

app/views/layouts/application.html.erb
spqr-praetoria: Leaving as is for now, refactor at a later date.

jebw (approved):
LGTM

PR #3401, opened 2026-06-10
[BAU] Fix flaky test for course ordering

ltello:
### Context

Ticket: BAU

The test "orders courses by name in ascending order" was comparing the
database order with Ruby Array#sort. They not use the same order
(collation), so the test fail sometimes. Now it compare with
Course.order(name: :asc), so it use the same order as the database.


jebw (approved):
LGTM 👍

PR #3405, opened 2026-06-11
[NPQ-3799] fix CSP violations

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3799

* the teacher auth authorize path redirects twice before showing the final sign-in-or-create page - I've added these additional paths to the CSP rule (both pre-prod and production domains)
* updated the secure headers gem - not required for this fix, but may as well

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -13,7 +13,15 @@ SecureHeaders::Configuration.default do |config|
   google_analytics = %w[*.google-analytics.com *.analytics.google.com *.googletagmanager.com tagmanager.google.com *.googleusercontent.com *.gstatic.com]
   tracking_pixels = %w[www.facebook.com px.ads.linkedin.com]
   identity_domain = [ENV["TRA_OIDC_DOMAIN"]]
-  teacher_auth_domain = [ENV["TEACHER_AUTH_DOMAIN"]].compact
+
+  teacher_auth_domains = [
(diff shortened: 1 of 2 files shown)


config/initializers/secure_headers.rb
```diff
@@ -13,7 +13,15 @@
...
   tracking_pixels = %w[www.facebook.com px.ads.linkedin.com]
   identity_domain = [ENV["TRA_OIDC_DOMAIN"]]
-  teacher_auth_domain = [ENV["TEACHER_AUTH_DOMAIN"]].compact
+
+  teacher_auth_domain = [
+    ENV["TEACHER_AUTH_DOMAIN"],
+    "oidc.integration.account.gov.uk",
+    "oidc.account.gov.uk",
+    "signin.integration.account.gov.uk",
+    "signin.account.gov.uk",
```
alkesh, line 22: I've added these here, rather than adding extra ENV vars that are only used for these CSP rules.
wdyt?

jebw: Just been debating this one - I think this is ok - we should be able to trust those domains, and its us handing across to TA, which we trust - which is in turn handing across to OneLogin - whilst one is not a production environment, it is their integration environment so presumably running the same code as production (as opposed to a beta/staging setup)

jebw (approved):
LGTM 🙌 very pleased to see this go in

PR #3407, opened 2026-06-11
[NPQ-3803] TRS webhooks - take no action if user not found

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3803

If TRS webhook messages are received, that either have a blank user UID, or have a UID that does not exist in NPQ,
we should just do nothing, and mark the webhook message as processed.


jebw (approved):
LGTM

PR #3384, opened 2026-06-04
[NPQ-3677] set selected workplace on autocomplete input

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3677

### Changes proposed in this pull request

set the workplace on the autocomplete text field

PR #3410, opened 2026-06-11
[BAU] use dataset instead of getAttribute

alkesh:
### Context

BAU

As highlighted by SonarQube, we should be using `dataset` rather than `getAttribute` in our javascript.

This change was originally in this PR: https://github.com/DFE-Digital/npq-registration/pull/3384,
but got accidentally lost in this PR: https://github.com/DFE-Digital/npq-registration/pull/3369


jebw (approved):
LGTM

PR #3413, opened 2026-06-15
[NPQ-3816] normalize closed registration emails records

jebw:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3816

1. Convert column to CI text to allow case insensitive lookup
2. Normalize ClosedRegistrationUser emails on save
3. Add spec for checking mismatched case

### DIFF
--- a/docs/content_editors/images/erd.mmd
+++ b/docs/content_editors/images/erd.mmd
@@ -1,0 +1,394 @@
+erDiagram
+	direction LR
+	APIToken {
+		string hashed_token
+		datetime last_used_at
+		integer lead_provider_id FK
+		enum scope
+	}
+	Adjustment {
+		decimal amount
+		string description
+		integer statement_id FK
+	}
+	Application {
+		string DEPRECATED_itt_provider
+		text DEPRECATED_private_childcare_provider_urn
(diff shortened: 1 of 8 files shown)


spec/models/user_spec.rb
```diff
@@ -314,6 +314,18 @@
       end
     end
 
+    context "with different case email addresses" do
+      before { ClosedRegistrationUser.create!(email: email.downcase) }
```
ltello: Do we need `.downcase` here to prove the normalisation works?
Maybe we can remove it and as the user email will be downcased at creation time, the test proves that `ClosedRegistrationUser` email normalisation actually downcases the value so the match still happens on `user.set_closed_registration_feature_flag`

jebw: Good spot 👍

docs/content_editors/images/erd.mmd
alkesh, line 1: nice.
Do we want to replace the `erd.png` with this?
or maybe link to this from `docs/data_model.md`?

jebw: Good thought - will update

alkesh (approved):

PR #3414, opened 2026-06-15
[NPQ-3787] - Change webhooks to use blocklist

spqr-praetoria:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3787

- Switch from using an allowlist to a blocklist.
- Unexpected messages types raise an exception so we get notified.
- Message types on our ignore list silently fail in the same way we currently do (Setting unhandled_message_type)


alkesh:
there's a SonarQube issue on this

jebw (approved):
LGTM 🚀

PR #3415, opened 2026-06-15
[NPQ-3817] prompt teacher auth users to login in via One Login when they try and log in via DfE Identity

alkesh:
### Context

Ticket: https://dfedigital.atlassian.net/browse/NPQ-3817

* if a user already has a teacher auth account, and they try and sign in via DfE Identity on the /registration_closed page, then show an error message, and show a "Sign in to your One Login account" button instead of the DfE Identity one.


jebw (approved):
LGTM

PR #3419, opened 2026-06-17
[BAU] Reduce notifications about null user emails

jebw:
### Context

Ticket: BAU

We only want to know about the nulling of email addresses for users who've applied. We also have users signing up via GAI to register for Expression of Interest who are unlikely to have a verified trn, will be applying via a different auth system (Teacher Auth) so will not be matched but its not a concern in this scenario and not visible on the API.

1. Only notify sentry if the user has any applications

### DIFF
--- a/app/services/users/archiver.rb
+++ b/app/services/users/archiver.rb
@@ -17,7 +17,10 @@ module Users
       user.provider = nil
       user.save!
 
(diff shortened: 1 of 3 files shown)


spqr-praetoria (approved):
LGTM

PR #3423, opened 2026-06-17
[NPQ-3766] process person.deactivated webhook messages

alkesh:
### Context

Ticket: [NPQ-3766](https://dfedigital.atlassian.net/browse/NPQ-3766)

process `person.deactivated` webhook messages

1. new processor `TeachingRecordSystem::Webhooks::PersonDeactivatedProcessor`
2. new scopes on `User` : `not_archived` and `with_trn`
3. added index on users `archived_at` column

[NPQ-3766]: https://dfedigital.atlassian.net/browse/NPQ-3766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/user.rb
```diff
@@ -126,7 +127,7 @@ def unsubscribe_from_email_updates
   end
 
   def archived?
-    archived_email.present?
+    archived_at.present?
```
jebw, line 133: Do we have any users in the prod data who have `archived_email` set but not `archived_at` so would be 'unarchived' by this changed?

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -95,7 +95,8 @@ def persist_token(user, provider_data)
     def verified_trn_matching_users
       @verified_trn_matching_users ||=
         User
-          .where(trn:, trn_verified: true, archived_at: nil)
+          .not_archived
+          .where(trn:, trn_verified: true)
```
jebw: You could use your 'with_trn' scope?

spec/factories/users.rb
```diff
@@ -26,6 +26,7 @@
 
       uid { teacher_auth_uid }
       provider { "teacher_auth" }
+      trn_verified { trn.present? }
```
jebw, line 29: 🙌

app/services/teaching_record_system/webhooks/person_deactivated_processor.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+
+      def update_deactivated_users_trn!
+        users_matching_deactivated_person.update!(trn: merged_with_trn)
+      end
+
+      def handle_merged_with_users
+        if user_to_keep.archived? && only_archived_teacher_auth_users_matching_deactivated_person?
+          update_deactivated_users_trn!
+          merge_and_archive_users(users_to_merge: user_to_keep, merge_into: users_matching_deactivated_person.first)
```
jebw: I'm not quite following this line and why it is merging the opposite of direction to what would seem logical _but_ I'd've maybe anticipated a simpler logic for the whole class, eg

```ruby
def call
  return unless merge_into

  update_trn(merge_into)
  merge_and_archive_users(users_to_merge:, merge_into:)
end

private

def users_to_merge
  all_matching_users[1..]
end

def merge_into
  all_matching_users[0]
end

def all_matching_users
  @all_matching_users ||= User
    .with_trn([deactivated_trn, merged_with_trn])
    .order(archived_at: :desc, provider: :desc, updated_at: :desc, id: :desc)
    .to_a
end
```

jebw (commented):
I'm thinking the merge logic can be much simpler here

I think its 

1) Which account do we have which we think is most relevant (thats the fragment from the ticket)
2) Update that account
3) Merge any other matching accounts into (1)?

app/models/user.rb
alkesh: nope - all archived users have both archived_email and archived_at set in production.

app/services/teaching_record_system/webhooks/person_deactivated_processor.rb
alkesh: Now simplified using the structure you suggested.
There were a couple of nuances:
- there are a couple of scenarios where we need to update multiple users TRNs - the new `update_trns!` method handles this.
- There is one scenario where the above logic does not work (multiple GAI users match the deactivated person, and one archived teacher auth user matches the merged-with person) - in this scenario we are keeping an archived user instead of an unarchived one. I've updated the `merge_into` method to cope with this scenario.

app/services/teaching_record_system/webhooks/person_deactivated_processor.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      def merge_into
+        matching_users_archived_teacher_auth_users = all_matching_users.archived.with_teacher_auth
+        matching_users_gai_non_archived_users = all_matching_users.not_archived.with_get_an_identity_id
+
+        if matching_users_archived_teacher_auth_users.one? && matching_users_gai_non_archived_users.any?
+          matching_users_archived_teacher_auth_users.first
+        else
+          all_matching_users.first
+        end
```
jebw: AFAICT this is effectively overriding the priorisation in `all_matching_users` - since everything but 'merge_into' will get merged it seems like it would be easier to change the priorisation to

```
.order(provider: :asc, archived_at: :desc, updated_at: :desc, id: :desc)
```

at which point this becomes `all_matching_users[0]` - i use the index instead of `#first` since `#first` will trigger a query if the scope is already loaded but another call will be needed later to load the rest of the data. An index will load all of the matching users in a single query and just return the first of them.

app/services/teaching_record_system/webhooks/person_deactivated_processor.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+
+      def all_matching_users
+        @all_matching_users ||= User
+          .with_trn([deactivated_trn, merged_with_trn])
+          .order(archived_at: :desc, provider: :asc, updated_at: :desc, id: :desc)
```
jebw: If you call `#to_a` then `@all_matching_users` memoizes data instead of a scope and you can further divide that list into 2 pots

`[0]` - the 'favoured' user record to merge into
`[1..]` - the rest to be merged into the favoured user record

app/services/teaching_record_system/webhooks/person_deactivated_processor.rb
```diff
@@ -0,0 +1,72 @@
...
+        webhook_message.message["mergedWithPerson"] && webhook_message.message["mergedWithPerson"]["trn"]
+      end
+
+      def users_matching_merged_with_person
+        @users_matching_merged_with_person ||= User.with_trn(merged_with_trn)
+      end
+
+      def update_trns!
+        if users_matching_merged_with_person.any?
+          merge_into.update!(trn: merged_with_trn)
```
jebw: Unless I'm missing something, this looks for users who have a given TRN, then changes them to have that given TRN (which they already have)?

jebw (commented):
I feel this still has a lot of complexity which is more than we need - TRS have told us 2 TRNs are being collapsed together. We need to find any matching users and collapse them together making sure the user which remains has the correct TRN?

AFAICT this needs

1) Fetch an array of all the relevant users, ordering in a manner which places the most favoured at the start - this may be 0 or more users
2) Update that first users TRN to our 'merged into' TRN -> since that is ultimately what TRS thinks the users TRN should be
3) Merge every user other user into the first one

I think the change to favour TeacherAuth users over whether a GAI account is archived or not makes a lot of sense.

jebw (approved):
LGTM - thanks for working through all this 🙏

PR #3424, opened 2026-06-17
[NPQ-3608] Swap webpack/babel for esbuild to bundle JS

ltello:
### Context

Ticket: [NPQ-3608](https://dfedigital.atlassian.net/browse/NPQ-3608)

We bundle JS with webpack and babel. That needs a lot of JS packages (webpack, webpack-cli, babel-loader...) that slows the builds and require frequent dependency versions upgrades.

This PR swaps that for esbuild. esbuild does the bundling, transpiling and minifying, so it is faster and we only keep one package for the job. The Rails side does not change: jsbundling-rails still runs the `build` script, and the CSS is still built by sass.

1. Add `esbuild.config.js`.
2. Point the `package.json` `build` script to esbuild and drop the webpack/babel/file-loader packages.
3. Delete `webpack.config.js`, `babel.config.js` and `postcss.config.js`.
4. Declare `lodash.debounce`, which `app/javascript/institution-picker.js` imports directly but was only there before as a dependency of babel.
5. `Dockerfile`: remove the `openssl-legacy-provider` workaround that only webpack needed.
6. Update docs to mention esbuild.

Now that we don't use webpack anymore, nothing prevents us from replacing `swagger-ui-dist` with `swagger-ui`.

7. Swap `swagger-ui-dist` for `swagger-ui` in `package.json`.
8. `swagger-ui.js`: import from `swagger-ui`.
9. `swagger-ui.scss`: import the CSS from the `swagger-ui` package.
10. Force React v18 in `package.json`. The new swagger-ui still calls ReactDOM.render. React 19 has removed it, so the api documentation page was blank. We make React stay at v18 to fix it. Only swagger-ui uses React, so this is safe.

[NPQ-3608]: https://dfedigital.atlassian.net/browse/NPQ-3608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/swagger.md
```diff
@@ -48,6 +48,6 @@ Once updates are made to the schemas, you will need to run the `rswag:specs:swag
 
 We currently have a simple controller that renders out a custom layout to pull in a JS pack containing `swagger-ui-dist`. This will boot Swagger UI using the version of the yaml file in the path, so `/api/docs/v3` will render `/api/docs/v3/swagger.yaml`.
 
-We have to use `swagger-ui-dist` instead of `swagger-ui` as the latter didn't transpile properly with the outdated version of webpacker/webpack we used to have.
+We have to use `swagger-ui-dist` instead of `swagger-ui` as the latter didn't transpile properly with the outdated version of webpacker/webpack we used to have (we now bundle assets with esbuild).
```
alkesh: does this mean we can now use `swagger-ui`? and delete this whole line?

ltello: Good question. I have added a couple of commits to actually replace `swagger-ui-dist` with `swagger-ui`. So now, the answer to your question is yes!

jebw (approved):
LGTM - so pleased to move away from webpack 🚀

PR #3425, opened 2026-06-17
[NPQ-3823] user clash bug

alkesh:
### Context

Ticket: [NPQ-3823](https://dfedigital.atlassian.net/browse/NPQ-3823)

This bug occurs when a use has a GAI user account (_account1_),
which is then converted to a Teacher Auth account by the login process.
After that, a newer GAI account is created with the same TRN (_account2_).

When the user tries to log in via One Login in this scenario, the code cannot update the details on _account2_ because they clash with _account1_.

1. check for teacher auth accounts with the TRN first, before checking GAI accounts.
2. do not show the 'sign up for email updates' content or button for one login accounts, as this only works for DfE Identity users.

[NPQ-3823]: https://dfedigital.atlassian.net/browse/NPQ-3823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/users/find_or_create_from_provider_data.rb
+++ b/app/services/users/find_or_create_from_provider_data.rb
@@ -28,6 +28,16 @@ module Users
                 "GAI sign-in matched a Teacher Auth user by email (user_id=#{user.id}); refusing to take over the account"
         end
 
+        teacher_auth_user_with_same_trn = User.find_by(
+          trn: provider_data.extra.raw_info.trn,
+          trn_verified: true,
+          provider: Omniauth::Strategies::TeacherAuth::NAME,
+        )
+        if teacher_auth_user_with_same_trn
+          raise TeacherAuthAccountExistsError,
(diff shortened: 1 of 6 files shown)


jebw (commented):
Thanks for debugging this - that makes more sense now - I'm not convinced this is the right solution though. 

I think this scenario is something we'd not originally expected to occur because we weren't expecting both auth systems concurrently.

I think its fine to keep the 'migrate the second GAI account' in there _but_ I also think we need to change the GAI sign in logic to prevent creation of a GAI account if it would have the same TRN as a TeacherAuth account. It should be sufficient to just through the same exception as gets thrown when a user tries to sign back in to an account via GAI after its been converted to TA.

jebw (approved):
LGTM

PR #3427, opened 2026-06-18
[NPQ-3745] archived accounts signing in to teacher auth

alkesh:
### Context

Ticket: [NPQ-3745](https://dfedigital.atlassian.net/browse/NPQ-3745)

allow users that have been archived, but have a verified TRN, to log in and have their account un-archived.

1. match on archived as well as unarchived users for verified TRNs
2. unarchive archived users if they are the matched user

### Notes
based off the `NPQ-3766-TRNs-merged-webhook` branch, because that introduces the `allow_archived_users` option on `Users::MergeAndArchive`.

[NPQ-3745]: https://dfedigital.atlassian.net/browse/NPQ-3745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spqr-praetoria (commented):
LGTM

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -95,9 +96,8 @@ def persist_token(user, provider_data)
     def verified_trn_matching_users
       @verified_trn_matching_users ||=
         User
-          .not_archived
           .with_trn(trn)
-          .order(updated_at: :desc)
+          .order(updated_at: :desc, archived_at: :desc)
```
jebw: Hmm - I think this needs to be a bit more complex - still focus on most recently updated at but prioritise those which are not archived

```ruby
.order("users.archived_at IS NOT NULL")
.order(updated_at: :desc, archived_at: :desc)
```

Or possibly just keep it simple and flip the order conditions around
```ruby
.order(archived_at: :desc, updated_at: :desc)
```

app/services/users/find_or_create_from_teacher_auth.rb
jebw: User needs to be unarchived here

If they've not matched on TRN, and they've had their email address blanked (because of re-use) then we should unarchived them

jebw (commented):
Looks good but I think it still needs some tweaks

app/services/users/find_or_create_from_teacher_auth.rb
jebw: These lines should be happening inside of the transaction otherwise we can end up in an inconsistent state

jebw (commented):
Looks good but we've got multiple changes happening for TRN matches outside of a transaction which could leave things in an inconsistent state

jebw (approved):
LGTM

PR #3428, opened 2026-06-18
[NPQ-3820] keep GAI signed in

spqr-praetoria:
### Context

Ticket: [NPQ-3820](https://dfedigital.atlassian.net/browse/NPQ-3820)

Keep the user signed in

- Prevent CSRF notifications
- Keep user signed in from GAI

[NPQ-3820]: https://dfedigital.atlassian.net/browse/NPQ-3820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/omniauth_controller.rb
+++ b/app/controllers/omniauth_controller.rb
@@ -1,4 +1,6 @@
 class OmniauthController < Devise::OmniauthCallbacksController
+  OMNIAUTH_ERROR_STRATEGY_KEY = "omniauth.error.strategy".freeze
+
   skip_before_action :verify_authenticity_token, only: [
     Omniauth::Strategies::TraOpenidConnect::NAME, Omniauth::Strategies::TeacherAuth::NAME
   ]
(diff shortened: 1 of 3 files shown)


app/controllers/omniauth_controller.rb
```diff
@@ -133,7 +143,24 @@ def after_sign_in_path_for(user)
...
+    request.env[OMNIAUTH_ERROR_STRATEGY_KEY]&.name&.to_s
+  end
+
+  def signed_in_with_attempted_provider?
+    attempted_strategy_name.present? && current_user.provider == attempted_strategy_name
+  end
+
+  def post_gai_failure_path
+    if Flipper.enabled?(Feature::CLOSED_REGISTRATION_ENABLED) &&
+        !Flipper.enabled?(Feature::REGISTRATION_OPEN, current_user)
```
jebw: I don't think this needs to feed in `current_user` ?

If reg is disabled
And early reg is enabled
And the user is trying to sign in via TA
Then they must have come from the closed reg page 

so we should send them back there irrespective of whether they were on the list of those who can sign up on that exception page

spec/requests/omniauth_failure_spec.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Omniauth failure", type: :request do
```
jebw: These specs look really good but they can go in the regular `spec/requests/omniauth_spec` ?

One thing i did spot is we're getting the omniauth errors in the spec output - is there anything we can do to turn that off in the test env?

jebw (commented):
Looks great, couple of minor tweaks and this can get merged up on monday 🙌

spec/requests/omniauth_failure_spec.rb
spqr-praetoria: Will take a look, didn't notice that myself

jebw (approved):
LGTM 👍 

Just needs the lint failure fixing - if you fix that I can merge it on monday - bit late for a merge today, thanks

PR #3433, opened 2026-06-19
[NPQ-3213] Add a session id to log lines and Sentry events

ltello:
### Context

Ticket: [NPQ-3213](https://dfedigital.atlassian.net/browse/NPQ-3213)

When there is an exception it is very hard to follow the steps a user took in the
registration journey. Our logs and Sentry reports are heavily redacted and there is no
single id shared by all the requests of one user. This adds a session id so we can go
from a Sentry error to all the Rails logs of that session: read the session id in Sentry
and search Logit for it.

1. Add `log_session_id` in `ApplicationController`, a generated UUID kept in the session.
2. Tag every log line of the request with this `session_id` (Semantic Logger named tag),
   so it is a searchable field in the JSON logs (Logit).
3. Send the same `session_id` to Sentry as a tag, so any exception in the request has it.
4. Add a request spec for this behaviour.

### Notes

- Only applicants and admin requests, not the API ones.
- The id is a generated UUID, not the real session value, so it is safe to expose in logs.

[NPQ-3213]: https://dfedigital.atlassian.net/browse/NPQ-3213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,11 +1,12 @@
 class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
+  around_action :add_session_id_to_logs
   around_action :set_time_zone
   before_action :clear_null_user_sessions
   before_action :set_cache_headers
   before_action :authenticate_user!
-  before_action :set_sentry_user
+  before_action :set_sentry_context
   before_action :initialize_store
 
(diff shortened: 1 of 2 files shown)


jebw (approved):
LGTM 🚀

PR #3434, opened 2026-06-22
[NPQ-3837] Add double click prevention to omniauth buttons

ltello:
### Context

Ticket: [NPQ-3837](https://dfedigital.atlassian.net/browse/NPQ-3837)

We still see some omniauth CSRF errors now and then. The omniauth_openid_connect gem keeps a secret in the session, sends it through the OIDC provider, and checks it when the user comes back. One likely cause is a double click on a sign in button: the second click can overwrite the token in the session, so the check on return fails.

This PR adds the standard GOV.UK double click prevention to all the omniauth buttons.

1. Added `data-prevent-double-click="true"` to every omniauth button on the site.
2. The buttons changed are: the Get an Identity and Teacher Auth redirect button partials, the start page "Sign in" button, the registration closed and temporarily closed pages, and the email updates unsubscribed page.
3. Added a test on the start page to check the buttons have the attribute.

[NPQ-3837]: https://dfedigital.atlassian.net/browse/NPQ-3837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/email_updates/unsubscribed.html.erb
+++ b/app/views/email_updates/unsubscribed.html.erb
@@ -22,7 +22,8 @@
         render GovukComponent::ContinueButtonComponent.new(
           text: "Re-subscribe",
           href: user_tra_openid_connect_omniauth_authorize_path(request_email_updates: "true"),
-          as_button: true
+          as_button: true,
+          html_attributes: { data: { "prevent-double-click": true } }
         )
       %>
     <% end %>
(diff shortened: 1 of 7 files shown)


jebw (approved):
LGTM

PR #3436, opened 2026-06-23
[NPQ-3847] Fix user email clash on sign in

ltello:
### Context

Ticket: [NPQ-3847](https://dfedigital.atlassian.net/browse/NPQ-3847)

When a user signs in with Teacher Auth and we match them by their verified TRN, we update that user with the email from Teacher Auth.
If a different account already had that email, the save failed because email
must be unique. 
The other two paths (match by UID, and create a new user) already free the email first, so the verified TRN path was the only one that could crash.

1. In the verified TRN path, merge the other users with the same TRN first. This archives them and frees the email if one of them had it.
2. Then blank the email on any other account that still has it, before updating the matched user and taking over the account.
3. Add specs for the verified TRN match when a different account holds the incoming email, and when one of the same TRN duplicates holds it.

[NPQ-3847]: https://dfedigital.atlassian.net/browse/NPQ-3847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/users/find_or_create_from_teacher_auth_spec.rb
+++ b/spec/services/users/find_or_create_from_teacher_auth_spec.rb
@@ -278,7 +278,7 @@ RSpec.describe Users::FindOrCreateFromTeacherAuth do
     end
 
     context "when one of the matching users by TRN already holds the incoming email" do
-      let(:older_user) { create(:user, trn:, trn_verified: true, updated_at: 2.days.ago, email:) }
+      let!(:older_user) { create(:user, trn:, trn_verified: true, updated_at: 2.days.ago, email:) }
 
       it "frees the email by merging and sets it on the kept user" do
         expect { subject }.not_to raise_error
--- a/spec/services/users/find_or_create_from_teacher_auth_spec.rb
+++ b/spec/services/users/find_or_create_from_teacher_auth_spec.rb
@@ -152,6 +152,17 @@ RSpec.describe Users::FindOrCreateFromTeacherAuth do
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/services/users/find_or_create_from_teacher_auth_spec.rb
```diff
@@ -266,6 +277,16 @@
       end
     end
 
+    context "when one of the matching users by TRN already holds the incoming email" do
+      let(:older_user) { create(:user, trn:, trn_verified: true, updated_at: 2.days.ago, email:) }
+
+      it "frees the email by merging and sets it on the kept user" do
+        expect { subject }.not_to raise_error
+        expect(older_user.reload).to be_archived
```
jebw, line 285: My reading of this is `older_user` is only created _after_ the service is called in the line above? Should this either user `before` or `let!` ?

jebw (commented):
LGTM - simple change and I can readily follow the logic. I've a query on one of the spec

spec/services/users/find_or_create_from_teacher_auth_spec.rb
ltello: Well, `older_user` is created on line 227 `travel_to(2.days.ago) { application }`  before running `subject`.
But it is not easy to detect. So great, a `let!` will probably make it fully evident. I'll update it. Thanks!

jebw (approved):
LGTM :rocket:

PR #3441, opened 2026-06-24
[BAU] do not use constantize on a param

alkesh:
### Context

BAU

avoid using `constantize` on a parameter

1. use a mapping hash instead of `constantize`


jebw (approved):
LGTM

PR #3442, opened 2026-06-24
[BAU] use `system` correctly so it sanitizes shell inputs

alkesh:
### Context

use `system` correctly so it sanitises its inputs

1. use `system` correctly, with multiple parameters
2. fix test to test system call is made correctly - stubbing the call the `system` for speed and test reliability


jebw (approved):
LGTM

PR #3446, opened 2026-06-25
[NPQ-3786] Prevent blank TRNs from being verified

ltello:
### Context

Ticket: [NPQ-3786](https://dfedigital.atlassian.net/browse/NPQ-3786)

It should not be possible to have a user with a TRN that is blank but marked as verified. We add a model validation and a database constraint to stop this.

1. Add a model validation to `User` so `trn_verified` cannot be `true` when the TRN is blank.
2. Add a database check constraint for the same rule, in two migrations: the first adds it with `validate: false`, the second validates it.
3. Before validating the constraint, the second migration cleans any rows that would break it: it sets `trn_verified` to `false` where the TRN is null or empty.
4. Fix `Users::FindOrCreateFromTeacherAuth` which set `trn_verified` to `true` even when the incoming TRN was blank. Now it only marks the TRN as verified when the TRN is present.
5. Update specs, and remove an old test that is no longer possible now the constraint exists.

### Notes

The data cleanup runs as part of the migration, so all environments are fixed automatically on deploy.

[NPQ-3786]: https://dfedigital.atlassian.net/browse/NPQ-3786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/users/find_or_create_from_teacher_auth.rb
+++ b/app/services/users/find_or_create_from_teacher_auth.rb
@@ -45,7 +45,7 @@ module Users
             always_updated_attributes.merge(
               email:,
               trn:,
-              trn_verified: true,
+              trn_verified: trn.present?,
             ),
           )
           persist_token(user_matched_using_uid, provider_data)
--- a/db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
+++ b/db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
@@ -1,17 +1,7 @@
 class ValidateTrnPresentWhenVerifiedConstraintOnUsers < ActiveRecord::Migration[8.0]
-  def up
-    # Clean rows that would fail the constraint before validating it.
-    safety_assured do
-      execute(<<~SQL)
-        UPDATE users
-        SET trn_verified = false
-        WHERE trn_verified = true
-          AND (trn IS NULL OR trn = '')
-      SQL
-    end
+  disable_ddl_transaction!
 
+  def change
     validate_check_constraint :users, name: "users_trn_present_when_verified"
   end
-
-  def down; end
 end
--- a/db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
+++ b/db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
@@ -1,0 +1,17 @@
+class ValidateTrnPresentWhenVerifiedConstraintOnUsers < ActiveRecord::Migration[8.0]
+  def up
+    # Clean rows that would fail the constraint before validating it.
+    safety_assured do
+      execute(<<~SQL)
+        UPDATE users
+        SET trn_verified = false
+        WHERE trn_verified = true
+          AND (trn IS NULL OR trn = '')
+      SQL
+    end
+
+    validate_check_constraint :users, name: "users_trn_present_when_verified"
+  end
+
+  def down; end
+end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -80,7 +80,7 @@ RSpec.describe User do
       expect(user).not_to be_valid
     end
 
-    it "allows a verified trn when the trn is present" do
+    it "allows trn_verified when the TRN is present" do
       user = build(:user, :with_verified_trn, trn: "1234567")
       expect(user).to be_valid
     end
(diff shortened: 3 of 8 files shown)


spec/models/user_spec.rb
```diff
@@ -79,6 +79,35 @@
       user = build(:user, email: nil)
       expect(user).not_to be_valid
     end
+
+    it "allows a verified trn when the trn is present" do
```
alkesh:
```suggestion
    it "allows trn_verified when the TRN is present" do
```

spec/models/user_spec.rb
```diff
@@ -79,6 +79,35 @@
...
+      user = build(:user, :with_verified_trn, trn: "1234567")
+      expect(user).to be_valid
+    end
+
+    it "does not allow trn_verified when the trn is nil" do
```
alkesh:
```suggestion
    it "does not allow trn_verified when the TRN is nil" do
```

spec/models/user_spec.rb
```diff
@@ -79,6 +79,35 @@
...
+      user = build(:user, :with_verified_trn, trn: nil)
+      expect(user).not_to be_valid
+      expect(user.errors[:trn_verified]).to include("TRN cannot be marked as verified when it is blank")
+    end
+
+    it "does not allow trn_verified when the trn is blank" do
```
alkesh:
```suggestion
    it "does not allow trn_verified when the TRN is blank" do
```

spec/models/user_spec.rb
```diff
@@ -79,6 +79,35 @@
...
+      user = build(:user, :with_verified_trn, trn: "")
+      expect(user).not_to be_valid
+    end
+
+    it "allows a blank trn when it is not verified" do
```
alkesh:
```suggestion
    it "allows a blank TRN when it is not verified" do
```

spec/models/user_spec.rb
```diff
@@ -79,6 +79,35 @@
...
+      user = build(:user, trn: nil, trn_verified: false)
+      expect(user).to be_valid
+    end
+  end
+
+  describe "database constraints" do
+    it "rejects a blank trn marked as verified" do
```
alkesh:
```suggestion
    it "rejects a blank TRN marked as verified" do
```

app/services/users/find_or_create_from_teacher_auth.rb
```diff
@@ -44,7 +44,7 @@ def call
             always_updated_attributes.merge(
               email:,
               trn:,
-              trn_verified: true,
+              trn_verified: trn.present?,
```
alkesh, line 48: 🙌🏽

alkesh (approved):

db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
```diff
@@ -0,0 +1,17 @@
...
+  def up
+    # Clean rows that would fail the constraint before validating it.
+    safety_assured do
+      execute(<<~SQL)
+        UPDATE users
+        SET trn_verified = false
+        WHERE trn_verified = true
+          AND (trn IS NULL OR trn = '')
+      SQL
+    end
```
jebw: I'd move this SQL to the earlier migration - logic being we want validating the check constraint to occur concurrently (otherwise it'll need to walk the entire table when running this migration)

db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    validate_check_constraint :users, name: "users_trn_present_when_verified"
```
jebw, line 5: Which will mean this migration only has the check constraint validation in it, so it can safely do the check concurrently (which requires `disable_ddl_transactions`

jebw (commented):
LGTM but I think reordering the migrations slightly would improve deployment?

db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
ltello: Good point! I'll make the change!

ltello: done

db/migrate/20260624114120_validate_trn_present_when_verified_constraint_on_users.rb
ltello: done

PR #3447, opened 2026-06-25
[NPQ-3450] schedule cron jobs on deployment

alkesh:
### Context

Ticket: [NPQ-3450](https://dfedigital.atlassian.net/browse/NPQ-3450)

[Brief description of objective driving this change]

1. if there is a new `CronJob` class that has not been scheduled - schedule it
2. if there is a delayed cron job that refers to a `CronJob` class that has been deleted - delete the delayed cron jobs
3. if there is a delayed cron job whose cron expression does not match that defined on the `CronJob` class, delete and reschedule the job
4. use an advisory lock to prevent concurrent instances of updating the jobs at the same time
5. add a `production_only` attribute to delayed cron jobs, so only certain jobs are scheduled on staging/sandbox environments
6. review app environments have no cron jobs scheduled, because `db:seed:replant` removes them all
7. run all of this when `db:migrate` is run

[NPQ-3450]: https://dfedigital.atlassian.net/browse/NPQ-3450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/jobs/crons/sentry_healthcheck_job.rb
+++ b/app/jobs/crons/sentry_healthcheck_job.rb
@@ -2,7 +2,9 @@
 class Crons::SentryHealthcheckJob < CronJob
   include Sentry::Cron::MonitorCheckIns
 
-  # run every hour
+  self.production_only = true
+
+  # run every hour at midnight
   self.cron_expression = "0 * * * *"
 
   sentry_monitor_check_ins slug: "sentry-healthcheck"
--- a/app/jobs/crons/sentry_healthcheck_job.rb
+++ b/app/jobs/crons/sentry_healthcheck_job.rb
@@ -4,7 +4,7 @@ class Crons::SentryHealthcheckJob < CronJob
 
   self.production_only = true
 
-  # run every hour at midnight
+  # run every hour at 0 minutes past
   self.cron_expression = "0 * * * *"
 
   sentry_monitor_check_ins slug: "sentry-healthcheck"
--- a/lib/tasks/delayed_cron_job.rake
+++ b/lib/tasks/delayed_cron_job.rake
@@ -41,7 +41,7 @@ private
       cron_job = cron_job_class_constant(cron_job_class)
 
       if cron_job
-        next unless cron_job.cron_expression != delayed_job.cron
+        next if cron_job.cron_expression == delayed_job.cron
 
         logger.info "delayed job found with out-of-date cron expression - rescheduling cron job: #{cron_job_class}"
         delayed_job.destroy!
--- a/lib/tasks/delayed_cron_job.rake
+++ b/lib/tasks/delayed_cron_job.rake
@@ -1,0 +1,79 @@
+class DelayedCronJobRakeTask
+  include Rake::DSL
+
+  def initialize
+    namespace :delayed_cron_job do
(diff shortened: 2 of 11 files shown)


lib/tasks/delayed_cron_job.rake
```diff
@@ -0,0 +1,79 @@
...
+        next unless cron_job.cron_expression != delayed_job.cron
+
+        logger.info "delayed job found with out-of-date cron expression - rescheduling cron job: #{cron_job_class}"
+        delayed_job.destroy!
+        cron_job.schedule
+      else
+        logger.info "delayed job found for deleted cron job - deleting delayed job"
+        delayed_job.destroy!
+      end
+    end
```
ltello, line 59: Should we also check if a production-only job exists already in a non-production environment and delete it?

lib/tasks/delayed_cron_job.rake
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def handle_existing_cron_delayed_jobs
+    cron_delayed_jobs.each do |delayed_job|
+      cron_job_class = cron_job_class(delayed_job)
+      cron_job = cron_job_class_constant(cron_job_class)
+
+      if cron_job
+        next unless cron_job.cron_expression != delayed_job.cron
```
ltello: Maybe `next if cron_job.cron_expression == delayed_job.cron` is easier to read?

app/jobs/crons/sentry_healthcheck_job.rb
```diff
@@ -2,7 +2,9 @@
 class Crons::SentryHealthcheckJob < CronJob
   include Sentry::Cron::MonitorCheckIns
 
-  # run every hour
+  self.production_only = true
+
+  # run every hour at midnight
```
ltello: I think the previous comment was correct, no?

ltello (commented):
It looks very good!. Just some minor comments!

app/jobs/crons/sentry_healthcheck_job.rb
alkesh: ah yes, my mistake - will correct 👍🏽 
it should be: `# run every hour at 0 minutes past` for consistency with the other comments.

lib/tasks/delayed_cron_job.rake
alkesh: good point - I've checked what is running in sandbox/staging already, and there aren't any jobs that need to be unscheduled - but there are some that will get scheduled so both environments are consistent - I've put details in the ticket.

ltello (approved):

lib/tasks/delayed_cron_job.rake
ltello: I have approved @alkesh. 
Do you think there is any risk of having this issue again in the future if it was decided to convert any all-envs cron job to production-only?

alkesh: I think for completeness, this rake task should cope if cron jobs change from all-env to production-only (or vice versa).
I've updated the rake task.

PR #3448, opened 2026-06-25
[NPQ-3857] increase admin OTP code entropy

ltello:
### Context

Ticket: [NPQ-3857](https://dfedigital.atlassian.net/browse/NPQ-3857)

Admin sign in OTP codes were 6 digits, so only 1,000,000 possible codes. We relied on rate limiting to keep them safe, but that is not enough when an attacker controls a large pool of IPs. The codes need a much bigger pool.

1. New OTP value object. OTP.generate makes a fresh code using Crockford's Base32 alphabet (8 chars, so 1 trillion possible codes instead of 1,000,000). OTP.from(code:, expires_at:) wraps a stored one. The object cannot exist in an invalid state: the constructor raises OTP::Invalid unless the code is 8 Crockford chars and the expiry is a valid time value.
2. The OTP now expires 5 minutes from now instead of 10 minutes.
3. matches? normalises the entered code the Crockford way  and compares it with ActiveSupport secure_compare.
4. The sign in code check moved into a custom UserOTPCodeValidator. It only builds an OTP when the user's stored code and expiry are valid, so old or invalid data returns "incorrect" instead of raising.

[NPQ-3857]: https://dfedigital.atlassian.net/browse/NPQ-3857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/otp.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def self.random_code
+    Array.new(CODE_LENGTH) { CROCKFORD_BASE32[SecureRandom.random_number(CROCKFORD_BASE32.size)] }.join
+  end
+  private_class_method :random_code
+
+  def self.valid_code?(code)
+    code.to_s.match?(CODE_PATTERN)
+  end
```
jebw: all these can be moved within `class << self` block

app/services/otp.rb
```diff
@@ -0,0 +1,51 @@
...
+  def self.normalize(code)
+    code.to_s.upcase.tr("OIL", "011")
+  end
+
+  def self.random_code
+    Array.new(CODE_LENGTH) { CROCKFORD_BASE32[SecureRandom.random_number(CROCKFORD_BASE32.size)] }.join
+  end
+  private_class_method :random_code
+
+  def self.valid_code?(code)
```
jebw: validity is ambiguous here - an expired code is invalid according to the validation will pass this - maybe call it `valid_format?` ? Or something more specific to the format of the code?

jebw (approved):
LGTM - couple of minor code formatting suggestions but after those happen I think this can be shipped?

app/services/otp.rb
ltello: done

app/services/otp.rb
ltello: renamed to `valid_code_format`

PR #3450, opened 2026-06-29
[NPQ-3858] TRN allocated email

alkesh:
### Context

Ticket: [NPQ-3858](https://dfedigital.atlassian.net/browse/NPQ-3858)

send an email when a TRN is allocated

1. an email is sent when a TRN is allocated immediately (allocate TRN job)
2. an email is send if a TRN is allocated later (TRN request completion webhook)

[NPQ-3858]: https://dfedigital.atlassian.net/browse/NPQ-3858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/teaching_record_system/webhooks/trn_request_completed_processor.rb
+++ b/app/services/teaching_record_system/webhooks/trn_request_completed_processor.rb
@@ -6,7 +6,7 @@ module TeachingRecordSystem
       def process!
         user.update!(trn: new_trn, trn_verified: true, trn_auto_verified: true)
 
-        if user.trn_previously_changed?(from: nil)
+        if user.trn_previously_changed?(from: nil) && user.email.present?
           TrnAllocatedMailer.trn_allocated_mail(to: user.email, full_name: user.full_name, trn: new_trn).deliver_later
         end
       end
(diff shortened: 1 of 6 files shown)


app/services/teaching_record_system/webhooks/trn_request_completed_processor.rb
```diff
@@ -5,6 +5,7 @@ class TrnRequestCompletedProcessor < Base
 
       def process!
         user.update!(trn: new_trn, trn_verified: true, trn_auto_verified: true)
+        TrnAllocatedMailer.trn_allocated_mail(to: user.email, full_name: user.full_name, trn: new_trn).deliver_later
```
jebw: I think this will need to check whether its just updated the users TRN from `nil` - if it had already been allocate via the initial API then this would result in duplicate emails being sent

jebw (commented):
Looks good but needs to handle the 'double notification' scenario where the API does return a TRN (likely in a majority of cases) and we then subsequently get an API callback.

In theory I guess there's a race condition where the API call triggers the webhook before it completes itself and we update our DB but because we process webhooks asychronously via background jobs I really don't see that as a probably scenario.

jebw (approved):
LGTM

PR #3454, opened 2026-06-30
[NPQ-3839] - remove npq_separation namespace entirely

spqr-praetoria:
### Context

Ticket: [NPQ-3839](https://dfedigital.atlassian.net/browse/NPQ-3839)

NPQ separation namespace no longer required

1. Move files from npq_separation namespace to admin 
2. redirect request to /npq_separation/admin to /admin

[NPQ-3839]: https://dfedigital.atlassian.net/browse/NPQ-3839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/features/admin/admin_navigation_spec.rb
alkesh:
```suggestion
```
I think this line can be deleted

ltello:
I wonder if we should rename all `ensure_super_admin` methods to `require_super_admin` (already existing) because both have the same purpose?

config/routes.rb
```diff
@@ -123,174 +121,178 @@
...
+      collection do
+        post :search
       end
     end
   end
 
+  # Redirect legacy /npq-separation/admin URLs to the new /admin namespace.
+  # Only needed during cut over; deep links land on the admin root, not their exact new page.
+  get "/npq-separation/admin", to: redirect("/admin")
+  get "/npq-separation/admin/*path", to: redirect("/admin")
```
ltello, line 294: We don't want `get "/npq-separation/admin/*path", to: redirect("/admin/%{path}")` instead?

alkesh:
> I wonder if we should rename all `ensure_super_admin` methods to `require_super_admin` (already existing) because both have the same purpose?

There is a slight different - `ensure_super_admin` does not redirect you to the sign-in page, but shows a warning instead - it is used for pages that admins can see, but there are super-admin only operations on that page.
I've renamed `ensure_super_admin` to `warn_if_not_super_admin` to make it clearer.

ltello (approved):

config/routes.rb
alkesh: The ticket said not to:
> this is only relevant during cut over so I wouldn’t worry about redirecting to the ‘correct’ new admin page

PR #3455, opened 2026-06-30
[BAU] Downscale webapps and workers

jebw:
### Context

Ticket: BAU

We are scaled quite large to allow for registration surge - that surge has now passed and we should downscale to reduce running costs

_This should also help with the occasional alerts about concurrent deploys_

1. Halved number of webapps from 12 to 6
2. Halved number of works from 4 to 2


ltello (approved):
Isn't it 6 web replicas still too much? We had 4 in RECT and everything was ok. But not sure if this service has kind of the same traffic as RECT

jebw:
Yeah its way overscaled but I'd figure we could drop in stages rather than all at once?

PR #3459, opened 2026-07-01
[NPQ-3451] large scale seeding

alkesh:
### Context

Ticket: [NPQ-3451](https://dfedigital.atlassian.net/browse/NPQ-3451)

rake task to allow ~20,000 applications to be seeded into a review app

1. change `add_applications.rb` and `add_declarations.rb` to use classes, so their functionality can be resued
2. create a job `SeedingJob` to allow seeding to happen in the background, as doing it in one go causes an out-of-memory error on the worker pod - the `multiplier` value has been carefully chosen, `30` causes the worker to run out of memory, `20` is fine.
3. switched standard DB seeds to use the background job - should speed up review app deployment
4. rake task to invoke the seeding
5. rake task to check the status of the seeding
6. created a document describing how to use the rake tasks: `docs/database_seeding.md`

[NPQ-3451]: https://dfedigital.atlassian.net/browse/NPQ-3451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,22 +8,9 @@ PaperTrail.enabled = false
 Faker::Config.locale = "en-GB"
 
 def load_base_file(file)
-  base_file = Rails.root.join("db", "seeds", "base", file)
+  base_file = Rails.root.join("db/seeds/base", file)
 
   load(base_file)
-end
-
-def with_versioning
-  was_enabled = PaperTrail.enabled?
-  was_enabled_for_request = PaperTrail.request.enabled?
-  PaperTrail.enabled = true
-  PaperTrail.request.enabled = true
-  begin
-    yield
-  ensure
-    PaperTrail.enabled = was_enabled
-    PaperTrail.request.enabled = was_enabled_for_request
-  end
 end
 
 def load_csv(file, model_class)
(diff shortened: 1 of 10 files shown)


docs/database_seeding.md
```diff
@@ -0,0 +1,20 @@
...
+
+To seed a database with a large number of applications - approx ~19,000 applications, ~6,000 declarations, and ~17,000 users,
+run the following rake task:
+
+```bash
+  rake 'large_seed:background'
+```
+
+This will set off a background job, to repeatedly run our application and declaration seeds,
+which should result in ~20,000 applications and ~20,000 declarations, and ~17,000 users being created in the database.
```
jebw: Minor nit but the numbers of this line are different to those on line 10?

db/seeds/base/add_applications.rb
```diff
@@ -1,165 +1,170 @@
...
+      { employment_type: :other },
+      { referred_by_return_to_teaching_adviser: "yes" },
+    ].each do |application_attrs|
+      FactoryBot.create(:application, :manual_review, **application_attrs.merge(course:))
+    end
+
+    # temporarily comment the below out, to see if the large-scale seeding job will complete
+    # Application.order(id: :desc).each.with_index do |a, i|
+    #   a.update!(created_at: i.days.ago)
+    # end
```
jebw: Not following this - shouldn't it be enabled? Or applicatoins themselves created with appropriate `created_at` dates, given the 'date range' 1 day per application covers I wonder if this should be something more algorithmic, eg

* random spread over 6 months around cohort application is using?
* 6 months around cohort divided by total number of applications being created in this job and incrementing by that amount?

jebw (commented):
Looks good to me, only query is around the historical relevance of the created at date on applications but part of me wonders how important that is for the data to be useful anyway?

db/seeds.rb
```diff
@@ -56,29 +43,30 @@ def load_csv(file, model_class)
...
+  "add_api_tokens.rb" => nil,
+  "process_statements.rb" => nil,
+  "add_delivery_partners.rb" => nil,
+  "add_eligibility_list_entries.rb" => nil,
+  "add_course_cohort_providers.rb" => nil,
+}.each do |seed_file, seed_class|
   Rails.logger.info("seeding #{seed_file}")
   ApplicationRecord.transaction do
     load_base_file(seed_file)
+    seed_class.constantize.new.load if seed_class
```
jebw: This could leverage our background jobs to speed up getting review apps available

```ruby
ApplicationRecord.transaction do
  load_base_file(seed_file)
  seed_class.constantize.new.load if seed_class && Rails.env.local?
end

SeedingJob.perform_later unless Rails.env.local?
```

jebw (commented):
Some how missed this one from the review yesterday - small change but should speed up deploys of review apps? WDYT?

db/seeds/base/add_applications.rb
alkesh: I've added a trait to the applications factory called `:with_random_cohort_bounded_created_at` which will now do this.

db/seeds.rb
alkesh: 👍🏽 It looks like it reduces the review app deploy time from 5mins to 4mins - so a 20% improvement,
but then the background seeding job is still running when the review app is first deployed.

jebw (approved):
LGTM

PR #3463, opened 2026-07-07
[NPQ-3872] Don't notify Sentry of a blank cohort selection in the application journey

ltello:
### Context

Ticket: [NPQ-3872](https://dfedigital.atlassian.net/browse/NPQ-3872)

We were getting a Sentry alert every time a user did not pick a cohort on the course start date screen. We only want it when the user picks a cohort we cannot find in the database.

1. Skip the `cohort_exists` check when the cohort is blank, so we no longer notify Sentry for an empty selection. The presence validation already handles the empty case.
2. Update the specs to cover both cases: an unknown cohort still notifies Sentry, and no cohort selected does not.

### Notes

[NPQ-3872]: https://dfedigital.atlassian.net/browse/NPQ-3872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/forms/questionnaires/course_start_date_spec.rb
+++ b/spec/forms/questionnaires/course_start_date_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe Questionnaires::CourseStartDate, type: :model do
     context "when the chosen cohort that does not correspond to an existing cohort" do
       before { instance.course_start_cohort = described_class::OPTIONS.keys.last }
 
-      it { is_expected.to have_error(:course_start_cohort, :invalid) }
+      it { is_expected.to have_error(:course_start_cohort, :invalid, "Cohort selected does not exist.") }
 
       it "notifies Sentry" do
         expect(Sentry).to receive(:capture_message).with(/Cohort selected by user does not exist/)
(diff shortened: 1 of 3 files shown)


spec/forms/questionnaires/course_start_date_spec.rb
```diff
@@ -12,8 +12,24 @@
...
       it { is_expected.to have_error(:course_start_cohort, :invalid) }
+
+      it "notifies Sentry" do
+        expect(Sentry).to receive(:capture_message).with(/Cohort selected by user does not exist/)
+        instance.valid?
+      end
+    end
+
+    context "when no cohort is selected" do
+      it { is_expected.to have_error(:course_start_cohort, :blank) }
```
alkesh:
```suggestion
      it { is_expected.to have_error(:course_start_cohort, :blank, "Choose your course start date.") }
```
should have the error message from locale file - this may be an existing bug.

ltello: Fixed in a new commit.

alkesh (approved):

PR #3464, opened 2026-07-09
[NPQ-3824] check if you're eligible page and routing

alkesh:
### Context

Ticket: [NPQ-3824](https://dfedigital.atlassian.net/browse/NPQ-3824)

new page "Check if you're eligible"

* updated `#next_step` on `course_start_date` step
* new step - `check_funding`
* updated `#previous_step` on the `choose_your_npq` step
* updated `#previous_step` on the `teacher_catchment` step
* new folder for post-MVP feature specs
* new feature spec for the `check_funding` step - includes testing back links
* disabled other journey feature specs for now by adding a tag: `:mvp`

[NPQ-3824]: https://dfedigital.atlassian.net/browse/NPQ-3824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/forms/questionnaires/check_funding.rb
+++ b/app/forms/questionnaires/check_funding.rb
@@ -1,0 +1,36 @@
+module Questionnaires
+  class CheckFunding < Base
+    QUESTION_NAME = :check_funding
+
+    OPTIONS = {
+      "yes" => "Check funding",
+      "no" => "Continue without DfE funding",
+    }.freeze
+
+    attribute QUESTION_NAME
+
+    validates QUESTION_NAME, presence: true, inclusion: { in: OPTIONS.keys }
+
+    def self.permitted_params
+      [QUESTION_NAME]
+    end
+
+    def options
+      OPTIONS.map do |value, label|
+        build_option_struct(value:, label:)
+      end
+    end
+
+    def previous_step
+      :course_start_date
+    end
+
+    def next_step
+      if check_funding == "yes"
+        :teacher_catchment
+      else
+        :choose_your_npq
+      end
+    end
+  end
+end
--- a/app/forms/questionnaires/choose_your_npq.rb
+++ b/app/forms/questionnaires/choose_your_npq.rb
@@ -22,6 +22,8 @@ module Questionnaires
              :employment_type_other?,
              :works_in_childcare?,
              :kind_of_nursery_public?,
+             :cohort_funded?,
+             :check_funding?,
              to: :query_store
 
     def self.permitted_params
@@ -91,22 +93,18 @@ module Questionnaires
     end
 
     def previous_step
-      if inside_catchment? && referred_by_return_to_teaching_adviser?
-        :referred_by_return_to_teaching_adviser
-      elsif inside_catchment? && works_in_school?
-        :choose_school
-      elsif inside_catchment? && works_in_childcare?
-        if kind_of_nursery_public?
-          :choose_childcare_provider
-        elsif has_ofsted_urn?
-          :choose_private_childcare_provider
+      if cohort_funded?
+        if check_funding?
+          if inside_catchment?
+            :teacher_catchment
+          else
+            :ineligible_for_funding
+          end
         else
-          :have_ofsted_urn
+          :check_funding
         end
-      elsif inside_catchment? && works_in_another_setting?
-        :your_employment
       else
-        :work_setting
+        :course_start_date
       end
     end
 
--- a/app/forms/questionnaires/course_start_date.rb
+++ b/app/forms/questionnaires/course_start_date.rb
@@ -41,7 +41,11 @@ module Questionnaires
     end
 
     def next_step
-      :provider_check
+      if Cohort.find_by(identifier: course_start_cohort).funded?
+        :check_funding
+      else
+        :choose_your_npq
+      end
     end
 
     def previous_step
--- a/app/services/registration_query_store.rb
+++ b/app/services/registration_query_store.rb
@@ -201,4 +201,15 @@ class RegistrationQueryStore
   def course_start_cohort
     store["course_start_cohort"]
   end
+
+  def cohort_funded?
+    cohort = Cohort.find_by(identifier: course_start_cohort)
+    return true unless cohort
+
+    cohort.funded?
+  end
+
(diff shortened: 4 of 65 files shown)


app/services/registration_query_store.rb
```diff
@@ -201,4 +201,12 @@ def institution
   def course_start_cohort
     store["course_start_cohort"]
   end
+
+  def course_started_autumn_2026?
```
ltello: Surely I'm miss the reason why we are making the code 2026-static? Can't this be generic for any cohort year?

app/forms/questionnaires/choose_your_npq.rb
```diff
@@ -91,22 +93,18 @@ def next_step
     end
 
     def previous_step
-      if inside_catchment? && referred_by_return_to_teaching_adviser?
-        :referred_by_return_to_teaching_adviser
-      elsif inside_catchment? && works_in_school?
-        :choose_school
```
ltello, line 97: Are we removing these decision branches because in the next coming tickets `choose_school` next step will never be `:choose_your_npq`? It still seems to be the case.

app/views/registration_wizard/check_funding.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l"><%= t("helpers.title.registration_wizard.check_funding") %></h1>
+
+    <p class="govuk-body">
+      We’ll ask a few questions to check if you can get funding for courses starting in
+      <%= Questionnaires::CourseStartDate::OPTIONS[@wizard.store["course_start_cohort"]][:cohort_description] %>.
```
ltello: This could raise an exception if `@wizard.store["course_start_cohort"]` is nil for any reason?. Maybe using `.dig` might mitigate the issue

app/forms/questionnaires/check_funding.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def next_step
+      if check_funding == "yes"
+        :teacher_catchment
+      else
+        :choose_your_npq
+      end
+    end
+  end
+end
```
ltello, line 36: Shouldn't we add some type of validation to the value received for `check_funding`?

spec/rails_helper.rb
```diff
@@ -86,6 +86,12 @@
   config.include RSpec::DefaultHttpHeader, type: :request
   config.include AxeHelper, type: :feature
 
+  # slightly hacky way exclude most features - can be renabled before the post-MVP feature branch (NPQ-3741) is merged into main
+  # post-MVP features that need default schedules can use:
+  #   include_context "with default schedules"
+  # instead of using the tag
+  config.filter_run_excluding :with_default_schedules
```
ltello: I think this is excluding some `spec/services` no related to this PR

app/forms/questionnaires/check_funding.rb
alkesh: I did have validation in here initially, but then removed it, because it removes the need to show an error on the page, and the backend logic only cares if the value is 'yes' or not, so hacking the value doesn't do anything.
The only thing it's adding is ensuring the value we store for `check_funding` is one we expect.
I guess it's worth adding back. 👍🏽

app/forms/questionnaires/choose_your_npq.rb
alkesh: yes, based on the new flow: https://lucid.app/lucidspark/7b85bb9d-194c-4202-b389-c90e9f6b213b/edit?view_items=0UKSeQ~apJw6&page=0_0&invitationId=inv_ea883e4b-ab19-4100-9ac4-b94c5a166475
the previous steps are only the ones below

app/services/registration_query_store.rb
alkesh: ah yes - let me change this - this should really be checking if the chosen cohort is funded or not.

spec/rails_helper.rb
alkesh: ah yes - let me check this - my initial intention was to try and only filter feature specs with this tag - I'll see if I can make that work.

alkesh: I've now added a tag: `:mvp` to all feature specs we don't want to run instead.

app/views/registration_wizard/check_funding.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l"><%= t("helpers.title.registration_wizard.check_funding") %></h1>
+
+    <p class="govuk-body">
+      We’ll ask a few questions to check if you can get funding for courses starting in
+      <%= Questionnaires::CourseStartDate::OPTIONS.dig(@wizard.store["course_start_cohort"],[:cohort_description]) %>.
```
ltello:
```suggestion
      <%= Questionnaires::CourseStartDate::OPTIONS.dig(@wizard.store["course_start_cohort"], :cohort_description) %>.
```

ltello (approved):

app/forms/questionnaires/course_start_date.rb
```diff
@@ -55,10 +59,11 @@ def return_to_regular_flow_on_change?
...
     def cohort_exists
-      return if Cohort.find_by(identifier: course_start_cohort)
-
-      errors.add(QUESTION_NAME, :invalid)
-      Sentry.capture_message("Cohort selected by user does not exist: #{course_start_cohort}")
+      cohort = Cohort.find_by(identifier: course_start_cohort)
+      unless cohort
+        errors.add(QUESTION_NAME, :invalid)
+        Sentry.capture_message("Cohort selected by user does not exist: #{course_start_cohort}")
+      end
```
ltello: This refactor has been done already on the base branch and that's the conflict this PR is getting. I'd just discard this change when solving the conflict and all good.

PR #3475, opened 2026-07-17
[NPQ-3920] merge other users with same TRN

alkesh:
### Context

Ticket: [NPQ-3920](https://dfedigital.atlassian.net/browse/NPQ-3920)

Merge and archive other users with same TRN in the webhooks

1. for the`trn_request.completed` webhook, merge and archive other users with the same TRN
2. for the `one_login_user.updated` webhook, merge users with the same TRN into the most recenlty created user

[NPQ-3920]: https://dfedigital.atlassian.net/browse/NPQ-3920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/teaching_record_system/webhooks/user_updated_processor.rb
+++ b/app/services/teaching_record_system/webhooks/user_updated_processor.rb
@@ -17,7 +17,7 @@ module TeachingRecordSystem
 
       def params_to_update
         { email: user_email }.tap do |params|
-          if new_trn.present? && new_trn != user.trn
+          if new_trn.present?
             params[:trn] = new_trn
             params[:trn_verified] = true
             params[:trn_auto_verified] = true
@@ -40,6 +40,7 @@ module TeachingRecordSystem
       def merge_and_archive_other_users_with_same_trn
         users_with_same_trn = User.not_archived.with_trn(new_trn).order(created_at: :desc).to_a
         user_to_keep = users_with_same_trn[0]
+
         users_with_same_trn[1..].each do |user_to_merge|
           Users::MergeAndArchive.new(user_to_merge:, user_to_keep:).call(dry_run: false)
         end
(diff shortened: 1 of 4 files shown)


app/services/teaching_record_system/webhooks/user_updated_processor.rb
```diff
@@ -34,6 +37,14 @@ def new_trn
         webhook_message.message["connectedPerson"]["trn"] if webhook_message.message["connectedPerson"]
       end
 
+      def merge_and_archive_other_users_with_same_trn
+        users_with_same_trn = User.not_archived.with_trn(new_trn).order(created_at: :desc).to_a
```
ltello, line 41: Are we sure this always include at least the user to keep? 
Otherwise, `users_with_same_trn[1..].each  ...` will raise an exception

alkesh: I think there is always a user to keep.
Archived users always have their UID blanked, so they will not be the main user matched by the webhook,
and the webhook has updated the main user's TRN with `new_trn` at this point, so there should be at least one user with that TRN.
I've added a couple of extra scenarios to the spec - for when the TRN is blank, and when the TRN is unchanged - both scenarios are fine.

jebw (approved):
LGTM

app/services/teaching_record_system/webhooks/user_updated_processor.rb
ltello: @jebw That's what I think Ruby should behave but surprisently you get nil back when the range start is out of the array!

ltello (approved):

PR #3479, opened 2026-07-21
[NPQ-3873] move login

alkesh:
### Context

Ticket: [NPQ-3873](https://dfedigital.atlassian.net/browse/NPQ-3873)

Move login to end of journey

1. clicking 'Start now' on the start page does not redirect to a login, but continues to /course-start-date
2. after checking answers, the continue button redirects to a  login.
3. created a `check_answers_log_in_and_submit` helper for the journey specs, that checks the values on the check answers page, continues to the login page, and then confirms the check answers page again, before hitting the submit button.
4. clear the query store in the controller after completing a registration - this is so we can use the fact the store is empty to determine if they have completed a registration and are then attempting to access steps directly. 
5. allow closed registration exception users to still log in and complete the journey
6. re-enabled `spec/features/journeys/sad_paths/navigating_directly_to_pages_spec.rb`

[NPQ-3873]: https://dfedigital.atlassian.net/browse/NPQ-3873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/registration_wizard_controller.rb
+++ b/app/controllers/registration_wizard_controller.rb
@@ -4,7 +4,6 @@ class RegistrationWizardController < PublicPagesController
   before_action :set_form
   before_action :check_end_of_journey, only: %i[update]
   before_action :check_course_defined, only: %i[show]
-  before_action :check_teacher_auth_user
 
   rescue_from FundingEligibility::MissingMandatoryInstitution, with: :redirect_to_institution_picker
   rescue_from RegistrationWizard::RemovedStep, with: :redirect_to_course_start_date
@@ -108,13 +107,15 @@ private
   def check_end_of_journey
     if @form.valid? && @form.last_step?
       @wizard.save!
+      @wizard.store.clear
       redirect_to registration_complete_accounts_user_registration_path(current_user.applications.last)
     end
   end
 
   def check_course_defined
-    redirect_to_course_start_date if !FORMS_FOR_STEPS_BEFORE_COURSE_IS_CHOSEN.include?(@form.class) && defined?(@form.course) && !@form.course
-    redirect_to_course_start_date if @form.instance_of?(Questionnaires::CheckAnswers) && !@wizard.course
+    return if FORMS_FOR_STEPS_BEFORE_COURSE_IS_CHOSEN.include?(@form.class)
+
+    redirect_to_course_start_date if defined?(@form.course) && !@form.course
   end
 
   def registration_closed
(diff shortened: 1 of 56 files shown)


app/controllers/registration_wizard_controller.rb
```diff
@@ -108,13 +112,20 @@ def set_form
...
-    redirect_to_course_start_date if !FORMS_FOR_STEPS_BEFORE_COURSE_IS_CHOSEN.include?(@form.class) && defined?(@form.course) && !@form.course
-    redirect_to_course_start_date if @form.instance_of?(Questionnaires::CheckAnswers) && !@wizard.course
+    return if FORMS_FOR_STEPS_BEFORE_COURSE_IS_CHOSEN.include?(@form.class)
+    return if @wizard.query_store.course
+
+    # redirect to course start date if the user has started the registration journey
+    return redirect_to_course_start_date if @wizard.query_store.has_answers?
+
+    # redirect to the start of the registration journey if the user has not started the registration journey
+    redirect_to "/"
```
jebw: `root_path` ?

app/forms/questionnaires/start.rb
```diff
@@ -5,14 +5,7 @@ def requirements_met?
...
     def next_step
-      if query_store.current_user
-        first_questionnaire_step
-      else
-        # This shouldn't really be reached, in this situation the user should have been
-        # presented the omniauth kickoff button on the start back. This is here as a fallback
-        # to send the user back to the start page where they'll be presented the oauth button again.
-        :start
-      end
+      first_questionnaire_step
```
jebw: AFAICT this is the only user of  `first_questionnaire_step`, which is hard coded to a single value - which means that value could be in here and `FlowHelper` removed as a nice clean up

app/forms/questionnaires/continue_to_login.rb
```diff
@@ -0,0 +1,11 @@
+module Questionnaires
+  class ContinueToLogin < Base
+    def previous_step
+      :check_answers
+    end
+
+    def next_step
+      :check_answers
+    end
```
jebw, line 9: The re-use of check answers at multiple points in the wizard is problematic for a future conversion to `dfe-wizard` and feels unnecessarily odd. I'd much prefer 2 steps for the 2 check your answers screens - they can share code or use inheritance but should be separate steps?

app/forms/questionnaires/choose_school.rb
```diff
@@ -16,6 +16,10 @@ def self.permitted_params
       ]
     end
 
+    def requirements_met?
+      query_store.work_setting
```
jebw: should this be `super && query_store.work_setting` in case any common requirements get added to `Base` in the future

jebw (commented):
Looks good but forward and back to the same step seems odd - I'd think this would be better with 2 steps (potentially one inheriting from the other and sharing partials)

```mermaid
flowchart TD
    A[Share provider] 
    A --> B{Logged In}
    B -->|Yes| E[Second Summary]
    B -->|No| C[First Summary]
    C --> D[Sign In]
    D --> E
```

app/services/registration_wizard_visualiser.rb
alkesh, line 1: `Questionnaires::FlowHelper#first_questionnaire_step` was the only helper method we had, and this has now been removed - so `flow_helper` parts of this visualiser are not needed any more.

PR #3510, opened 2026-07-27
[NPQ-3898] Add declarations dashboard

ltello:
### Context

Ticket: [NPQ-3898](https://dfedigital.atlassian.net/browse/NPQ-3898)

Contract managers need one place in the admin console to see declaration data for a provider and a cohort. Right now they have to use external reporting for this. This PR adds the new Declarations dashboard page: the navigation link, the page shell and the summary table.

1. Add a "Declarations dashboard" link to the Dashboards side navigation.
2. Add the new page with a Provider dropdown and a Cohort dropdown side by side, and a Select button.
3. Show an inset panel and no table when no provider and cohort are chosen yet.
4. Add a filter class that checks a provider and a cohort are chosen, and shows errors on the form when they are not.
5. Show the summary table when both are chosen, with the columns Declaration type, Expected, Received and Remaining, one row per declaration stage (Started, Retained-1, Retained-2, Completed) and a Total declarations row.

### Notes

The counts show a dash instead of real numbers because this ticket does not cover the calculations.

[NPQ-3898]: https://dfedigital.atlassian.net/browse/NPQ-3898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/forms/admin/dashboards/declarations_filter.rb
```diff
@@ -0,0 +1,31 @@
...
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :lead_provider_id
+    attribute :cohort_id
+
+    validates :lead_provider_id, presence: true
+    validates :cohort_id, presence: true
+
+    def initialize(attributes = {}, submitted: false)
```
jebw: Rather than overriding the initialize behaviour, you could have an attribute of submitted which defaults to false?

```ruby
attribute :submitted, :boolean, default: false
```

And then do the marking submitted in the controller

```ruby
@filter = Filter.new(filter_params)
@filter.submitted = params.key?(:declarations_filter)
@filter.validate if filter.submitted?
```

app/controllers/admin/dashboards_controller.rb
```diff
@@ -1,11 +1,31 @@
...
+      build_declarations_filter
+    elsif params[:cohort_id].present?
       @cohort = Cohort.find_by(id: params[:cohort_id])
       @applications = Application.where(cohort: @cohort)
     else
       @applications = Application
     end
   end
+
+private
```
jebw: Given this is quite different to the other two dashboards which are both doing grouped application counts I think it makes more sense to have this as its own controller? If more dashboards get added in the future the risk is this ends up as a catch all with many private methods? A lot of the private method code could become regular `#show` action code?

app/forms/admin/dashboards/declarations_filter.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Admin::Dashboards
+  class DeclarationsFilter
```
jebw: Given we'll also want somewhere to do the calculations - rather than call this filter, maybe name it something like `DeclarationStatistics` or `DeclarationsDashboard`?

jebw: At that point it might belong in a service rather than a `/forms` ?

Or do you think it makes more sense to have the service separate from the form object - my main consideration is I think we'll want to do the same checks for ensuring valid data is being fed into the service as we are doing for the form object - at which point it might be simpler to combine them?

jebw (commented):
Looks good - part I'm a bit wary of is if we end up with that dashboards controller getting more complex over time so I'd wondered about just splitting it too its own controller?

jebw (approved):
Thanks for refactoring this - to my eyes at least, thats much easier to follow and sets up a nice pattern if we're to add other dashboards in the future

I think we'll need a spec for the service adding in the next PR but given it doesn't do much in this one that doesn't seem significant 👍

PR #3524, opened 2026-07-29
NPQ-3918 Add eligibility column to workplaces admin page

jebw:
### Context

Ticket: [NPQ-3918](https://dfedigital.atlassian.net/browse/NPQ-3918)

There is a need to see if a workplace appears on any of the eligibility lists.

1. Add a column to the Workplaces table showing which eligibility lists a workplace appears on
2. Only show the sidenav column if we do have a sidenav for the page - _a lot of the admin pages don't have a sidenav and always showing this restricts the available horizontal space for the wider tables_

### Notes

Optional section of anything unusual in the PR which requires further explanation

[NPQ-3918]: https://dfedigital.atlassian.net/browse/NPQ-3918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -5,6 +5,14 @@ class School < ApplicationRecord
   PRIMARY_PHASE = "Primary".freeze
   MIDDLE_DEEMED_PRIMARY_PHASE = "Middle deemed primary".freeze
 
+  ELIGIBILITY_LISTS = [
+    EligibilityList::Pp50School,
+    EligibilityList::Pp50FurtherEducation,
+    EligibilityList::DisadvantagedEarlyYearsSchool,
+    EligibilityList::LocalAuthorityNursery,
+    EligibilityList::RiseSchool,
+  ].map(&:name).freeze
+
   ELIGIBLE_ESTABLISHMENT_TYPE_CODES = {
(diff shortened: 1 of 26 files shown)


app/models/school.rb
```diff
@@ -5,6 +5,14 @@ class School < ApplicationRecord
   PRIMARY_PHASE = "Primary".freeze
   MIDDLE_DEEMED_PRIMARY_PHASE = "Middle deemed primary".freeze
 
+  ELIGIBILITY_LISTS = [
+    EligibilityList::Pp50School,
+    EligibilityList::Pp50FurtherEducation,
+    EligibilityList::DisadvantagedEarlyYearsSchool,
```
alkesh: I'm wondering if we should maybe not include the DisadvantagedEarlyYearsSchool list here,
as it's not currently used, and there's a question whether how we used to use it was correct.
I'm thinking it is maybe worth just removing the EligibilityList::DisadvantagedEarlyYearsSchool model, and admin tab, and we'll re-implement it if it's required.

PR #3525, opened 2026-07-29
2811 - Review app using Postgres v17 container

hopep-mobilise:
### Context

https://trello.com/c/0Ux3mX0O/2811-upgrade-npq-postgres-version

We're looking to upgrade the Postgres version to 17.

- This PR will create a review app to check application functionality with Postgres v17 containerised.
- A new variable, postgres_server_version will allow configuration between environments. The default is set to the current version 14.
- postgres_server_version set to 17 in review.tfvars.json
- Change the Docker container versions of Postgres to v17

### Guidance to review

- Run `make review terraform-plan PULL_REQUEST_NUMBER=3525 DOCKER_IMAGE=main` only the Kubernetes image should change.
- Check the npq-registration-review-3525-postgres deployment to ensure the image is ghcr.io/dfe-digital/teacher-services-cloud:postgres-17-alpine 
- Run `make staging terraform-plan DOCKER_IMAGE=main`  only the Kubernetes image should change.
- Test the [review app](https://npq-registration-review-3525-web.test.teacherservices.cloud/) crated by this PR.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

### DIFF
--- a/terraform/application/config/review.tfvars.json
+++ b/terraform/application/config/review.tfvars.json
@@ -7,5 +7,6 @@
     "command": ["/bin/sh", "-c", "RAILS_ENV=review bundle exec rails db:environment:set db:migrate && bundle exec rails server -b 0.0.0.0"],
     "enable_logit": true,
     "enable_dfe_analytics_federated_auth": true,
-    "dataset_name": "npq_events_review"
+    "dataset_name": "npq_events_review",
+    "postgres_server_version": 17
 }
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     command: sleep infinity
     network_mode: service:postgres
   postgres:
-    image: postgres:14
+    image: postgres:17
     restart: unless-stopped
     volumes:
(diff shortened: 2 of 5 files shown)


terraform/application/config/review.tfvars.json
```diff
@@ -2,10 +2,11 @@
     "cluster": "test",
     "namespace": "cpd-development",
     "environment": "review",
-    "deploy_azure_backing_services": false,
+    "deploy_azure_backing_services": true,
```
jebw: So ideally I'd like this PR to be merged so all use of review apps are testing against PG17 - and this would also swap all review apps to creating azure pg and redis? Would it make more sense to raise this with it set to false and another PR building off it set to true if we want a specific review app backed by 'real' services?

hopep-mobilise: This is not the way review apps are set up and would increase costs.
We can't merge as is, because the azure.extensions are hard coded list and have to include PLPGSQL for a new database. If this was merged it would push that change to all servers and all would restart.

jebw (approved):
LGTM

PR #3526, opened 2026-07-30
[NPQ-3945] Only offer NPQs with a provider in the chosen cohort

ltello:
### Context

Ticket: [NPQ-3945](https://dfedigital.atlassian.net/browse/NPQ-3945)

Users who chose the Spring cohort could get stuck. They picked an NPQ, answered all the questions, and then the provider page showed an empty list, with no way to go on and submit. It happened both when choosing Spring from the start and when changing the course start date from Autumn to Spring on the check answers page.

The cause is that the choose your NPQ step listed every course and never looked at the cohort. Some courses have no lead provider in Spring, so the provider list came back empty.

1. New `Course.offered_in(cohort)` scope: the courses that at least one lead provider delivers in that cohort.
2. The choose your NPQ step now only lists those courses. The validation uses the same rule, so a user who changes the start date from Autumn to Spring cannot keep a course with no Spring provider.
3. Added `requirements_met?` to the step, so a session with no course start date is sent back instead of failing.
4. Tests for the new scope, for the step, and a full journey test that a Spring user is not offered courses with no Spring provider.
5. Fixed some journey specs. Some of them never really answered the course start date step because they had no cohorts, so they now use `:with_cohorts`. Others built the form with no wizard to list the courses, which needs a cohort now.

[NPQ-3945]: https://dfedigital.atlassian.net/browse/NPQ-3945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/forms/questionnaires/choose_your_npq.rb
+++ b/app/forms/questionnaires/choose_your_npq.rb
@@ -28,6 +28,10 @@ module Questionnaires
       [QUESTION_NAME]
     end
 
+    def requirements_met?
+      super && cohort.present?
+    end
+
     def questions
       [
         QuestionTypes::RadioButtonGroup.new(
@@ -127,12 +131,15 @@ module Questionnaires
     end
 
     def valid_providers
-      cohort = Cohort.find_by!(identifier: wizard.query_store.course_start_cohort)
       LeadProvider.for(course:, cohort:)
     end
 
+    def cohort
+      @cohort ||= Cohort.find_by(identifier: wizard.query_store.course_start_cohort)
+    end
+
     def courses
-      Course.where(display: true).order(:position)
+      Course.offered_in(cohort).where(display: true).order(:position)
     end
 
     def previous_course
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -16,6 +16,13 @@ class Course < ApplicationRecord
   validates :identifier, presence: true, uniqueness: true
   validates :ecf_id, uniqueness: { case_sensitive: false }, allow_nil: true
 
+  # Only courses at least one lead provider delivers in the given cohort
+  scope :offered_in, lambda { |cohort|
+    joins(course_cohorts: :course_cohort_providers)
+      .where(course_cohorts: { cohort: })
+      .distinct
+  }
+
   # npq-additional-support-offer is replaced by npq-early-headship-coaching-offer
   IDENTIFIERS = %w[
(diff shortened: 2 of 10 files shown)


app/forms/questionnaires/choose_your_npq.rb
```diff
@@ -127,12 +131,15 @@ def lead_provider_valid?
...
       LeadProvider.for(course:, cohort:)
     end
 
+    def cohort
+      @cohort ||= Cohort.find_by(identifier: wizard.query_store.course_start_cohort)
+    end
+
     def courses
-      Course.where(display: true).order(:position)
+      Course.offered_in(cohort).where(display: true).order(:position)
```
avinhurry, line 142: If the Early headship coaching offer is not available for this cohort, this filter removes it from the list. The page then shows "or" before the last normal course (for example, SENCO). Is the divider intended to be shown only when the Early headship coaching offer is present?

image

avinhurry (approved):
Just had one question on the course list divider but looks good overall to me.

app/forms/questionnaires/choose_your_npq.rb
ltello: the divider is meant to be present whenever there are several courses displayed

ltello: Thanks for the review @avinhurry !!

avinhurry (approved):

app/forms/questionnaires/choose_your_npq.rb
peteryates: Usually the 'or' is reserved for cases where one option is significantly different to the others. I don't think it's necessary here.

peteryates (approved):
Code-wise this looks fine to me.

I left a comment about the 'or'. I don't think the final course needs it as it's just another course not a completely different option - but that needn't block if this is affecting people and can easily be fixed later.

app/forms/questionnaires/choose_your_npq.rb
jebw: I'll feed this back to the BA's - we've not currently got a UCD team but the PR itself is addressing a bug so I'll get this reviewed and merged as is for now

jebw (approved):
LGTM

PR #3539, opened 2026-08-07
[NPQ-3940] page to show archived users with blank emails and applications

alkesh:
### Context

Ticket: [NPQ-3940](https://dfedigital.atlassian.net/browse/NPQ-3940)

New admin page called "Archived duplicate users"

1. new admin page, under the "Users" tab
2. updated the seeds to create some archived users, with blank emails, that have an application that is not rejected
3. added a new method to `User`: `verified_trn` - which returns the TRN only if it is verified - we seemed to use this logic in a few places

[NPQ-3940]: https://dfedigital.atlassian.net/browse/NPQ-3940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/admin/users/duplicates_controller.rb
+++ b/app/controllers/admin/users/duplicates_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Admin::Users
+  class DuplicatesController < AdminController
+    def index
+      scope = User
+        .where(email: nil)
+        .joins(:applications)
+        .where.not(applications: { lead_provider_approval_status: "rejected" })
+        .distinct
+        .order(created_at: :desc)
+
+      @pagy, @users = pagy(scope)
+      @potential_matches = User.where(email: @users.map(&:archived_email)).group_by(&:email)
+    end
+  end
+end
--- a/app/services/users/archiver.rb
+++ b/app/services/users/archiver.rb
@@ -7,7 +7,7 @@ module Users
 
     attribute :user
 
-    def archive!(blank_email: false)
+    def archive!(blank_email: false, enable_sentry: true)
       raise ArgumentError, "User already archived" if user.archived?
 
       user.archived_email = user.email
@@ -20,7 +20,7 @@ module Users
       # If we're archiving a user who still has applications then we want to
       # know about it otherwise its a user account for expression of interest,
       # or from a past incomplete registration
-      send_sentry_capture_message if blank_email && user.applications.any?
+      send_sentry_capture_message if enable_sentry && blank_email && user.applications.any?
     end
 
     def set_uid_to_nil!
--- a/app/views/admin/users/_users_side_navigation.html.erb
+++ b/app/views/admin/users/_users_side_navigation.html.erb
@@ -1,0 +1,20 @@
+<% content_for :side_navigation do %>
+  <%=
+    render SubNavigationComponent.new(
+      request.path,
+      structure:
+        [
+          NavigationStructure::Node.new(
+            name: "All users",
+            href: admin_users_path,
+            prefix: /\/admin\/users(?!\/duplicates)$/,
+            ),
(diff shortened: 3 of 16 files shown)


db/seeds/base/add_users.rb
alkesh, line 1: slightly confusing diff - I've moved the old `add_users.rb` to `add_admin_users.rb`, so this file is actually totally new and different content.

app/controllers/admin/users/duplicates_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Admin::Users
+  class DuplicatesController < AdminController
+    def index
+      scope = User
+        .where(email: nil)
+        .joins(:applications)
+        .where.not(applications: { lead_provider_approval_status: "rejected" })
+        .distinct
+        .order(created_at: :desc)
```
jebw: Can this be `created_at: :desc, id: :desc` - just helps avoid flaky specs

app/views/admin/users/_users_side_navigation.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        [
+          NavigationStructure::Node.new(
+            name: "All users",
+            href: admin_users_path,
+            prefix: /\/admin\/users(?!\/duplicates)$/,
+            ),
+          NavigationStructure::Node.new(
+            name: "Archived duplicate users",
+            href: admin_duplicate_users_path,
+            prefix: /\/admin\/users\/duplicates$/,
```
jebw: These can just be added to the global subnav structure and it'll get handled automatically - since there's nothing dynamic here I think it'll be easier to keep the global nav all in one place?

config/routes.rb
```diff
@@ -210,6 +210,9 @@
 
     resources :courses, only: %i[index show]
     resources :users, only: %i[index show] do
+      collection do
+        resources :duplicates, controller: "users/duplicates", as: "duplicate_users", only: :index
```
jebw, line 214: Isn't `controller: "users/duplicates"` the default value for that anyway?

app/services/users/archiver.rb
```diff
@@ -7,7 +7,7 @@ class Archiver
 
     attribute :user
 
-    def archive!(blank_email: false)
+    def archive!(blank_email: false, enable_sentry: true)
```
jebw: Can this be `notify_sentry` since its controlling notification - not enablement

db/seeds/base/add_admin_users.rb
```diff
@@ -0,0 +1,19 @@
+otp_testing_code = "00000"
```
jebw: Lets remove this and have the factory generate codes automatically, with short timestamps

db/seeds/base/add_users.rb
jebw: `git mv` is your friend here 😄 

I think that would result in the 'changes' flowing as you'd expect - ie `git mv` in one commit, then add the other in a following commit

db/seeds/base/add_admin_users.rb
alkesh: actually - do we even need otp_hash to be populated?
it can be left blank, and generated when the admin logs in.

config/routes.rb
alkesh: no, without that, the default controller-action is: `admin/duplicates#index`,
but with it, it is: `admin/users/duplicates#index`

db/seeds/base/add_admin_users.rb
jebw: ~I think we should move populating it to a trait - and use that for any specs which aren't signing in via the controller (if any)~ ignore me - looks like specs are passing without

